PR #6, opened 2017-01-14
Update README

tiagopog:
Fixing a non-aliased module `Helloworld.Greeter` in the example and also making clear to eventual gRPC newbies that the expected response in the stubbed method should be like the one defined in the sample proto.


tony612:
Thanks!

PR #17, opened 2017-07-22
elixir 1.4.5 and otp 20

jaigouk:
tested against,
```
elixir:
  - 1.4.4
  - 1.4.5
otp_release:
  - 19.3
  - 20.0
```


jaigouk:
i will spend more time to clean up the codes. there are more than 50 credo warnings.

darkofabijan:
@jaigouk lots of good stuff here 👏 :)

What do you think about splitting this PR into two or more PRs? Mainly I can think of (1) new Elixir/Erlang versions and (2) credo related work.

jaigouk:
@darkofabijan well, i think it's more about keeping chatterbox and cowboy up to date without breaking tests.

darkofabijan:
@jaigouk yea, you are right! `192eda6` 🛩 :)

tony612:
There're some problems:
- Elixir 1.4.5 and OTP20 are supported. There's no need to add 1.4.4.
- cowboy and exprotobuf have been updated to make the tests pass. And we'd better use the stable version of the dependencies if we can.
- I'm not very familiar with credo. But why not just using the default config? I think that should be enough.
- I can accept credo but not excoveralls and mix_test_watch. Some code is unstable and can changes often, so I don't want to introduce so many dependencies for the moment.
- Just as Darko said, you should split this PR to several.
- There're too many related small commits.

So, please:
- Just keep the credo related part in this PR. And send other PR if needed.
- Squash the commits to keep the git history clean.
- Keep your code clean like
  - Don't change the forked branch(cowboy & chatterbox) to your own forked one.
  - Remove `test/fixtures/helloworld.proto`, we don't need it now.

(I found you just do some changes, but I'll keep this comments so that you can refer to them).

jaigouk:
@tony612 thanks for the hard work. i will squash commits and try to keep this simple and clean. I am not trying to make things complicated. I respect your work. This is the only elixir version of grpc lib. 👍

tony612:
@jaigouk Thanks! Just some thoughts for things like credo and excoveralls: I do care about improving the tests and code quality using tools. But it's not very important for the moment so I won't spend time on that. If you or others think it's important now, you can fix them but please keep things simple so that we can maintain the code more easily. Hope you can understand😁

jaigouk:
@tony612 sure. no problem. We can make credo more softer. It checks Readability, Refactoring opportunities and Consistency. I think checking `Refactoring` would not bother you too much. there were only 3 of them and they were not that hard to fix. I am just trying to make this project awesome. So, I will follow what you think is important at this point. And there must be other important things than this PR.

For example, I wanted to add server side custom grpc error codes like mentioned [here](https://github.com/grpc/grpc/blob/master/doc/statuscodes.md)
* INVALID_ARGUMENT
* NOT_FOUND
* ABORTED
* OUT_OF_RANGE

tony612:
@jaigouk  Sounds good. But why not adding the codes grpc supports now like CANCELLED and DEADLINE_EXCEEDED. btw, maybe we can discuss this in an issue?

jaigouk:
@tony612 I will spend more time around grpc. and try to contribute. my knowledge is not yet enough to dig into cowboy and chatterbox. i just hope i can do something helpful instead of creating PRs like this one.

PR #24, opened 2017-08-10
GRPC module version loading

DamjanBecirovic:
Mix can not be used in escript or release.

### DIFF
--- a/lib/grpc.ex
+++ b/lib/grpc.ex
@@ -1,8 +1,10 @@
 defmodule GRPC do
+  @version GRPC.Mixfile.project[:version]
+
   @doc """
   Returns version of this project.
   """
-  def version, do: GRPC.Mixfile.project[:version]
+  def version, do: @version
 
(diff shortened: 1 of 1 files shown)


lib/grpc.ex
```diff
@@ -1,8 +1,10 @@
 defmodule GRPC do
+  @ver GRPC.Mixfile.project[:version]
```
darkofabijan: Why not `@version`? :)

DamjanBecirovic: I was not sure if it would overlap with function or if there is reserved attribute with that name.

darkofabijan: I think it will work fine. Please try it.

tony612: Of course not. There's no magic here :)

DamjanBecirovic: Tried it, works fine, I will commit update now.

Issue #27, opened 2017-08-29
Message is split in chunks of 2^14 bytes


arnaud-lb:
It seems that `Channel.unary` returns a response whose body is split in chunks of approximately 2^14 bytes.

As a result, `Stub.parse_unary_response` can fail to parse a message, since it's seeing only the first 2^14 bytes of it.

Due to this, I'm seeing the following error:

```
** (MatchError) no match of right hand side value: "(very long string of a bit less than 2^14 bytes)"
             lib/protobuf/decoder.ex:163: Protobuf.Decoder.decode_type/3
             lib/protobuf/decoder.ex:30: Protobuf.Decoder.do_decode/4
    (elixir) lib/enum.ex:1229: Enum."-map/2-lists^map/1-0-"/2
             lib/grpc/stub.ex:157: GRPC.Stub.parse_unary_response/2
```

arnaud-lb:
Changing parse_unary_response like this seems to fix the problem for me:

```
  defp parse_unary_response({_headers, data_list}, unmarshal) do
    data_list
    |> Enum.join("")
    |> GRPC.Message.from_data()
    |> unmarshal.()
  end
```

tony612:
@arnaud-lb I think you are right. How about sending a PR to fix it? 😁

arnaud-lb:
I'll try, but I promise nothing, given my current level in elixir.

PR #35, opened 2017-10-13
Add support for streamed messages split in multiple frames

ssboisen:
Hi,

As I mentioned in tony612/protobuf-elixir#11 I'm working on a Google Cloud Bigtable client and I'm running into problems working with googles developer emulator.

It turns out that the server might split the protobuf payload into multiple data frames that are less than the size of the payload even though the window size can contain the full payload. This results in an exception in [stub.ex L166](https://github.com/tony612/grpc-elixir/blob/87066828170f535330615f15dfe800b74380efdb/lib/grpc/stub.ex#L166) when the non-complete payload is unmarshalled. The exception is similar to the to the problem mentioned in #27 that was fixed in `parse_unary_response`.

This PR fixes this problem for streamed responses by buffering incomplete data frames until they are complete and can be treated as complete grpc messages.

One question:

The unfold accumulator keep a list of all the received data frames but doesn't use it for anything, what's the reason for this?

### DIFF
--- a/lib/grpc/stub.ex
+++ b/lib/grpc/stub.ex
@@ -158,18 +158,41 @@ defmodule GRPC.Stub do
   end
 
   defp response_stream(%{unmarshal: unmarshal} = stream, opts) do
-    Stream.unfold([], fn acc ->
+    Stream.unfold(%{ complete: [], buffer: :empty, message_length: -1 }, fn acc ->
       case Channel.recv(stream, opts) do
         {:data, data} ->
-          reply = data
-          |> GRPC.Message.from_data
-          |> unmarshal.()
-          {reply, [acc] ++ reply}
+          if GRPC.Message.complete?(data) do
+            reply = data |> GRPC.Message.from_data() |> unmarshal.()
+            {reply, Map.update!(acc, :complete, &([&1] ++ reply))}
+          else
+            case acc do
+              %{ buffer: :empty } ->
+                {
+                  :skip,
+                  Map.merge(acc, %{
+                    buffer: data, message_length: GRPC.Message.message_length(data)
+                  })
+                }
+              %{ buffer: buffer, message_length: ml } when byte_size(buffer) + byte_size(data) >  ml ->
+                final_buffer = buffer <> data
+                reply = final_buffer
+                        |> GRPC.Message.from_data()
+                        |> unmarshal.()
+                {
+                  reply,
+                  Map.update!(acc, :complete, &([&1] ++ reply))
+                  |> Map.merge( %{ buffer: :empty, message_length: -1 })
+                }
+              %{ buffer: buffer } ->
+                next_buffer = buffer <> data
+                {:skip, Map.put(acc, :buffer, next_buffer) }
+            end
+          end
         {:end_stream, _resp} ->
           nil
         other ->
           other
       end
-    end)
+    end) |> Stream.reject(&match?(:skip, &1))
   end
 end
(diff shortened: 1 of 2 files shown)


lib/grpc/stub.ex
```diff
@@ -158,18 +158,41 @@ defmodule GRPC.Stub do
...
+          else
+            case acc do
+              %{ buffer: :empty } ->
+                {
+                  :skip,
+                  Map.merge(acc, %{
+                    buffer: data, message_length: GRPC.Message.message_length(data)
+                  })
+                }
+              %{ buffer: buffer, message_length: ml } when byte_size(buffer) + byte_size(data) >  ml ->
```
tony612: Why not using `byte_size(buffer) + byte_size(data) - 5 == ml`?

lib/grpc/stub.ex
```diff
@@ -158,18 +158,41 @@ defmodule GRPC.Stub do
   end
 
   defp response_stream(%{unmarshal: unmarshal} = stream, opts) do
-    Stream.unfold([], fn acc ->
+    Stream.unfold(%{ complete: [], buffer: :empty, message_length: -1 }, fn acc ->
```
tony612: Please use the compact format for map, like `%{complate: [], ...}`

tony612:
@ssboisen ping

ssboisen:
I AFK this week, will get back to you monday :-)
tir. 17. okt. 2017 kl. 16.12 skrev Bing Han <notifications@github.com>:

jeffutter:
I think I'm running into this issue as well, any updates?

tony612:
@jeffutter I just merge this PR. Could you try the master branch to see if it's solved?

jeffutter:
@tony612 The request that was consisently failing before now works _most_ of the time. I'm not sure if I'm running into more of the same issue or a different one but I sometimes am getting 

```elixir
iex(23)> JobProto.JobService.Stub.get_jobs_for_interpreter(channel, JobProto.Id.new(%{id: 26})) |> Enum.to_list
** (MatchError) no match of right hand side value: <<0, 0, 0>>
    lib/grpc/message.ex:69: GRPC.Message.complete?/1
    lib/grpc/stub.ex:171: anonymous fn/4 in GRPC.Stub.response_stream/2
    (elixir) lib/stream.ex:1382: Stream.do_unfold/4
    (elixir) lib/stream.ex:1433: Enumerable.Stream.do_each/4
    (elixir) lib/enum.ex:1841: Enum.reverse/1
    (elixir) lib/enum.ex:2599: Enum.to_list/1
```

With varying numbers of `0` in `<<0,0,0>>>`

tony612:
@jeffutter Could you try with `try-fix-message-complete` branch to see if it's fixed? If not, we'd better try to reproduce your problem in an easy way so that I can fix it faster.

jeffutter:
@tony612 that branch seems to fix those errors that I was seeing. I've run the above query ~20 times with no errors.

I have noticed another problem but this one _may_ be with the ruby server I have. After 10-20 calls it just stops responding. The elixir client ends up returning an empty list `[]` rather than the expected list of 1910 items. However, if I kill the beam and restart I get the same behavior on the elixir side until I restart the ruby server - I'm pretty sure the issue is over there.

With the `try-fix-message-complete` I think this looks pretty good to me.

tony612:
@jeffutter `try-fix-message-complete` is now merged to master.

PR #36, opened 2017-10-14
WIP: Add grpc-status

xenolf:
WIP PR to add error handling based on GRPC error codes.

### DIFF
--- a/lib/grpc/status.ex
+++ b/lib/grpc/status.ex
@@ -1,0 +1,113 @@
+defmodule GRPC.Status do
+  
+  @doc """
+  Not an error; returned on success.
+  """
+  def ok, do: "0"
+
+  @doc """
+  The operation was cancelled (typically by the caller).
+  """
+  def cancelled, do: "1"
+
+  @doc """
+  Unknown error.  An example of where this error may be returned is
+  if a Status value received from another address space belongs to
+  an error-space that is not known in this address space.  Also
+  errors raised by APIs that do not return enough error information
+  may be converted to this error.
+  """
+  def unknown, do: "2"
+
+  @doc """
+  Client specified an invalid argument.  Note that this differs
+  from FAILED_PRECONDITION.  INVALID_ARGUMENT indicates arguments
+  that are problematic regardless of the state of the system
+  (e.g., a malformed file name).
+  """
+  def invalid_argument, do: "3"
+
+  @doc """
+  Deadline expired before operation could complete.  For operations
+  that change the state of the system, this error may be returned
+  even if the operation has completed successfully.  For example, a
+  successful response from a server could have been delayed long
+  enough for the deadline to expire.
+  """
+  def deadline_exceeded, do: "4"
+
+  @doc """
+  Some requested entity (e.g., file or directory) was not found.
+  """
+  def not_found, do: "5"
+
+  @doc """
+  Some entity that we attempted to create (e.g., file or directory) already exists.
+  """
+  def already_exists, do: "6"
+
+  @doc """
+  The caller does not have permission to execute the specified
+  operation.  PERMISSION_DENIED must not be used for rejections
+  caused by exhausting some resource (use RESOURCE_EXHAUSTED
+  instead for those errors).  PERMISSION_DENIED must not be
+  used if the caller can not be identified (use UNAUTHENTICATED
+  instead for those errors).
+  """
+  def permission_denied, do: "7"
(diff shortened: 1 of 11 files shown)


xenolf:
@tony612 tell me what you think of the changes.
This implementation works with other clients for streaming and unary requests. There is a problem with the elixir client though as there does not seem to be a place where the client checks the headers returned from the server. Do you have a hint as to where this should happen?

lib/grpc/status.ex
```diff
@@ -0,0 +1,113 @@
+defmodule GRPC.Status do
+  
+  @doc """
+  Not an error; returned on success.
+  """
+  def ok, do: "0"
```
tony612: Why no use int instead of string? When sending that on the wire, we can change the int to string.

lib/grpc/stub.ex
```diff
@@ -150,11 +150,18 @@ defmodule GRPC.Stub do
...
+    {_, status} = Enum.find(headers, nil, fn(header) -> elem(header, 0) == "grpc-status" end)
+    if status != GRPC.Status.ok do
+      {_, message} = Enum.find(headers, nil, fn(header) -> elem(header, 0) == "grpc-message" end)
+      {:error, %GRPC.RPCError{status: status, message: message}}
+    else
+      result = data_list
+      |> Enum.join("")
+      |> GRPC.Message.from_data
+      |> unmarshal.()
+      {:ok, result}
```
tony612: I prefer using `result` directly, like what [Erlang rpc](http://erlang.org/doc/man/rpc.html#call-4) does.

lib/grpc/stub.ex
```diff
@@ -150,11 +150,18 @@ defmodule GRPC.Stub do
...
-  defp parse_unary_response({_headers, data_list}, unmarshal) do
-    data_list
-    |> Enum.join("")
-    |> GRPC.Message.from_data
-    |> unmarshal.()
+  defp parse_unary_response({headers, data_list}, unmarshal) do
+    {_, status} = Enum.find(headers, nil, fn(header) -> elem(header, 0) == "grpc-status" end)
+    if status != GRPC.Status.ok do
+      {_, message} = Enum.find(headers, nil, fn(header) -> elem(header, 0) == "grpc-message" end)
+      {:error, %GRPC.RPCError{status: status, message: message}}
```
tony612: In some ways, I prefer raising the errors directly. After all, the client code raises the error. If we treat the RPCs as plain functions, the errors should be raised.

xenolf:
> I thought you've found that? I noticed you handled the headers in parse_unary_response. Do you mean the streaming calls?

Yes, streaming calls. I'm sorry I should have been more specific 😄

lib/grpc/stub.ex
xenolf: I was under the impression that elixir preferred to use this tuple construct versus raising errors? Failing RPCs are to be expected and I would not say that it's an exceptional thing. My main drive though was to be able to pattern match on the result of the call to be able to handle the error without diverting the execution flow with an exception.

lib/grpc/status.ex
xenolf: I had it as an int at first but if I return int here we can't match against the values returned from the servers without converting them again as they are returned as strings.

tony612: But enum seems a natural choice. Anyway, we can keep them strings and can change them later only if other will not use the literal values.

tony612:
@xenolf You're right. There's no handling for received HEADERS. How about implement errors handling for unary type in the PR. We can add that after I add the HEADERS logic?

lib/grpc/stub.ex
xenolf: How about offering both? As in `recv!` and `recv` where the former raises and the latter returns a tuple?

tony612: But for the unary rpc, we have to define two functions `xxx` and `xxx!`, which is annoying. Well, let's use the `{:error, }` tuple, just like what Erlang's rpc does.

lib/grpc/stub.ex
predrag-rakic: @tony612 please consider either returning `{:ok, result}` or having both `xxx()` and `xxx!()` forms generated.

It is hard to consume `result` | `{:error, reason}` response with `with ... do` construct.

tony612:
Sorry for late. Thank you! This is very helpful!

PR #40, opened 2017-12-03
Expose all ssl options in GRPC.Credential

cjab:
GRPC.Credential requires paths to certs and keys. It's not always convenient to store certs and keys as files. For example, one common way of including a CA bundle is [certifi](https://github.com/certifi/erlang-certifi). Using certifi the certs are returned as a binary.

Instead of creating options structs specific to grpc-elixir maybe it would be better to pass an `ssl_option` list (from [erlang ssl](erlang.org/doc/man/ssl.html)) through. Both the client and the server would accept the same options so it would mean you could do away with the client/server specific credentials. This would also be a little bit closer to the way credentials are implemented in some of the [official GRPC libraries](https://grpc.io/docs/guides/auth.html#using-client-side-ssltls).

### DIFF
--- a/examples/route_guide/lib/app.ex
+++ b/examples/route_guide/lib/app.ex
@@ -18,7 +18,7 @@ defmodule Routeguide.App do
 
   defp start_args do
     if System.get_env("TLS") do
-      cred = GRPC.Credential.server_tls(@cert_path, @key_path)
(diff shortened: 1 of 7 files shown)


tony612:
Thanks!

PR #42, opened 2017-12-16
Separate credentials and client state

cjab:
This started out as an attempt to fix some type errors in my project. I ran into a few in grpc-elixir that were bubbling up. In the process of cleaning up the types I made a few changes to `GRPC.Adapter.Chatterbox.Client`. These changes made defining types easier and my hope is that they make the client module clearer also.

There are still a couple of type errors that I would like to tackle but I didn't want this PR to get out of control 😃.


lib/grpc/channel.ex
```diff
@@ -10,18 +10,20 @@ defmodule GRPC.Channel do
     * `:host` - server's host to connect
     * `:port` - server's port to connect
     * `:scheme` - scheme of connection, like `http`
+    * `:cred` - credentials used for authentication
```
cjab: It feels like credentials are separate from the underlying adapter. If a new adapter was needed I think it would use the same credential information chatterbox does. Instead of storing them with adapter specific data I moved them out into the `Channel`.

lib/grpc/channel.ex
```diff
@@ -10,18 +10,20 @@ defmodule GRPC.Channel do
...
     * `:adapter` - a client adapter module, like `GRPC.Adapter.Chatterbox.Client`
-    * `:payload` - the payload needed by the adapter
+    * `:client` - an instance of a client adapter, like `%GRPC.Adapter.Chatterbox.Client{}`
```
cjab: It looks like `Stream` also has a payload. I kept getting confused by the name `payload` here. It seems like what this really is is the state of the client. In the case of chatterbox we can get away with just storing the `pname` in the client state.

tony612:
This PR is OK for me. Could you rebase master and resolve the conflicts. I'll merge it. 😁

cjab:
Everything should be good to go now. Thanks!

PR #50, opened 2018-03-22
Define @default_timeout in usecs

shiroyasha:
I think I've figured out part of the problem in this issue https://github.com/tony612/grpc-elixir/issues/48.

The default timeout was set to `5000`, but when it is encoded it is sent in microseconds.

- https://github.com/tony612/grpc-elixir/blob/master/lib/grpc/transport/http2.ex#L81
- https://github.com/tony612/grpc-elixir/blob/master/lib/grpc/transport/utils.ex#L13

I've changed the constant to `5_000_000` to reflect this in the stub.

### DIFF
--- a/lib/grpc/stub.ex
+++ b/lib/grpc/stub.ex
@@ -41,8 +41,9 @@ defmodule GRPC.Stub do
   @insecure_scheme "http"
   @secure_scheme "https"
   @canceled_error GRPC.RPCError.exception(GRPC.Status.cancelled(), "The operation was cancelled")
+
   # 5 seconds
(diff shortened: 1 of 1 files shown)


tony612:
Could you try the `fix-timeout` branch? I think it's a better solution to this problem https://github.com/tony612/grpc-elixir/commit/96163c50f79ffbeda27f8055c29908c2a1a865ef

shiroyasha:
@tony612 I agree, that solution looks better.

I've tried it out on a test app, and it seems to be working correctly. 👍

PR #62, opened 2018-05-24
Support interceptor

tony612:
I implemented a simple Logger middleware in lib/grpc/logger.ex and a prometheus middleware, which is a little complex, in https://github.com/tony612/elixir-grpc-prometheus. And usage can be found in `interop` as an example.

Some places I feel a little uncomfortable:
1. Many changes about `endpoint` in `lib/grpc/adapter/cowboy.ex`
2. (Server) Result of `next` has many formats, so we have to handle every case like 
```
# https://github.com/tony612/elixir-grpc-prometheus/blob/master/lib/grpc_prometheus/server_interceptor.ex#L102
    code =
      case result do
        {:ok, stream} -> # server streaming
          GRPC.Status.code_name(0)

        {:ok, stream, reply} -> # not server streaming
          GRPC.Status.code_name(0)

        {:error, %GRPC.RPCError{} = error} -> # user returns an error on purpose
          GRPC.Status.code_name(error.status)

        {:error, %{kind: kind, reason: reason, stack: stack}} -> # error raised in code
          GRPC.Status.code_name(GRPC.Status.unknown())
      end
```
3. (Server) For hooks around `send_reply`(only for server streaming), I added `__interface__` in `GRPC.Server.Stream` for override like
```
# https://github.com/tony612/elixir-grpc-prometheus/blob/master/lib/grpc_prometheus/server_interceptor.ex#L91
    send_reply = fn stream, reply ->
      stream = interface[:send_reply].(stream, reply)
      Counter.inc(registry: registry, name: :grpc_server_msg_sent_total, labels: labels)
      stream
    end
```
4. The order of interceptors' `call/4` in server and client are different: `req, stream, next, level` in server side and `stream, req, next, level` in client side. The reason is I want to keep the same order with users' code like `Stub.client_rpc(stream, req)` and `def server_rpc(req, stream)`.
5. (Client) For hooks around `send_request` and `recv`, I use `__interface__` like what I did in server side.
6. (Client) It's a little hard to handle server streaming enum when server side is streaming because Elixir's `Stream` don't provide a simpler API to tell us when the stream finishes, so we have to use `Stream.transform` like
```
# https://github.com/tony612/elixir-grpc-prometheus/blob/master/lib/grpc_prometheus/client_interceptor.ex#L130
  defp new_res_enum(monitor, enum) do
    Stream.transform(enum, fn() -> :ok end, fn(elem, acc) ->
      msg_received(monitor)
      {[elem], acc}
    end, fn(_) ->
      handled_rpc(monitor, GRPC.Status.code_name(0))
    end)
  end
```
7. (Client) Result of `next` has many formats as server side like `{:ok, reply}, {:ok, reply, map}, {:error, error}`


tyrchen:
When will this featured be released?

PR #66, opened 2018-06-08
Support domain sockets

drowzy:
* Allow starting server listening on a domain socket
* Allow stub client to connect to domain sockets

I had to update the deps to use the latest [gun 1.0.0-rc.1](https://github.com/ninenines/gun/tree/1.0.0-rc.1), since it's not on hex yet it uses the git ref. I can update the PR when it's available on hex.


tony612:
Could you fix the CI and squash commits into one?

drowzy:
Fixed :)

PR #72, opened 2018-07-29
Passing JWT token in authorization header

clearjs:
This pull request implements ability to specify a JWT access token per invocation of any client stub function. Example usage is provided in https://github.com/tony612/grpc-elixir/issues/71.

If I did anything wrong, please let me know.


clearjs:
There's a better way to implement this which doesn't require a user to supply tokens per call. I'll make another pull request later.

PR #75, opened 2018-10-02
Support Ranch 1.6 map-style transport opts

JayH5:
This updates grpc to use Ranch 1.6's map-style transport options. Previously a keyword list was used. When using a keyword list and both Ranch options and socket options are specified, Ranch logs deprecation warnings.

This change requires a newer version of Cowboy with a fix for ninenines/cowboy#1324.


JayH5:
Seems to be a transient test failure on Elixir 1.4.5/OTP 19.3. The tests passed for me in another Travis run: https://travis-ci.com/JayH5/grpc-elixir/builds/86719993

Issue #88, opened 2019-01-24
client reports deadline expired


AleksandarFilipov:
On (inifinite) stream subscription client reports 
```elixir
{:error, %GRPC.RPCError{message: "Deadline expired", status: 4}
```
after 10 seconds. Due to the timeout the connections is dropped. Until dropped connection everything works as expected.

The following dosent't change anything
```elixir
{:ok, channel} = GRPC.Stub.connect("localhost:50051", deadline: 2000)
```

however the following does:
```elixir
channel |> Myapi.NetworkService.Stub.subscribe_to_stream(request, timeout: 12000)
```

effectively the 12000 call postopones the "Deadline expired" from 10s to 12s. Agan until dropped connection everything works a charm

in stub I can ser that timeout is set to 10000
https://github.com/tony612/grpc-elixir/blob/fc975b43b95f29232b27ba5268a9967b838bd740/lib/grpc/stub.ex

```elixir
@default_timeout 10000
```

I would dare to say that the problem is client side. Other clients (as python) seem to work fine on the same server (also running elixir)

tony612:
How about setting timeout to `:infinity`?

AleksandarFilipov:
As soon as I'm in context again I'll give it a try.

AleksandarFilipov:
So confirming that :infinity works well and solves my issue. However i'm confused on the timeout mechanism which doesn't (which I would expect) postpone the timeout if traffic arrives. Is this intentional?

tony612:
From your code, I know the problem. Timeout for streaming RPCs is for the all connection. Like if you calls have 3 streaming calls, each of which take 200ms, then you have to set timeout to > 600ms.

PR #89, opened 2019-01-29
Added deadline test on streams

AleksandarFilipov:
Reveals potential bug #88 

Pull request does not contain fix.


AleksandarFilipov:
Added test should make it easier to pinpoint the issue (unless I've misunderstood). Deadline should not kick in if data is continuously received.

PR #94, opened 2019-03-13
Update typespec to satisfy dialyzer

fredr:
Since marshal and unmarshal is not yet set when calling `GRPC.Stub.call/5`, this typespec will be broken https://github.com/tony612/grpc-elixir/blob/master/lib/grpc/stub.ex#L190

Otherwise dialyzer will warn that the call made [here](https://github.com/tony612/grpc-elixir/blob/master/lib/grpc/stub.ex#L75) breaks the contract


fredr:
@tony612 any chance this could get merged?

PR #104, opened 2019-06-25
Add support for setting authorization header

c-h-:


lib/grpc/transport/http2.ex
```diff
@@ -33,6 +33,7 @@ defmodule GRPC.Transport.HTTP2 do
   @spec client_headers_without_reserved(GRPC.Client.Stream.t(), map) :: [{String.t(), String.t()}]
   def client_headers_without_reserved(%{codec: codec} = stream, opts \\ %{}) do
     [
+      {"authorization", "#{opts[:authorization] || ""}"},
```
tony612: How about using `opts[:authorization]` directly?

beligante: What about remove this line and have something like this:

`|> append_authorization(opts[:authorization])`

and then:
```
defp append_authorization(headers, authorization) when is_bitstring(authorization) do
	[{"authorization", "#{authorization}"} | headers]
end
```

This way we avoid send extra headers

tony612:
This can be added by connection level headers(https://github.com/elixir-grpc/grpc/pull/112) or `:metadata` option.

PR #112, opened 2019-08-30
Add mechanism to set headers in Stub.connect/3

zph:
👋. I'm excited to see all the work and progress happening here with GRPC + Elixir! Thank you.

I spiked out a general solution for setting headers at the channel level using some of the existing tooling. I see this patch providing a general solution for https://github.com/elixir-grpc/grpc/pull/104 and https://github.com/elixir-grpc/grpc/pull/72, without needing a per header function set at the request level (better ux for users).

This patch provides a generic way to set headers that will be attached
to each request from this `channel` by re-using
append_custom_metadata/2.

The use case is to provide a location for setting authorization headers
that does not require manually adding it to each Stub fn call.

With this change, headers can be set as:

```
channel = Stub.connect(host, port, headers: [{"authorization", "Bearer TOKEN"}])
```

On the receiving server side they can be accessed using existing getter
function. For the authorization example, this can be done in an
Interceptor:

```elixir
defmodule Authenticator do
  def init(_), do: []
  def call(req, stream, next, opts) do
    with headers when is_map(headers) <- GRPC.Stream.get_headers(stream),
         token when not is_nil(token) <- Map.get(headers, "authorization") do
         .... # use token
         .... # if authenticated
         next.(req, stream)
    end
  end
end
```

Adds two tests, one to verify on client side that headers can be set and
accessible, a second test is for integration to verify the roundtrip of
a header from client to server, then encode the header into a GRPC
message response for test verification.

Thank you @clearjs for the initial concept that I used as initial concept leading to this PR.

@tony612 I'm happy to have feedback on approach or advice on extra areas to test that are unaddressed. 👍


lib/grpc/transport/http2.ex
```diff
@@ -40,6 +40,7 @@ defmodule GRPC.Transport.HTTP2 do
     ]
     |> append_compressor(stream.compressor)
     |> append_accepted_compressors(stream.accepted_compressors)
+    |> append_custom_metadata(stream.channel.headers)
```
zph: Re-use existing merge mechanism.

test/grpc/channel_test.exs
```diff
@@ -13,4 +13,10 @@ defmodule GRPC.ChannelTest do
     {:ok, channel} = GRPC.Stub.connect("10.1.0.0:50051", adapter: ClientAdapter, cred: cred)
     assert %Channel{host: "10.1.0.0", port: 50051, scheme: "https", cred: ^cred} = channel
   end
+
+  test "connect/2 allows setting default headers" do
+    headers = [{"authorization", "Bearer TOKEN"}]
+    {:ok, channel} = GRPC.Stub.connect("10.1.0.0:50051", adapter: ClientAdapter, headers: headers)
+    assert %Channel{host: "10.1.0.0", port: 50051, headers: ^headers} = channel
```
zph: Verify basic structure and setting.

test/grpc/integration/server_test.exs
```diff
@@ -20,6 +20,11 @@ defmodule GRPC.Integration.ServerTest do
     def say_hello(req, _stream) do
       Helloworld.HelloReply.new(message: "Hello, #{req.name}")
     end
+
+    def check_headers(_req, stream) do
+      token = GRPC.Stream.get_headers(stream)["authorization"]
+      Helloworld.HeaderReply.new(authorization: token)
```
zph: This RPC helps with integration testing the headers.

test/grpc/integration/server_test.exs
```diff
@@ -157,4 +162,20 @@ defmodule GRPC.Integration.ServerTest do
...
+
+      {:ok, channel} =
+        GRPC.Stub.connect("localhost:#{port}",
+          headers: [{"authorization", token}]
+        )
+
+      {:ok, reply} =
+        channel |> Helloworld.Greeter.Stub.check_headers(Helloworld.HeaderRequest.new())
+
+      assert reply.authorization == token
```
zph: Verify roundtrip.

My understanding is that headers must be set on client and if desired set anew on server for the return trip. So I return them here in the `Reply` for testing.

test/support/helloworld.pb.ex
```diff
@@ -1,31 +1,56 @@
 defmodule Helloworld.HelloRequest do
-  use Protobuf
+  @moduledoc false
```
zph: Regenerated. I used the general command for recreation, but if I need to try again with other args, I can do so.

test/support/helloworld.proto
```diff
@@ -6,6 +6,7 @@ package helloworld;
 service Greeter {
   // Sends a greeting
   rpc SayHello (HelloRequest) returns (HelloReply) {}
+  rpc CheckHeaders (HeaderRequest) returns (HeaderReply) {}
```
zph: Helper for testing.

zph:
@tony612 If I can put together a helper for regenerating `proto` in a more automatic fashion, would you like me to submit that in a separate PR? I like the idea of removing that as a manual step when in `:dev` so that proto and pb.ex never diverge.

clearjs:
@zph thank you. It looks like you had more requirements in mind than I did in https://github.com/clearjs/grpc-elixir/commit/3dfa4648711ec0a59bb3bba2766de9ce770a0789. I only needed to be able to authenticate to Google PubSub using this application. I didn't spend enough time to understand your requirements and the approach you took, but I think that sharing mine might be useful.

So, with the above commit I'm able to write the following code:

```elixir
project_id="..." # your project_id from Google PubSub
token="..." # token string from Goth
{:ok, channel} = GRPC.Stub.connect("pubsub.googleapis.com:443", [cred: GRPC.Credential.new(ssl: [])])
req = Google.Pubsub.V1.ListTopicsRequest.new(project: "projects/#{project_id}", page_size: 5)
{_, _reply} = channel |> Google.Pubsub.V1.Publisher.Stub.list_topics(req, token: token, content_type: "application/grpc")
```

This way, I get authenticated to Google PubSub and can use it. I retrieve `token` using Goth configured with a secret json file generated to provide access to PubSub. I retrieve and provide this token each time I make an API call, so it's always fresh. I think that Goth refreshes it about once an hour and keeps the result cached. The downside is that it isn't convenient to pass it with each call.

So, my intention was instead to provide a function when creating the channel which would get the fresh value of token from Goth and return it. Something like:

```elixir
scope = "https://www.googleapis.com/auth/pubsub"
{:ok, channel} = GRPC.Stub.connect("pubsub.googleapis.com:443", [cred: GRPC.Credential.new(ssl: []), auth: fn () -> "Bearer #{Goth.Token.for_scope(scope).token}" end])
```

Then the grpc application should invoke this lambda function for each api request and add the result to authentication header just like I did above.

My understanding, after I spent very little time reading the code and text (it's late here in Ukraine), is that your approach is probably more generic and covers additional use cases, but it may after about an hour keep stale token value specified initially. Hope that this information helps.

Also, note that it is probably not needed to specify `content_type` explicitly anymore. I didn't check, but I remember somebody wrote that `"application/grpc"` is now the default value.

clearjs:
@zph thank you. It looks like you had more requirements in mind than I did in https://github.com/clearjs/grpc-elixir/commit/3dfa4648711ec0a59bb3bba2766de9ce770a0789. I only needed to be able to authenticate to Google PubSub using this application. I didn't spend enough time to understand your requirements and the approach you took, but I think that sharing mine might be useful.

So, with the above commit I'm able to write the following code:

```elixir
project_id="..." # your project_id from Google PubSub
token="..." # token string from Goth
{:ok, channel} = GRPC.Stub.connect("pubsub.googleapis.com:443", [cred: GRPC.Credential.new(ssl: [])])
req = Google.Pubsub.V1.ListTopicsRequest.new(project: "projects/#{project_id}", page_size: 5)
{_, _reply} = channel |> Google.Pubsub.V1.Publisher.Stub.list_topics(req, token: token, content_type: "application/grpc")
```

This way, I get authenticated to Google PubSub and can use it. I retrieve `token` using Goth configured with a secret json file generated to provide access to PubSub. I retrieve and provide this token each time I make an API call, so it's always fresh. I think that Goth refreshes it about once an hour and keeps the result cached. The downside is that it isn't convenient to pass it with each call.

So, my intention was instead to provide a function when creating the channel which would get the fresh value of token from Goth and return it. Something like:

```elixir
scope = "https://www.googleapis.com/auth/pubsub"
{:ok, channel} = GRPC.Stub.connect("pubsub.googleapis.com:443", [
    cred: GRPC.Credential.new(ssl: []), 
    auth: fn () -> "Bearer #{Goth.Token.for_scope(scope).token}" end
])
```

Then the grpc application should invoke this lambda function for each api request and add the result as the `authorization` header just like I did above.

My understanding, after I spent very little time reading your code and comments (it's late here in Ukraine), is that your approach is probably more generic and covers additional use cases, but it may after about an hour keep stale token value specified initially. Hope that this information helps.

Also, note that it is probably not needed to specify `content_type` explicitly anymore. I didn't check, but I remember somebody wrote that `"application/grpc"` is now the default value.

clearjs:
@zph thank you. Your approach is more generic than mine was in https://github.com/clearjs/grpc-elixir/commit/3dfa4648711ec0a59bb3bba2766de9ce770a0789. I only needed to be able to authenticate to Google Cloud (for example, PubSub). You allow specifying arbitrary headers per channel. Whether there are use cases for this is unclear, but this generalization might be very useful. It would help if a few possible use cases were mentioned.

So, with the above commit I'm able to write the following code:

```elixir
project_id="..." # your project_id from Google PubSub
token="..." # token string from Goth
{:ok, channel} = GRPC.Stub.connect("pubsub.googleapis.com:443", [cred: GRPC.Credential.new(ssl: [])])
req = Google.Pubsub.V1.ListTopicsRequest.new(project: "projects/#{project_id}", page_size: 5)
{_, _reply} = channel |> Google.Pubsub.V1.Publisher.Stub.list_topics(req, token: token, content_type: "application/grpc")
```

(Note that it is probably not necessary to specify `content_type` explicitly anymore. I didn't check, but I remember somebody wrote that `"application/grpc"` is now the default value.)

With the above code, I get authenticated to Google PubSub and can use it. I retrieve `token` using Goth configured with a secret JSON file generated to provide access to PubSub. I retrieve and provide this token each time I make an API call, so it's always fresh. I think that Goth refreshes it about once an hour and keeps the result cached. The downside is that it isn't convenient to pass it with each call.

I planned to allow passing a function when creating the channel. This function would get a fresh value of token from Goth and return it. Something like:

```elixir
scope = "https://www.googleapis.com/auth/pubsub"
{:ok, channel} = GRPC.Stub.connect("pubsub.googleapis.com:443", [
    cred: GRPC.Credential.new(ssl: []), 
    auth: fn () -> "Bearer #{Goth.Token.for_scope(scope).token}" end
])
```

Then the grpc application would invoke this lambda function for each api request and add the result as the `authorization` header just like I did above.

With your approach after some time (about an hour according to my tests) the channel may need to be recreated, because the token value specified initially would become stale. This can be solved in a generic way if you allow passing a function that would return a list of headers:


```elixir
scope = "https://www.googleapis.com/auth/pubsub"
{:ok, channel} = GRPC.Stub.connect("pubsub.googleapis.com:443", [
    cred: GRPC.Credential.new(ssl: []), 
    per_call_headers: fn () ->
      [{"authorization", "Bearer #{Goth.Token.for_scope(scope).token}"}]
    end
])
```

Passing constant headers can also be kept to simplify client code in cases where the token cannot change during the channel lifetime. But I'm not sure that the minor inconvenience of wrapping headers inside a function in, likely, just one place where the channel is created, is significant.

lib/grpc/stub.ex
```diff
@@ -185,13 +188,15 @@ defmodule GRPC.Stub do
...
+    uniform_fn =
       if function_exported?(:rand, :uniform_real, 0) do
-        (:rand.uniform_real() - 0.5) / 2.5
+        :uniform_real
       else
-        (:rand.uniform() - 0.5) / 2.5
+        :uniform
       end
 
+    jitter = (apply(:rand, uniform_fn, []) - 0.5) / 2.5
```
yordis: Wouldn't be better to do this?

```suggestion
    jitter = :rand.uniform_fn() - 0.5) / 2.5
```

tony612:
Merged. Thank you!

PR #113, opened 2019-09-05
DO NOT MERGE: Use Cowboy/Gun upstream master

essen:
I've reimplemented the retry using the functionality available in Gun master. Both Gun and Cowboy now have graceful shutdown; Gun has flow control for the messages it sends (unused currently); and Cowboy has better WINDOW_UPDATE handling.

There's some intermittent test failures, and I don't know Elixir so going to be tough for me to solve.

Obviously do not merge before Gun/Cowboy get new releases.

Tell me what you think.


lib/grpc/adapter/gun.ex
```diff
@@ -12,24 +12,26 @@ defmodule GRPC.Adapter.Gun do
...
 
     do_connect(channel, open_opts)
   end
 
   defp do_connect(%{host: host, port: port} = channel, open_opts) do
     open_opts =
-      Map.merge(%{retry: :infinity, retry_timeout: &GRPC.Stub.retry_timeout/1}, open_opts)
+      Map.merge(%{retry: 1, retry_fun: fn(retries, _) ->
+        %{retries: retries + 1, timeout: GRPC.Stub.retry_timeout(retries)}
+      end}, open_opts)
```
sinorga: Hi @essen, @tony612 
I am trying to speed up the progress of this PR.
I tested a little bit the code and trace the gun implementation. When you set `retry: 1`, it does not retry actually. It is minus 1 before pass to `not_connected` function. So it causes some test cases failed. https://github.com/ninenines/gun/blob/master/src/gun.erl#L1439

To make the behavior keep the same as before, ie. "retry forever and increase timeout after each retry", the retry should be set to 2. But I think this code didn't show the intention well.

essen: Sounds like a bug in Gun, if retry is set to 1 it should retry at least once. I will look into it.

essen: Thanks for the debugging by the way!

essen: Please try the most recent Gun I should have fixed the issue.

sinorga: Cool, it's working at my local, thanks! 
In case if you are not familiar with elixir tools, you could run `mix deps.update gun` to update `mix.lock`, then push the changes. Hopefully, it should pass the CI test. 😄

sinorga: Hi @essen, do you have any release schedule about the cowboy/gun? thanks

essen: Cowboy 2.7 is next week. Gun 2.0 probably at the end of the year, but I'd recommend using a pre-release before that if that's OK for the grpc project. I can issue a pre-release with fixed dependencies if necessary. The blocker for Gun 2.0 is HTTP/2 CONNECT/Websocket which might or might not require some public interface changes.

sinorga: cool, thanks, I think the cowboy is more commonly used, so it helps a lot.

btw, I try to update the cowboy to the latest commit: cc54c207e35f3ab7a2dfc105eef39fe7d3bf1633, but the test failed.  I didn't figure out the root cause yet. The server process seems to be blocked in https://github.com/ninenines/cowboy/blob/cc54c207e35f3ab7a2dfc105eef39fe7d3bf1633/src/cowboy_req.erl#L522 until timeout.

essen: Don't forget to also update Cowlib.

essen: If you could try tracing to see what's happening that would help, I can't look right now. Which test is it by the way? It could just be that the HTTP/2 CVE fixes get in the way of the tests, as was the case in some of Cowboy's own test suite.

sinorga:
You should be able to reproduce the error by the test case below.

```
$> mix test test/grpc/integration/server_test.exs:80

Excluding tags: [:test]
Including tags: [line: "80"]

  1) test multiple servers works (GRPC.Integration.ServerTest)
     test/grpc/integration/server_test.exs:80
     ** (MatchError) no match of right hand side value: {:error, %GRPC.RPCError{message: "Deadline expired", status: 4}}
     code: run_server([FeatureServer, HelloServer], fn port ->
     stacktrace:
       test/grpc/integration/server_test.exs:84: anonymous fn/1 in GRPC.Integration.ServerTest."test multiple servers works"/1
       (grpc) test/support/integration_test_case.ex:15: GRPC.Integration.TestCase.run_server/3
       test/grpc/integration/server_test.exs:81: (test)



Finished in 10.2 seconds
7 tests, 1 failure, 6 excluded
```

Yeah, I will try to trace the code to see if I can help to find out the issue.

essen:
Yeah I have no idea how to read this test. Looking forward to your investigation!

sinorga:
Hi I think the issue is caused by the latest cowboy changed a message format in this [commit](https://github.com/ninenines/cowboy/commit/20660d7566b63977e80f694724fee890d875ec1b).
From `{read_body, Ref, Length, Period}` to  `{read_body, self(), Ref, Length, Period}`

And gRPC uses a customized stream handler `src/grpc_stream_h.erl` which can't handle the new message format of read_body info. I don't know the story why gRPC library of the customized handler, but it passes the test after I switch back to the default handler `cowboy_stream_h.erl`. @tony612 could you provide some information about it?

essen:
Good job figuring this out! Maybe it was a workaround for an issue that should now be fixed, the HTTP/2 window was not handled very well before.

essen:
Cowboy 2.7 is out.

deadtrickster:
Hi, does this still error with latest cowboy?

essen:
It's not a problem in Cowboy, it's due to using a custom stream handler replacing `cowboy_stream_h`. I do not know why the custom stream handler is used for this or if it's still necessary. Probably isn't, but I am not currently working on this.

tony612:
This is fixed on master.

tony612:
@sinorga The reason why a custom grpc_stream_h is used is because at early days, it's hard to implement gRPC streaming feature with the default handler. Now there is `auto` mode, but there's still some problems, I'll create a ticket to address them.

PR #117, opened 2019-11-04
Route guide example simplification

alanraul:
### What did I do?

- code documentation
- README documentation
- separate content from route_guide protobuf
- code simplification


tony612:
I don't think it's necessary to change much:
* proto is same with this https://github.com/grpc/grpc/blob/master/examples/protos/route_guide.proto, so we don't want to write it by hand. And also for https://github.com/grpc/grpc/blob/master/examples/cpp/route_guide/route_guide_db.json
* There's already `protoc` command in the README now.
* There's no need to create folders for client.ex, data.ex and so on.

I think we only need to add more logs to client.ex if you want to.

tony612:
Please rebase/merge master.

danielwillemse:
Any chance this could get rebased/merged, so it can be merged? @alanraul 
I think the examples and documentation are really valuable.
Seeing as the current master is missing examples for features like streams and bidirectional streams.

polvalente:
@alanraul Is this still relevant? If so, please update the branch.

polvalente:
Closing this because it had been stale with changes requested even around the time it was opened.

PR #118, opened 2019-11-04
Allow client interceptor to send headers

mairbek:


test/grpc/integration/interceptor_headers.exs
```diff
@@ -0,0 +1,60 @@
+defmodule GRPC.Integration.EndpointTest do
```
tony612: Could you rename this file to `interceptor_headers_test.exs` and rename the module to `GRPC.Integration.InterceptorHeadersTest`? btw, it's better to rename them to "client_interceptor..."

test/grpc/integration/interceptor_headers.exs
```diff
@@ -0,0 +1,60 @@
...
+    def init(_), do: []
+
+    def call(req, stream, next, _) do
+      headers = GRPC.Stream.get_headers(stream)
+      %{name: original_name} = req
+      label = headers["x-test-label"]
+      updated = Helloworld.HelloRequest.new(name: "#{original_name} #{label}")
+      next.(updated, stream)
+    end
+  end
```
tony612: Why is this interceptor is needed? I guess we only need verify headers in HelloServer?

lib/grpc/transport/http2.ex
```diff
@@ -43,6 +43,7 @@ defmodule GRPC.Transport.HTTP2 do
     |> append_custom_metadata(stream.channel.headers)
     |> append_encoding(opts[:grpc_encoding])
     |> append_timeout(opts[:timeout])
+    |> append_stream_metadata(stream)
```
tony612: Could we use `append_custom_metadata(stream.headers)`?

mairbek:
thank you @tony612 , i have addressed the comments

PR #119, opened 2019-11-08
Add ability to retreive the peer

szahariev:
This is mainly driven by the fact that you cannot retrieve the client's IP address, which is vital for logging.


tony612:
Could you add a test and merge/rebase master?

szahariev:
@tony612: Can you take a look at the test and tell me if it looks ok to you?

PR #121, opened 2019-11-27
Remove get_stacktrace

vikger:
Special exception handling can be removed, there is no need to separate OTP 19 stacktrace handling.
The new try/catch stacktrace handling mechanism was introduced in OTP 21, making the get_stacktrace function deprecated.
Safe to merge, the current solution is not using the new feature of OTP 21, and OTP 19 is not supported any more.


tony612:
Most of `src/grpc_stream_h.erl` are copied from cowboy's `cowboy_stream_h.erl`, so I prefer not to change it.

Issue #141, opened 2020-02-08
Timeout in bi-drectional stream


mairbek:
**Describe the bug**
After upgrading to the HEAD bi-directional RPCs started timing out. A debugging session that we ran indicates that 

```
  {s, body, req} = :cowboy_req.read_body(req, opts)
```

is timing out with a default timeout which is 15 seconds + 1. I was able to write a test that indicates the bug, https://github.com/mairbek/grpc/commit/0d92937c65f7ed1862062347e07f64e0aa8c1e18

Note that downgrading to the older version of the library fixes this issue. `cowboy_req.read_body` returns in 15 seconds.

```

14:07:23.938 [warn]  sending request {%{bindings: %{}, body_length: :undefined, cert: :undefined, has_body: true, headers: %{"content-type" => "application/grpc+proto", "te" => "trailers", "user-agent" => "grpc-elixir/0.4.0"}, host: "localhost", host_info: :undefined, method: "POST", path: "/routeguide.RouteGuide/RouteChat", path_info: :undefined, peer: {{127, 0, 0, 1}, 56062}, pid: #PID<0.272.0>, port: 56061, qs: "", ref: "GRPC.Integration.ServiceTest.FeatureServer", resp_headers: %{"content-type" => "application/grpc+proto"}, scheme: "http", sock: {{127, 0, 0, 1}, 56061}, streamid: 1, version: :"HTTP/2"}, %{}}

14:07:38.938 [warn]  done requests

14:07:38.940 [warn]  sending request {%{bindings: %{}, body_length: :undefined, cert: :undefined, has_body: true, headers: %{"content-type" => "application/grpc+proto", "te" => "trailers", "user-agent" => "grpc-elixir/0.4.0"}, host: "localhost", host_info: :undefined, method: "POST", path: "/routeguide.RouteGuide/RouteChat", path_info: :undefined, peer: {{127, 0, 0, 1}, 56062}, pid: #PID<0.272.0>, port: 56061, qs: "", ref: "GRPC.Integration.ServiceTest.FeatureServer", resp_headers: %{"content-type" => "application/grpc+proto"}, scheme: "http", sock: {{127, 0, 0, 1}, 56061}, streamid: 1, version: :"HTTP/2"}, %{}}

14:07:40.925 [warn]  done requests

[...]

PR #142, opened 2020-02-08
Default read timeout is :infinity

mairbek:
Potential fix for https://github.com/elixir-grpc/grpc/issues/141

### DIFF
--- a/lib/grpc/adapter/cowboy/handler.ex
+++ b/lib/grpc/adapter/cowboy/handler.ex
@@ -401,7 +401,7 @@ defmodule GRPC.Adapter.Cowboy.Handler do
   defp timeout_left_opt(timer, opts \\ %{}) do
     case timer do
       nil ->
-        opts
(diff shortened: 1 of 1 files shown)


isaacsanders:
Should this not just be configurable from public interface? Defaulting to infinity feels broken.

prydonius:
The gRPC protocol suggests that the default should be infinity:

> If Timeout is omitted a server should assume an infinite timeout. Client implementations are free to send a default minimum timeout based on their deployment requirements.

https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md

tony612:
@isaacsanders I think it's okay to use `infinity` here because it's used only when no timer is set, which means clients doesn't specify a timeout.

tony612:
I created a PR to set default timeout to `:infinity` only for streaming calls. Please leave some comments.

PR #153, opened 2020-04-17
Don't format parens around `field`

ethransom:
Our project is importing elixir-grpc, and all DSL functions are making it through the formatter except for `field`. This PR adds `field` to the formatter DSL whitelist.


ethransom:
I found it a little curious that the check format action passed with this change, despite no code changes, so I investigated. 

`field` comes from the `protobuf` library, not this repo. As such, the fix was adding `protobuf` to our project's `import_deps` value. Sorry for the spurious PR, thanks for an amazing library!

PR #157, opened 2020-06-01
perf: Remove unnecessary iteration in http2 decoding

asummers:
Noticed this earlier when looking through the code.


tony612:
I guess this doesn't improve the performance much, but I'll merge this. Thank you!

PR #161, opened 2020-06-09
feat: expose max_connections ranch transport option

jhowarth:
This exposes the `max_connections` option for ranch and raises the default from `1024` to `16384`. `1024` is not an ideal default for a production system.


tony612:
Thank you!

PR #171, opened 2020-08-19
Provide a mechanism to inject a custom status handler

carrascoacd:
## Problem
In order to support Traefik we need to listen in the same port the h1 health checks, more info here https://github.com/containous/traefik/issues/6211

## Expected outcome
Add a way to inject a custom status handler, so we can listen for h1 requests in the same port using the desired handler.

Let me know if it works for you in order to update the doc as well or if you have another solution.

The configuration could be:
```elixir
config :grpc,
       :status_handler,
       {"/status", Phoenix.Endpoint.Cowboy2Handler, {My.Endpoint, []}}
```


carrascoacd:
@tony612 could you review it when you have time, please?

polvalente:
@carrascoacd is this still relevant?

lib/grpc/adapter/cowboy.ex
```diff
@@ -153,11 +153,21 @@ defmodule GRPC.Adapter.Cowboy do
   end
 
   defp cowboy_start_args(endpoint, servers, port, opts) do
-    dispatch =
-      :cowboy_router.compile([
-        {:_, [{:_, GRPC.Adapter.Cowboy.Handler, {endpoint, servers, Enum.into(opts, %{})}}]}
-      ])
+    # Custom handler to be able to listen in the same port, more info: 
+    # https://github.com/containous/traefik/issues/6211
+    status_handler = Application.get_env(:grpc, :status_handler)
```
polvalente, line 158: I think we should avoid using global config as per the general library guidelines for Elixir. Instead, we should accept this through `opts`.

Is this able to be passed down already, or would something more need to change?

carrascoacd: Hi @polvalente, thanks for answering. This is something we still need. Let me try injecting the configuration using `opts` and check if it works.

polvalente: Yes! This is what I had in mind.
The benefit of doing it this way is that 2 processes can be started with different configs

carrascoacd: Great, much better. I've submitted the change!

PR #175, opened 2020-09-21
fix: Don't crash on errors from the adapter

frekw:
This fixes errors such as:

```
    Process #PID<0.32571.2> terminating
** (CaseClauseError) no case clause matching: {:error, %GRPC.RPCError{message: ":stream_error: {:stream_error, :cancel, :\"Stream reset by server.\"}", status: 13}}
    (grpc 0.5.0-beta.1) lib/grpc/stub.ex:462: GRPC.Stub.recv_body/5
    (grpc 0.5.0-beta.1) lib/grpc/stub.ex:433: GRPC.Stub.do_recv/2
```

which I've been seeing.


ryanwinchester:
Fixes #188?

polvalente:
@ryanwinchester looks like this does fix the issue you pointed. I'll mark that as closed upon merge

PR #183, opened 2021-01-19
use official gun 2.0.0-rc.1 version (instead of grpc_gun)

tyler-nguyen:
Now `gun` has its own `2.0.0-rc.1` version. I think it's better to use that instead of the `grpc_gun` version.


mix.exs
```diff
@@ -37,10 +37,10 @@ defmodule GRPC.Mixfile do
   defp deps do
     [
       {:protobuf, "~> 0.5"},
-      {:cowboy, "~> 2.7"},
-      {:gun, "~> 2.0.0", hex: :grpc_gun},
-      # 2.9.0 fixes some important bugs, so it's better to use ~> 2.9.0
-      # {:cowlib, "~> 2.9.0", override: true},
+      {:cowboy, "~> 2.8"},
+      {:gun, "~> 2.0.0-rc.1"},
```
polvalente, line 41:
```suggestion
      {:gun, "~> 2.0.0-rc.2"},
```

Let's go with rc2 since we're already down this path and it now exists :)

polvalente:
I'll apply this update in #224. Thanks!

PR #189, opened 2021-02-15
Fix and update GitHub Actions config

kamilkowalski:
Some GitHub Actions are failing on `master`, so this PR:

* updates `actions/checkout` to v2 (which is [faster](https://github.com/actions/checkout/blob/main/CHANGELOG.md#v2-beta))
* switches to `erlef/setup-elixir` since [`actions/setup-elixir` is archived](https://github.com/actions/setup-elixir#setup-elixir),
* unifies Elixir setup by removing `container` uses,
* adds OTP 23 and Elixir 1.11 to the test matrix,
* drops OTP 20 from the test matrix (Elixir 1.10 doesn't support it and it's 4 years old),
* removes unnecessary steps like installing Hex/Rebar.

Let me know what you think :raised_hands:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,15 +12,14 @@ jobs:
   check_format:
     runs-on: ubuntu-latest
     name: Check format
-    container:
-      image: elixir:1.9-slim
     steps:
-      - uses: actions/checkout@v1
-      - name: Install Dependencies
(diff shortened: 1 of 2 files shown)


polvalente:
#224 touches a lot of the stuff covered here, so I'm closing this in favor of that PR

Thanks!

PR #195, opened 2021-05-03
Allow to use iolists in GRPC.Message.to_data/2

bszaf:
This is small change that allows to use iolists returned from `codec` modules with `GRPC.Message.to_data/2` function.


lib/grpc/message.ex
```diff
@@ -29,9 +29,10 @@ defmodule GRPC.Message do
       {:error, "Encoded message is too large (9 bytes)"}
   """
```
polvalente, line 30: We should document the new option with a doctest

polvalente: After that, this looks ready to merge

PR #197, opened 2021-06-18
Provide a mechanism to capture exceptions during request handling

freshtonic:
Interceptors are insufficient for capturing errors in the general case
because they execute *after* a request has already been successfully
decoded.

This commit introduces a new concept called an ExceptionSink. It is a
behavior that exposes an `error/2` function.

ExceptionSinks are attached to an Endpoint like this:

```elixir
defmodule EndpointWithExceptionSink do
  use GRPC.Endpoint

  exception_sink ExceptionSink1
  exception_sink ExceptionSink2, foo: 1

  run Server
end
```

An ExceptionSink receives the raised exception and a stack trace,
allowing the developer to process/forward that information to the
observability tool of their choice.

Exceptions that are caught before sending to the exception sinks are
reraised so that the existing behaviour is preserved.

### DIFF
--- a/lib/grpc/server.ex
+++ b/lib/grpc/server.ex
@@ -107,16 +107,23 @@ defmodule GRPC.Server do
          %{request_mod: req_mod, codec: codec, adapter: adapter, payload: payload} = stream,
          func_name
        ) do
-    {:ok, data} = adapter.read_body(payload)
 
-    case GRPC.Message.from_data(stream, data) do
-      {:ok, message} ->
-        request = codec.decode(message, req_mod)
+    try do
+      {:ok, data} = adapter.read_body(payload)
 
-        call_with_interceptors(res_stream, func_name, stream, request)
+      case GRPC.Message.from_data(stream, data) do
+        {:ok, message} ->
+          request = codec.decode(message, req_mod)
 
-      resp = {:error, _} ->
-        resp
+          call_with_interceptors(res_stream, func_name, stream, request)
+
+        resp = {:error, _} ->
+          resp
+      end
+    rescue
+      err ->
+        report_to_exception_sinks(stream, err, __STACKTRACE__)
+        reraise err, __STACKTRACE__
     end
   end
 
@@ -199,6 +206,10 @@ defmodule GRPC.Server do
         Map.get(endpoint.__meta__(:server_interceptors), server, [])
 
     interceptors |> Enum.reverse()
+  end
+
+  defp exception_sinks(endpoint) do
+    endpoint.__meta__(:exception_sinks) |> Enum.reverse()
   end
 
   # Start the gRPC server. Only used in starting a server manually using `GRPC.Server.start(servers)`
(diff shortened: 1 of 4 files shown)


lessless:
@tony612 can you please at this PR? this is a heavily needed feature and will make debugging much simpler. thanks

lib/grpc/server.ex
```diff
@@ -107,16 +107,23 @@ defmodule GRPC.Server do
...
-        call_with_interceptors(res_stream, func_name, stream, request)
+          call_with_interceptors(res_stream, func_name, stream, request)
 
-      resp = {:error, _} ->
-        resp
+        resp = {:error, _} ->
+          resp
+      end
+    rescue
+      err ->
```
polvalente, line 124: I would use:

```elixir

catch
x -> ... # handle throw
 x, y -> ... # handle erlang error and raise
 ...
 end
 ```

Instead of rescue so this also captures `:erlang.error`s and `throws`:

```elixir
iex(3)> try do               
...(3)> :erlang.error("asdf")
...(3)> catch
...(3)> x, y -> {:ok, x, y}
...(3)> end
{:ok, :error, "asdf"}
iex(4)> try do               
...(4)> raise "asdf"
...(4)> catch              
...(4)> x, y -> {:ok, x, y}
...(4)> end
{:ok, :error, %RuntimeError{message: "asdf"}}
iex(5)> try do             
...(5)> throw "asdf"       
...(5)> catch
...(5)> x -> {:ok, x}
...(5)> end
{:ok, "asdf"}
```

polvalente: Other than this, the overall concept looks good. I'm not sure on the name, though, given we handle exceptions, erlang errors and throws

polvalente: Also, "sink" doesn't seem like the best suffix because it re-raises the exception (and would probably re-throw or re-execute the error)

polvalente: @freshtonic @lessless what do you think?

lessless: I would go for the as complete set of data as possible, so suggestion to capture `:erlang.errors` and `throws` is an absolute 👍

polvalente: @freshtonic @lessless I'm wondering if this feature is still gonna be relevant given the telemetry support being added in #229. Wouldn't it be the case that one would just attach to the relevant client/server exception events and filter according to the %Stream{} struct attached in the metadata?

tony612:
It looks good overall. I just wonder if we can implement this feature with only interceptors?

lib/grpc/server.ex
```diff
@@ -107,16 +107,23 @@ defmodule GRPC.Server do
...
+        report_to_exception_sinks(stream, err, __STACKTRACE__)
```
polvalente, line 125: I wonder if we still need this feature given that we now have specific telemetry events that capture the exact data that would be reported to `exception_sinks`. @freshtonic thoughts on this?

PR #198, opened 2021-06-29
Add support for `request_id` in `GRPC.Server.Stream`

ilyashuma:
The small change adds a new field `request_id` to the `GRPC.Server.Stream` struct. It may be handy for logging purposes.


lib/grpc/logger.ex
```diff
@@ -22,6 +22,7 @@ defmodule GRPC.Logger.Server do
 
   def call(req, stream, next, level) do
     if Logger.compare_levels(level, Logger.level()) != :lt do
+      Logger.metadata(request_id: stream.request_id)
```
polvalente, line 25: I would only override the key if it's not present. It's a really common key and as it is done here, can be nil-ified if `stream.request_id` is `nil`
```suggestion
      Logger.metadata(request_id: Logger.metadata()[:request_id] || stream.request_id)
```

lib/grpc/server.ex
```diff
@@ -88,11 +88,22 @@ defmodule GRPC.Server do
...
   end
 
+  defp generate_request_id do
+    binary = <<
+      System.system_time(:nanosecond)::64,
+      :erlang.phash2({node(), self()}, 16_777_216)::24,
+      :erlang.unique_integer()::32
+    >>
+
+    Base.url_encode64(binary)
```
polvalente, line 104:
```suggestion
    Base.url_encode64(binary, padding: false)
```

lib/grpc/server/stream.ex
```diff
@@ -25,6 +25,7 @@ defmodule GRPC.Server.Stream do
           endpoint: atom,
           rpc: tuple,
           request_mod: atom,
+          request_id: String.t(),
```
polvalente, line 28:
```suggestion
          request_id: String.t() | nil,
```

polvalente:
@ilyashuma let me know if you want me to take over these changes

ilyashuma:
Hi, @polvalente, yup, go ahead. Thank you!

Issue #199, opened 2021-07-01
Support more HTTP2 clients


dbernheisel:
**Is your feature request related to a problem? Please describe.**
Gun is great, but there are other maintained and proven HTTP2 clients, such as [Mint](https://github.com/elixir-mint/mint) and those that build upon it like [Finch](https://hex.pm/packages/finch), [Mojito](https://hex.pm/packages/mojito), and [Tesla](https://hex.pm/packages/tesla)

**Describe the solution you'd like**
A behaviour defined for HTTP adapters. I see that there is generally one already but not documented with `@behaviour` and `@callback`.

**Describe alternatives you've considered**
n/a

**Additional context**
I'd like to use gRPC, but with HTTP2 clients I'm more familiar with and widely used.

I'll be happy to contribute a Mint client, but before I do wanted to check if the project is ready to commit to an interface.

wingyplus:
After read the stub source, the adapter behaviour will looks like this:

```
defmodule GRPC.Adapter do
  @moduledoc """
  HTTP client adapter for GRPC.
  """

  @callback connect(GRPC.Channel.t(), any) :: {:ok, GRPC.Channel.t()} | {:error, any}

  @callback disconnect(GRPC.Channel.t()) :: {:ok, GRPC.Channel.t()} | {:error, any}

  @callback send_request(GRPC.Client.Stream.t(), binary, map) :: GRPC.Client.Stream.t()

  @callback send_headers(GRPC.Client.Stream.t(), any) :: GRPC.Client.Stream.t()

  @callback end_stream(GRPC.Client.Stream.t()) :: GRPC.Client.Stream.t()

  @callback cancel(any, GRPC.Client.Stream.stream_payload()) :: :ok | {:error, any}
end
```

I'm a bit interesting in implement adapter in Mint or Finch. And I think we can do on external repository. :D

polvalente:
Can I work on making the HTTP Client a configurable behaviour/implementation instead of a client lock-in?
We could provide a Mint and a Gun adapter by default (Gun being for retrocompatibility), but people would be able to choose other ones if they wanted :)

yordis:
> Can I work on making the HTTP Client a configurable behaviour/implementation instead of a client lock-in?

Strongly encourage you to make things better for the ecosystem.

You can learn from it and (maybe, we never know if the core contributors will take it 🤷🏻 ) give us all a greater experience.

So please do!

polvalente:
[...]

PR #200, opened 2021-07-07
Define behaviours for client and server adapters

avillen:
Hi 👋 

To give some context, I was trying to add tests to my integration using `grpc`. I found that you can set the adapter that you want on both client and server, and that makes it easier to test my code. I'm a big fan of `Mox`, I really think that simplifies the way to add tests, but we need a `behaviour` over we can create the mock.

This MR aims to create behaviours for both clients and servers, but for my personal use-case, with the `client` one would be enough. With this, an incomplete test example could be:

```
      expected_proto_response = ProtoSampe.new()

      grpc_channel = %GRPC.Channel{
        adapter: GRPCCLientMock
      }

      stream = %GRPC.Client.Stream{
        channel: grpc_channel,
        codec: GRPC.Codec.Proto,
        response_mod: ProtoSample
      }

      expect(GRPCCLientMock, :connect, fn _, _ -> {:ok, grpc_channel} end)
      expect(GRPCCLientMock, :send_request, fn _, _, _ -> stream end)
      expect(GRPCCLientMock, :recv_headers, fn _, _, _ -> {:ok, [], :fin} end)

      {:ok, data, _} =
        ProtoSample
        |> GRPC.Message.Protobuf.encode(expected_proto_response)
        |> GRPC.Message.to_data()

      expect(GRPCCLientMock, :recv_data_or_trailers, fn _, _, _ -> {:data, data} end)

      expect(GRPCCLientMock, :recv_data_or_trailers, fn _, _, _ ->
        {:trailers, %{"grpc-status" => "0"}}
      end)
      
     // Your request
```

I'm not a person with a lot of knowledge of both `grpc` and this library itself. This MR could be considered in WIP because and not pretty sure if I have properly defined the `typespecs`, or be closed without any problem if you consider.

Cheers and thanks for the library! Awesome job!


lib/grpc/adapter/client.ex
```diff
@@ -0,0 +1,27 @@
...
+  alias GRPC.Channel
+
+  @type opts :: any
+
+  @type channel :: Channel.t()
+  @callback connect(channel, opts) :: {:ok, channel} | {:error, any}
+  
+  @callback disconnect(channel) :: {:ok, channel} | {:error, any}
+
+  @type stream :: Stream.t()
```
polvalente, line 14: I would use `Stream.t()` directly instead of adding this proxy type. Is there any reason for this path?

lib/grpc/adapter/client.ex
```diff
@@ -0,0 +1,27 @@
+defmodule GRPC.ClientAdapter do
+  @moduledoc false
+
+  alias GRPC.Client.Stream
+  alias GRPC.Channel
+
+  @type opts :: any
```
polvalente, line 7: I think it's better to use `opts :: keyword()` instead (perhaps even avoiding a custom type, inlining this wherever needed).

This would warrant some changes to the current implementation, but provides a better interface

lib/grpc/adapter/client.ex
polvalente: For clarity, I would also move all `@type` definitions to the top of the module

lib/grpc/adapter/client.ex
```diff
@@ -0,0 +1,27 @@
...
+  @type channel :: Channel.t()
+  @callback connect(channel, opts) :: {:ok, channel} | {:error, any}
+  
+  @callback disconnect(channel) :: {:ok, channel} | {:error, any}
+
+  @type stream :: Stream.t()
+  @type message :: binary()
+  @callback send_request(stream, message, opts) :: stream
+
+  @type headers :: [{String.t(), String.t()}]
```
polvalente, line 18: The current implementation would need to change, as I think it currently returns a string-map

lib/grpc/adapter/client.ex
```diff
@@ -0,0 +1,27 @@
...
+  @type fin :: :fin | :nofin
```
polvalente, line 19: A typedoc would be great here

lib/grpc/adapter/client.ex
```diff
@@ -0,0 +1,27 @@
+defmodule GRPC.ClientAdapter do
```
polvalente, line 1:
```suggestion
defmodule GRPC.Client do
```

I think `ClientAdapter` is kinda redundant since the "adapter" part is actually what implements the behaviour specified here

lib/grpc/adapter/client.ex
```diff
@@ -0,0 +1,27 @@
...
+  @moduledoc false
```
polvalente, line 2: Adding a moduledoc here is really important so that this module is discoverable in the docs

lib/grpc/adapter/server.ex
```diff
@@ -0,0 +1,24 @@
+defmodule GRPC.ServerAdapter do
+  @moduledoc false
```
polvalente, line 2: Basically the same considerations as the client behaviour

polvalente (commented):
I started looking into this for issue #199.

Since there was already your pull request opened, perhaps it's best we focus on it instead!

polvalente:
Also, disclaimer: I'm not a maintainer for the library and have just added my 2 cents because I'm interested in having this library work in production

sleipnir:
> Also, disclaimer: I'm not a maintainer for the library and have just added my 2 cents because I'm interested in having this library work in production

Hi Paulo, it's a pleasure to speak with you again.
I decided to comment since you gave me the cue.
Today this project doesn't seem to be being maintained by anyone, there are numerous requests for PRs and Issues open without any response. For that reason I created a fork of the project called Falco to try to continue, we are also trying to improve the performance of the library that has suffered a lot in the gRPCs benchmark tests.
If you want to take a look at the project https://github.com/eigr/falco

**Note, some of the PRs open here I already include in Falco. But we are not yet ready for production.I'm sharing this just for your knowledge.

polvalente:
> > Also, disclaimer: I'm not a maintainer for the library and have just added my 2 cents because I'm interested in having this library work in production
> 
> Hi Paulo, it's a pleasure to speak with you again.
> I decided to comment since you gave me the cue.
> Today this project doesn't seem to be being maintained by anyone, there are numerous requests for PRs and Issues open without any response. For that reason I created a fork of the project called Falco to try to continue, we are also trying to improve the performance of the library that has suffered a lot in the gRPCs benchmark tests.
> If you want to take a look at the project https://github.com/eigr/falco
> 
> **Note, some of the PRs open here I already include in Falco. But we are not yet ready for production.I'm sharing this just for your knowledge.

@sleipnir that's awesome to know! For sure I'll take a look into it!

polvalente:
@avillen let me know if you want me to take over this PR

avillen:
hi @polvalente!! thanks for your review :smile: I can take care of apply your suggestions, but where are you saying to apply those changes, on this repo or in this one https://github.com/eigr/falco?

polvalente:
> hi @polvalente!! thanks for your review :smile: I can take care of apply your suggestions, but where are you saying to apply those changes, on this repo or in this one https://github.com/eigr/falco?

Here! I'm a maintainer for the repo now.
The idea is to concentrate the efforts on this repo.

I'm in touch with the people at falco and at another fork as well, so we can bring the best changes from the forks here

avillen:
> > hi @polvalente!! thanks for your review 😄 I can take care of apply your suggestions, but where are you saying to apply those changes, on this repo or in this one https://github.com/eigr/falco?
> 
> Here! I'm a maintainer for the repo now. The idea is to concentrate the efforts on this repo.
> 
> I'm in touch with the people at falco and at another fork as well, so we can bring the best changes from the forks here

Nice to hear that! I'll take care of it along the week 😄

avillen (commented):
@polvalente I was reading your suggestions and I totally agree with them, I just have a few smaill questions

lib/grpc/adapter/client.ex
polvalente: I think `GRPC.Ports.Client/Server` might be a better naming pattern (from the Ports and Adapters design pattern), and then we could have `GRPC.Adapters.Client.Gun` or something like that

lib/grpc/adapter/client.ex
polvalente: Using the in-line type is better for readability:

```elixir
# Module defs
defmodule T do

@type t :: integer

end

defmodule M do
  @type custom :: T.t()
  @spec f(x :: T.t(), y :: custom) :: T.t()
  def f(x, y), do: x + y
end

# iex calls
iex(1)> t M  
@type custom() :: T.t()

iex(2)> h M.f

                                  def f(x, y)                                   

  @spec f(x :: T.t(), y :: custom()) :: T.t()
```

When exploring the API via IEx, the proxy type will force the user to call `t module` to get the typedefs and discover it is indeed a proxy

lib/grpc/adapter/client.ex
wingyplus: Just my opinion, curently we have `lib/grpc/client` and `lib/grpc/server` which provides client code and server code separately. How about we move the adapter behaviour to inside those folders? Like:

```
# For server adapter behaviour.
lib/grpc/server/adapter.ex
# For client adapter behaviour.
lib/grpc/client/adapter.ex
```

polvalente: I tend to prefer grouping all behaviours under a given namespace to make it easy to find all of them (so adapters.xxx), but I'm not set too hard on either option :)

wingyplus: I’m ok with either option. :)

BTW It may apply to interceptor code to make it to the same pattern.

avillen: I was thinking about grouping the code by context, something like this:

- /client/
  - stream.ex
  - adapter.ex
  - adapter/
    - gun.ex
- /server.ex
- /server/
  - stream.ex
  - supervisor.ex
  - adapter.ex
  - adapter/
    - cowboy.ex
    - cowboy/
      - handler.ex

WDYT? You guys have a bigger picture and will work on the project more than me 😅 , so let me know which one suits better to you 😄

BTW, @polvalente if you are in a rush to merge this, please, go ahead and take the PR if you want!! In the end, I can only take this from time to time...

polvalente: I agree with the suggested tree, but I would have the `adapters` directory instead of `adapter`

avillen: > I agree with the suggested tree, but I would have the `adapters` directory instead of `adapter`

Nice! Agree with that! I'll push a commit with that ASAP 😄

config/config.exs
```diff
@@ -1,3 +1,3 @@
 use Mix.Config
 
-config :grpc, http2_client_adapter: GRPC.Adapter.Gun
+config :grpc, http2_client_adapter: GRPC.Client.Adapters.Gun
```
polvalente, line 3: I like how this looks :)

We need to deprecate the need for application env config, though, but that's another issue

lib/grpc/server.ex
```diff
@@ -214,21 +218,21 @@ defmodule GRPC.Server do
...
   end
 
   @doc false
-  @spec start_endpoint(atom, non_neg_integer, Keyword.t()) :: {atom, any, non_neg_integer}
+  @spec start_endpoint(endpoint, server_port, Keyword.t()) :: {atom, any, non_neg_integer}
   def start_endpoint(endpoint, port, opts \\ []) do
     servers = endpoint.__meta__(:servers)
     servers = GRPC.Server.servers_to_map(servers)
-    adapter = Keyword.get(opts, :adapter, GRPC.Adapter.Cowboy)
+    adapter = Keyword.get(opts, :adapter, GRPC.Server.Adapters.Cowboy)
```
polvalente, line 235: Could you replace all of these Keyword.get(a, b, c) with Keyword.get(a, b) || c?

polvalente: Just the ones this PR touches is fine, though

lib/grpc/server.ex
```diff
@@ -77,11 +77,15 @@ defmodule GRPC.Server do
     end
   end
 
-  @type servers_map :: %{String.t() => [module]}
+  @type endpoint :: Adapter.endpoint()
+  @type server_port :: Adapter.server_port()
   @type servers_list :: module | [module]
+  @type servers_map :: Adapter.servers_map()
+  @type stream :: Adapter.stream()
+  @type headers :: Adapter.headers()
```
polvalente, line 85: let's avoid proxy types

lib/grpc/stub.ex
```diff
@@ -98,6 +98,8 @@ defmodule GRPC.Stub do
     end
   end
 
+  @type channel :: GRPC.ClientAdapter.channel()
```
polvalente, line 101: proxy type

lib/grpc/stub.ex
```diff
@@ -125,7 +127,7 @@ defmodule GRPC.Stub do
     * `:accepted_compressors` - tell servers accepted compressors, this can be used without `:compressor`
     * `:headers` - headers to attach to each request
   """
-  @spec connect(String.t(), Keyword.t()) :: {:ok, GRPC.Channel.t()} | {:error, any}
+  @spec connect(String.t(), Keyword.t()) :: {:ok, channel} | {:error, any}
```
polvalente, line 130: I think we should use GRPC.Channel.t() here as it was before

lib/grpc/client/adapter.ex
```diff
@@ -0,0 +1,27 @@
+defmodule GRPC.Client.Adapter do
+  @moduledoc false
+
+  alias GRPC.Client.Stream
+  alias GRPC.Channel
+
+  @type opts :: any
```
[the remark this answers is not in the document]
polvalente: I think currently the code passes maps as options. We need to rework both this and header maps to become keywords and lists of tuples instead

lib/grpc/server/adapter.ex
```diff
@@ -0,0 +1,24 @@
...
+
+  @callback stop(endpoint, servers_map) :: :ok | {:error, :not_found}
+
+  @type stream :: Stream.t()
+
+  @type state :: GRPC.Server.Adapters.Cowboy.Handler.state()
+  @type reply :: binary()
+  @callback send_reply(state, reply, opts) :: any()
+
+  @type headers :: map()
```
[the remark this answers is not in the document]
avillen: I'd say yes! I can take care of it 😄

avillen:
@polvalente @wingyplus I think I have address all your suggestions, but pretty sure I'm missing something 😅 Do you mind taking a look again? Those missing things I think that we can take care of it in another PR:
https://github.com/elixir-grpc/grpc/pull/200#discussion_r924902211
https://github.com/elixir-grpc/grpc/pull/200#discussion_r924900104

lib/grpc/server/adapter.ex
```diff
@@ -0,0 +1,16 @@
+defmodule GRPC.Server.Adapter do
+  @moduledoc """
+  HTTP server adapter for GRPC.
+  """
+
+  alias GRPC.Server.Adapters.Cowboy.Handler
+
+  @callback start(atom, %{String.t() => [module]}, non_neg_integer(), Keyword.t()) ::
```
wingyplus, line 8: Should be atom() instead of just atom?

wingyplus: module might be the same

wingyplus:
```suggestion
  @callback start(atom(), %{String.t() => [module()]}, non_neg_integer(), Keyword.t()) ::
```

lib/grpc/server/adapter.ex
```diff
@@ -0,0 +1,16 @@
...
+              {atom, any, non_neg_integer}
```
wingyplus, line 9:
```suggestion
              {atom(), any(), non_neg_integer()}
```

lib/grpc/server/adapter.ex
```diff
@@ -0,0 +1,16 @@
...
+
+  @callback stop(atom, %{String.t() => [module]}) :: :ok | {:error, :not_found}
```
wingyplus, line 11:
```suggestion
  @callback stop(atom(), %{String.t() => [module()]}) :: :ok | {:error, :not_found}
```

wingyplus:
Just a quick look from my mobile:

* Found type declaration inconsistency ( () vs no () ).
  * I already add code suggestion. You may take a look at it.
* I think we should add `@impl true` to callback function in cowboy adapter and removing the typespec from it.

lib/grpc/client/adapter.ex
```diff
@@ -0,0 +1,23 @@
...
+  @type fin :: :fin | :nofin
+
+  @callback connect(Channel.t(), map()) :: {:ok, Channel.t()} | {:error, any}
+
+  @callback disconnect(Channel.t()) :: {:ok, Channel.t()} | {:error, any}
+
+  @callback send_request(Stream.t(), binary(), map()) :: Stream.t()
+
+  @callback recv_headers(map(), map(), map()) ::
+              {:ok, %{String.t() => String.t()}, fin} | {:error, GRPC.RPCError.t()}
```
wingyplus, line 19: Is it better if we just declare fin type here instead of declare at the module level?

polvalente: For `fin` I think we should declare a `@type` because we can have a typedoc specifying what it is and what the values represent

lib/grpc/server/adapter.ex
polvalente: I think we can standardize on having parens :)

wingyplus: Agreed :)

avillen: Agree and done (at least everywhere I see 😂 ), but I think that we should think about how we can add a validation about that 😅

avillen:
> Just a quick look from my mobile:
> 
> * Found type declaration inconsistency ( () vs no () ).
>   
>   * I already add code suggestion, but not all places. You may take a look at it.
> * I think we should add `@impl true` to callback function in cowboy adapter and removing the typespec from it.

@wingyplus I think it is done 😄

lib/grpc/client/adapter.ex
```diff
@@ -0,0 +1,23 @@
+defmodule GRPC.Client.Adapter do
+  @moduledoc """
+  HTTP client adapter for GRPC.
+  """
+
+  alias GRPC.Client.Stream
+  alias GRPC.Channel
+
+  @typedoc "Determines if the headers are finished reading."
```
polvalente, line 9:
```suggestion
  @typedoc "Determines if the headers have finished being read."
```

lib/grpc/stub.ex
```diff
@@ -136,24 +136,21 @@ defmodule GRPC.Stub do
...
 
   def connect(host, port, opts) when is_integer(port) do
     adapter =
-      Keyword.get(
-        opts,
-        :adapter,
-        Application.get_env(:grpc, :http2_client_adapter, GRPC.Adapter.Gun)
-      )
+      Keyword.get(opts, :adapter) ||
+        Application.get_env(:grpc, :http2_client_adapter, GRPC.Client.Adapters.Gun)
```
polvalente, line 148: Oh nice, this way we don't need to remove the global config :)

PR #206, opened 2021-09-02
Adds support for grpc-web

### LINKED ISSUE #193: gRPC Web Support
**Describe the question**
This library support [gRPC Web](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md)?
If not, are there plans to support?
If there are no plans, is there a strategy that we could implement to make this work? What would it take to make this work?

drowzy:
Adds support for [grpc-web](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md#protocol-differences-vs-grpc-over-http2).

* Adds two new functions to the `GRPC.Codec` behaviour in order to base64 decode / encode the payload.

There's an example [repo](https://github.com/drowzy/helloworld-elixir-grpc-web) to see it in action.

closes #193

### DIFF
--- a/lib/grpc/codec.ex
+++ b/lib/grpc/codec.ex
@@ -8,7 +8,22 @@ defmodule GRPC.Codec do
   @callback encode(any) :: binary
   @callback decode(any, module :: atom) :: any
 
+  @doc """
+  This function is optionally invoked before the gRPC payload is transformed into a protobuf message.
+
+  This can be used to apply a transform over the gRPC message before decoding it. For instance grpc-web using the `application/grpc-web-text`
+  content type requires the message to be base64 encoded, so a server receving messages using grpc-web-text will be required to
+  do a Base64 decode on the payload before decoding the gRPC message.
+  """
   @callback unpack_from_channel(binary) :: binary
+
+  @doc """
+  This function is optionally invoked after the protobuf message has been transformed to a gRPC payload.
+
+  This can be used to apply a transform over the gRPC message before sending it.
+  For instance grpc-web using the `application/grpc-web-text` content type requires the message to be base64 encoded, so a server sending messages using grpc-web-text will be required to
+  do a Base64 encode on the payload before sending the gRPC message.
+  """
   @callback pack_for_channel(binary) :: binary
   @optional_callbacks unpack_from_channel: 1, pack_for_channel: 1
 end
--- a/lib/grpc/codec.ex
+++ b/lib/grpc/codec.ex
@@ -7,6 +7,8 @@ defmodule GRPC.Codec do
   @callback name() :: String.t()
   @callback encode(any) :: binary
   @callback decode(any, module :: atom) :: any
-  @callback prepare_decode(binary) :: binary
-  @callback pack_encoded(binary) :: binary
+
+  @callback unpack_from_channel(binary) :: binary
+  @callback pack_for_channel(binary) :: binary
+  @optional_callbacks unpack_from_channel: 1, pack_for_channel: 1
 end
--- a/lib/grpc/codec.ex
+++ b/lib/grpc/codec.ex
@@ -24,6 +24,6 @@ defmodule GRPC.Codec do
   For instance grpc-web using the `application/grpc-web-text` content type requires the message to be Base64-encoded, so a server sending messages using grpc-web-text will be required to
   do a Base64 encode on the payload before sending the gRPC message.
   """
-  @callback pack_for_channel(binary) :: binary
+  @callback pack_for_channel(iodata()) :: binary
   @optional_callbacks unpack_from_channel: 1, pack_for_channel: 1
 end
--- a/lib/grpc/codec.ex
+++ b/lib/grpc/codec.ex
@@ -9,19 +9,19 @@ defmodule GRPC.Codec do
   @callback decode(any, module :: atom) :: any
 
   @doc """
-  This function is optionally invoked before the gRPC payload is transformed into a protobuf message.
+  This function is invoked before the gRPC payload is transformed into a protobuf message whenever it is defined.
 
   This can be used to apply a transform over the gRPC message before decoding it. For instance grpc-web using the `application/grpc-web-text`
-  content type requires the message to be base64 encoded, so a server receving messages using grpc-web-text will be required to
+  content type requires the message to be Base64-encoded, so a server receving messages using grpc-web-text will be required to
   do a Base64 decode on the payload before decoding the gRPC message.
   """
   @callback unpack_from_channel(binary) :: binary
 
   @doc """
-  This function is optionally invoked after the protobuf message has been transformed to a gRPC payload.
+  This function is invoked whenever it is defined after the protobuf message has been transformed into a gRPC payload.
 
   This can be used to apply a transform over the gRPC message before sending it.
-  For instance grpc-web using the `application/grpc-web-text` content type requires the message to be base64 encoded, so a server sending messages using grpc-web-text will be required to
+  For instance grpc-web using the `application/grpc-web-text` content type requires the message to be Base64-encoded, so a server sending messages using grpc-web-text will be required to
   do a Base64 encode on the payload before sending the gRPC message.
   """
   @callback pack_for_channel(binary) :: binary
--- a/lib/grpc/codec.ex
+++ b/lib/grpc/codec.ex
@@ -5,6 +5,8 @@ defmodule GRPC.Codec do
 
   @doc ~s(Name is identity of the codec, which will be suffix after content-type "application/grpc+" such as "proto".)
   @callback name() :: String.t()
-  @callback encode(any()) :: binary()
-  @callback decode(any(), module()) :: any()
+  @callback encode(any) :: binary
+  @callback decode(any, module :: atom) :: any
+  @callback prepare_decode(binary) :: binary
+  @callback pack_encoded(binary) :: binary
 end
--- a/lib/grpc/codec/web_text.ex
+++ b/lib/grpc/codec/web_text.ex
@@ -9,11 +9,11 @@ defmodule GRPC.Codec.WebText do
     Protobuf.Encoder.encode(struct)
   end
 
-  def pack_encoded(binary) do
+  def pack_for_channel(binary) do
     Base.encode64(binary)
   end
 
-  def prepare_decode(binary) do
+  def unpack_from_channel(binary) do
     Base.decode64!(binary)
   end
 
--- a/lib/grpc/codec/web_text.ex
+++ b/lib/grpc/codec/web_text.ex
@@ -9,6 +9,13 @@ defmodule GRPC.Codec.WebText do
     Protobuf.Encoder.encode(struct)
   end
 
+  def pack_for_channel(data) when is_list(data) do
+    data
+    |> IO.iodata_to_binary()
+    |> pack_for_channel()
+    |> List.wrap()
+  end
+
   def pack_for_channel(binary) do
     Base.encode64(binary)
   end
--- a/lib/grpc/codec/web_text.ex
+++ b/lib/grpc/codec/web_text.ex
@@ -13,7 +13,6 @@ defmodule GRPC.Codec.WebText do
     data
     |> IO.iodata_to_binary()
     |> pack_for_channel()
-    |> List.wrap()
   end
 
   def pack_for_channel(binary) do
--- a/lib/grpc/codec/web_text.ex
+++ b/lib/grpc/codec/web_text.ex
@@ -12,7 +12,7 @@ defmodule GRPC.Codec.WebText do
   def pack_for_channel(data) when is_list(data) do
     data
     |> IO.iodata_to_binary()
-    |> pack_for_channel()
+    |> Base.encode64()
   end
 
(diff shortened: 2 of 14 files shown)


sleipnir:
This is awesome

AleksandarFilipov:
Coming from an embedded perspective, being able to use this saves a lot of cpu cycles compared to my previous proxies.

Since the client now can be a web browser the library need to handle CORS. To get this operational I needed to do quite a lot of coors catchup resulting in the following changes from this https://github.com/elixir-grpc/grpc/blob/5b4c66c9d72d231c0be27745ca82ba9e2277b3b1/lib/grpc/transport/http2.ex#L11-L17

to:

```elixir 
  def server_headers(%{codec: GRPC.Codec.WebText = codec}) do
    Logger.error("Setting headers1")
    %{"content-type" => "application/grpc-web-#{codec.name}",
      "Access-Control-Allow-Origin" => "*",
      "Access-Control-Allow-Headers" => "Content-Type, x-grpc-web, x-user-agent",
    }
  end

  def server_headers(%{codec: codec}) do
    Logger.error("Setting headers2")
    %{"content-type" => "application/grpc+#{codec.name}",
     "Access-Control-Allow-Origin" => "*",
      "Access-Control-Allow-Headers" => "Content-Type, x-grpc-web, x-user-agent",
  }
  end
```
I'm assuming there is already a mechanism in place on this but I can't wrap my around it so please advice me. Some conclusions.
* cowboy can be configured for CORS but i don't see how to get pass the configuration properly.
* middleware is supported, possibly this can be managed using middleware, for some reason it doesn't feel as "the right way"

@drowzy Possibly you have the answer up your sleeve :)

sleipnir:
Any predictions for this to be approved? I think this library seems to have been abandoned by its creators

lib/grpc/adapter/cowboy/handler.ex
```diff
@@ -453,10 +453,16 @@ defmodule GRPC.Adapter.Cowboy.Handler do
...
   defp extract_subtype("application/grpc;"), do: {:ok, "proto"}
-
   defp extract_subtype(<<"application/grpc+", rest::binary>>), do: {:ok, rest}
   defp extract_subtype(<<"application/grpc;", rest::binary>>), do: {:ok, rest}
 
+  defp extract_subtype("application/grpc-web"), do: {:ok, "proto"}
+  defp extract_subtype("application/grpc-web+"), do: {:ok, "proto"}
+  defp extract_subtype("application/grpc-web;"), do: {:ok, "proto"}
+  defp extract_subtype("application/grpc-web-text"), do: {:ok, "text"}
+  defp extract_subtype(<<"application/grpc-web+", rest::binary>>), do: {:ok, rest}
```
polvalente, line 463:
```suggestion
  defp extract_subtype("application/grpc-web+" <> rest), do: {:ok, rest}
```

purely cosmetic :)

lib/grpc/adapter/cowboy/handler.ex
```diff
@@ -453,10 +453,16 @@ defmodule GRPC.Adapter.Cowboy.Handler do
...
+  defp extract_subtype(<<"application/grpc-web-text+", rest::binary>>), do: {:ok, rest}
```
polvalente, line 464:
```suggestion
  defp extract_subtype("application/grpc-web-text+" <> rest), do: {:ok, rest}
```

same as above

lib/grpc/codec.ex
```diff
@@ -7,4 +7,6 @@ defmodule GRPC.Codec do
   @callback name() :: String.t()
   @callback encode(any) :: binary
   @callback decode(any, module :: atom) :: any
+  @callback prepare_decode(binary) :: binary
```
polvalente, line 10: Not sure if prepare_decode is a good name. This is still part of the decoding process, right?

polvalente: Perhaps something more akin to "decode_channel" as in channel-encoding vs application-encoding?

I don't really have a better suggestion. If we can't find any, let's go with `prepare_decode`

polvalente: Perhaps since we have `pack_encoded`, we could have `pack_for_channel` and `unpack_from_channel`

polvalente: Also, I think a `@doc` would be desirable for both of the functions

polvalente: These could be optional callbacks and we could just check if they are defined by the codec instead of having to define no-op impls

polvalente (commented):
This looks great overall! I've left some comments so we can iterate a bit :)

AleksandarFilipov:
wonderful to see some activity here! Let's try and get some nice solution the the cors issue described above, possibly in a separate issue. 

Short recap, 
Adding this support suggested that web-client can access the services. Given the (cors) nature of web clients, they will likely be blocked if the accessing web-client is hosted separately. I've posted a solution, it caters for my needs, however it's not generic enough to be PR candidate. 

We need to be able to provide headers in a configurable way. I've played with Interceptors (for other purposes), however I don't see a clean match since it will affect all services - regardless if they use grpc-web or not.

polvalente:
@AleksandarFilipov I think we could use the interceptors in a similar way to how Tesla modifies the current response given the result of `next.(stream, req)`

https://github.com/elixir-tesla/tesla/blob/master/lib/tesla/middleware/compression.ex

In this example, they decide if the resp body needs to be decompressed based on the result tuple tag. We can provide a custom interceptor that injects response headers based on a predicate MFA tuple (that receives the stream and returns true if the headers should be injected). Something like:

```elixir
defmodule MyServer do
  use GRPC.Endpoint

  intercept GRPC.Interceptors.ResponseHeaders, should_inject: {__MODULE__, :"add_cors?", 1}, headers: [{"Access-Control-Allow-Origin", "*"}, {"Access-Control-Allow-Headers", "Content-Type, x-grpc-web, x-user-agent"}]

  def add_cors?(...), do: ...
end
```

The one thing that might need to change (I'm not _that_ familiar with the code base yet) is that this only works if `intercept`s are called before `stream_reply`. 

I'm almost sure that `intercepts` are only called after the reply, though. If so, we'll need to introduce something like `reply_intercept` that will be called right before the reply, but the overall idea remains the same.
This idea would allow us to remove the Codec from the `server_headers` match as well, while still allowing us to modify the headers (the whole connection, really, since as we see fit.

Shall we move this into a new issue?

lib/grpc/codec.ex
drowzy: Agree! Let's call them `pack_for_channel`, `unpack_for_channel`.

lib/grpc/message.ex
```diff
@@ -47,7 +48,13 @@ defmodule GRPC.Message do
       {:error, "Encoded message is too large (#{length} bytes)"}
     else
       result = <<compress_flag, length::size(4)-unit(8), message::binary>>
-      {:ok, result, length + 5}
+
+      data =
+        if not is_nil(codec) and function_exported?(codec, :pack_for_channel, 1),
```
polvalente, line 53:
```suggestion
        if function_exported?(codec, :pack_for_channel, 1),
```

We might have to deal with `Code.ensure_loaded!` here, but I think it's safe to do it just like this for now

lib/grpc/transport/http2.ex
```diff
@@ -8,8 +8,12 @@ defmodule GRPC.Transport.HTTP2 do
...
   require Logger
 
+  def server_headers(%{codec: GRPC.Codec.WebText = codec}) do
+    %{"content-type" => "application/grpc-web-#{codec.name()}"}
+  end
+
   def server_headers(%{codec: codec}) do
-    %{"content-type" => "application/grpc+#{codec.name}"}
+    %{"content-type" => "application/grpc+#{codec.name()}"}
   end
```
polvalente, line 17: How do we assert on these in tests? I see no tests that ensure this part of the feature is working

lib/grpc/codec.ex
```diff
@@ -7,4 +7,23 @@ defmodule GRPC.Codec do
   @callback name() :: String.t()
   @callback encode(any) :: binary
   @callback decode(any, module :: atom) :: any
+
+  @doc """
+  This function is optionally invoked before the gRPC payload is transformed into a protobuf message.
```
polvalente, line 12:
```suggestion
  This function is invoked before the gRPC payload is transformed into a protobuf message whenever it is defined.
```

lib/grpc/codec.ex
```diff
@@ -7,4 +7,23 @@ defmodule GRPC.Codec do
...
+
+  This can be used to apply a transform over the gRPC message before decoding it. For instance grpc-web using the `application/grpc-web-text`
+  content type requires the message to be base64 encoded, so a server receving messages using grpc-web-text will be required to
```
polvalente, line 15:
```suggestion
  content type requires the message to be Base64-encoded, so a server receving messages using grpc-web-text will be required to
```

lib/grpc/codec.ex
```diff
@@ -7,4 +7,23 @@ defmodule GRPC.Codec do
...
+  do a Base64 decode on the payload before decoding the gRPC message.
+  """
+  @callback unpack_from_channel(binary) :: binary
+
+  @doc """
+  This function is optionally invoked after the protobuf message has been transformed to a gRPC payload.
```
polvalente, line 21:
```suggestion
  This function is invoked whenever it is defined after the protobuf message has been transformed to a gRPC payload.
```

lib/grpc/codec.ex
```diff
@@ -7,4 +7,23 @@ defmodule GRPC.Codec do
...
+
+  This can be used to apply a transform over the gRPC message before sending it.
+  For instance grpc-web using the `application/grpc-web-text` content type requires the message to be base64 encoded, so a server sending messages using grpc-web-text will be required to
```
polvalente, line 24:
```suggestion
  For instance grpc-web using the `application/grpc-web-text` content type requires the message to be Base64-encoded, so a server sending messages using grpc-web-text will be required to
```

test/grpc/integration/codec_test.exs
```diff
@@ -20,32 +20,35 @@ defmodule GRPC.Integration.CodecTest do
   defmodule HelloServer do
     use GRPC.Server,
       service: Helloworld.Greeter.Service,
-      codecs: [GRPC.Codec.Proto, GRPC.Codec.Erlpack]
+      codecs: [GRPC.Codec.Proto, GRPC.Codec.Erlpack, GRPC.Codec.WebText]
```
polvalente, line 23: I think some tests need to be added here to ensure this does indeed make a difference (it puts the proper headers, etc)

lib/grpc/transport/http2.ex
drowzy: Added a test to ensure content types are set appropriately based on codec.

lib/grpc/message.ex
```diff
@@ -47,7 +48,13 @@ defmodule GRPC.Message do
       {:error, "Encoded message is too large (#{length} bytes)"}
     else
       result = <<compress_flag, length::size(4)-unit(8), message::binary>>
-      {:ok, result, length + 5}
+
+      data =
+        if function_exported?(codec, :pack_for_channel, 1),
```
polvalente, line 53: please update with master because this function has changed. We also need to ensure that the coded can handle the new possible iodata format instead of just binaries in pack_for_channel

polvalente: let me know if I should do this instead :)

drowzy: Updated to match master :). So `pack_for_channel` should accept `iodata` and return a `iolist` instead of a binary?
Or do you think the function should maintain the same shape as the input type? i.e `iolist -> iolist | binary -> binary`. Sort of:

```elixir
def pack_for_channel(data) when is_list(data) do
  data
  |> IO.iodata_to_binary()
  |> Base.encode64()
  |> List.wrap()
end
```

lib/grpc/codec/web_text.ex
```diff
@@ -0,0 +1,30 @@
...
+
+  def encode(struct) do
+    Protobuf.Encoder.encode(struct)
+  end
+
+  def pack_for_channel(data) when is_list(data) do
+    data
+    |> IO.iodata_to_binary()
+    |> pack_for_channel()
+    |> List.wrap()
```
polvalente, line 16:
```suggestion
```

We don't need this because a binary is also a valid iodata instance :)

lib/grpc/codec/web_text.ex
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  def encode(struct) do
+    Protobuf.Encoder.encode(struct)
+  end
+
+  def pack_for_channel(data) when is_list(data) do
+    data
+    |> IO.iodata_to_binary()
+    |> pack_for_channel()
```
polvalente, line 15:
```suggestion
    |> Base.encode64()
```

polvalente (approved):

lib/grpc/codec/web_text.ex
polvalente: https://hexdocs.pm/elixir/1.13/typespecs.html#built-in-types

sleipnir:
Awesome work guys!

PR #218, opened 2022-04-29
Allow RPC handlers to return errors

mpalmer:
Having to raise an exception to signal an error is slower (because
exceptions are, well, exceptional), and it also interferes with
exception logging interceptors.  This change allows RPC handler
methods to return {:error, ...} and have the expected thing happen --
which happens to be exactly how interceptors work, as a bonus.

### DIFF
--- a/lib/grpc/server.ex
+++ b/lib/grpc/server.ex
@@ -161,12 +161,16 @@ defmodule GRPC.Server do
          req
        ) do
     last = fn r, s ->
-      reply = apply(server, func_name, [r, s])
+      case apply(server, func_name, [r, s]) do
+        error = {:error, _} ->
+          error
 
-      if res_stream do
-        {:ok, stream}
-      else
-        {:ok, stream, reply}
+        reply ->
+          if res_stream do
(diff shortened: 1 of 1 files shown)


lib/grpc/server.ex
```diff
@@ -161,12 +161,16 @@ defmodule GRPC.Server do
          req
        ) do
     last = fn r, s ->
-      reply = apply(server, func_name, [r, s])
-
-      if res_stream do
-        {:ok, stream}
-      else
-        {:ok, stream, reply}
+      case apply(server, func_name, [r, s]) do
```
polvalente, line 164: This is interesting, but it is a breaking change. I think some tests are needed so we can move forward with this PR

polvalente:
@mpalmer do you plan on continue developing this PR?

mpalmer:
No, I have ceased to be involved in supporting a gRPC-using Elixir application.

polvalente:
I'll convert this to an issue then, so I can tackle this perhaps for v0.7! Thanks for the response :)

PR #231, opened 2022-07-19
fix: add @impl to logger interceptor callback

wingyplus:


lib/grpc/logger/client.ex
```diff
@@ -13,13 +11,19 @@ defmodule GRPC.Logger.Client do
...
   """
 
+  require Logger
+
+  @behaviour GRPC.ClientInterceptor
+
+  @impl true
   def init(opts) do
-    level = Keyword.get(opts, :level) || :info
+    level = opts[:level] || :info
```
polvalente, line 20:
```suggestion
    level = Keyword.get(opts, :level) || :info
```

I don't want to support maps here

lib/grpc/logger/server.ex
```diff
@@ -13,14 +13,18 @@ defmodule GRPC.Logger.Server do
...
   """
+
   require Logger
+
   @behaviour GRPC.ServerInterceptor
 
+  @impl true
   def init(opts) do
-    level = Keyword.get(opts, :level) || :info
+    level = opts[:level] || :info
```
polvalente, line 23:
```suggestion
    level = Keyword.get(opts, :level) || :info
```

PR #236, opened 2022-07-19
fix: avoid overriding user config if possible

### LINKED ISSUE #176: Unable to specify gun tcp_opts field when using gun v2
**Describe the bug**
When passing options for gun via the `:adapter_opts` key in the 3rd argument to the function `GRPC.Stub.connect/3`, if you specify the `tcp_opts` field (which should be of type `[gen_tcp:connect_option()]`, see: https://ninenines.eu/docs/en/gun/2.0/manual/gun/) the field is always replaced with the value of `@default_transport_opts`. 

**To Reproduce**
Ensure you are using gun v2 (you should be unless you specify a different version with `override: true` in your `mix.exs`) and make sure the `gun` application is loaded. Add some logging to `GRPC.Stub.connect_insecurely/2` or trace `:gun.open/3` using e.g. `recon_trace:calls`, and then call `GRPC.Stub.connect(my_addr, my_port, [adapter_opts: %{tcp_opts: my_tcp_opts}])` where `my_tcp_opts` is a proplist containing some options you want to give to gun's `tcp_opts` field.

**Expected behavior**
The options passed to `gun:open/3` should include the values you provided for the `tcp_opts` field. Instead you will find that you can only ever get the default tcp_opts value provided by the module attribute `@default_transport_opts` in the `GRPC.Adapter.Gun` module

**Logs**
```elixir
Interactive Elixir (1.10.1) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> :code.ensure_loaded(:gun)
{:module, :gun}
iex(2)> :recon_trace.calls({:gun, :open, :_}, 1)
2
iex(3)> grpc_opts = [adapter_opts: %{tcp_opts: [port: 6969, nodelay: true]}]
[adapter_opts: %{tcp_opts: [port: 6969, nodelay: true]}]
iex(4)> GRPC.Stub.connect("localhost", 50051, grpc_opts)
14:26:13.021314 <0.250.0> gun:open("localhost", 50051, #{http2_opts=>#{settings_timeout=>infinity}, protocols=>[http2], retry=>100, retry_fun=>fun 'Elixir.GRPC.Adapter.Gun':retry_fun/2, tcp_opts=>[{nodelay,true}], transport=>tcp})
Recon tracer rate limit tripped.
```

**Versions:**
- OS: NixOS 20.03
- Erlang: Erlang/OTP 22 [erts-10.5.5]
- Elixir: 1.10.1
- mix.lock(grpc, gun, cowboy, cowlib): 
 ```elixir
"grpc": {:git, "https://github.com/elixir-grpc/grpc.git", "9b8bfce9ba529805ac0bd6d183bf9563a9f24b0c", []}
 "cowboy": {:hex, :cowboy, "2.7.0", "91ed100138a764355f43316b1d23d7ff6bdb0de4ea618cb5d8677c93a7a2f115", [:rebar3], [{:cowlib, "~> 2.8.0", [hex: :cowlib, repo: "hexpm", optional: false]}, {:ranch, "~> 1.7.1", [hex: :ranch, repo: "hexpm", optional: false]}], "hexpm", "04fd8c6a39edc6aaa9c26123009200fc61f92a3a94f3178c527b70b767c6e605"},
  "cowlib": {:hex, :cowlib, "2.9.1", "61a6c7c50cf07fdd24b2f45b89500bb93b6686579b069a89f88cb211e1125c78", [:rebar3], [], "hexpm", "e4175dc240a70d996156160891e1c62238ede1729e45740bdd38064dad476170"},
  "gun": {:hex, :grpc_gun, "2.0.0", "f99678a2ab975e74372a756c86ec30a8384d3ac8a8b86c7ed6243ef4e61d2729", [:rebar3], [{:cowlib, "~> 2.8.0", [hex: :cowlib, repo: "hexpm", optional: false]}], "hexpm", "03dbbca1a9c604a0267a40ea1d69986225091acb822de0b2dbea21d5815e410b"}
```

**Additional context**
The problem is that on lines [27](https://github.com/elixir-grpc/grpc/blob/master/lib/grpc/adapter/gun.ex#L27) and [45](https://github.com/elixir-grpc/grpc/blob/master/lib/grpc/adapter/gun.ex#L45) of `GRPC.Adapter.Gun`, the call to `Map.merge/2` always replaces the value of `tcp_opts` in the first argument (which is the value supplied by the user) with the value in the second argument (the default value). 

Note that things work as expected if the user is using gun v1 and instead uses the key `transport_opts`. Do we really need the checks to the version of gun here? I think it makes sense to just pass through whichever options the user supplies, since gun itself will validate the options when `gun:open` is called anyway.

### PR DESCRIPTION
**Pull request by polvalente**
closes #176

### CODE REVIEW

**Comment by polvalente** [comment]
@wingyplus could you review this please?

**Comment by wingyplus** [comment]
I think it is good to go.

File: mix.exs [hunk 1]
```diff
@@ -40,9 +40,9 @@ defmodule GRPC.Mixfile do
       {:gun, "~> 2.0.0-rc.2"},
       {:cowlib, "~> 2.11"},
       {:protobuf, "~> 0.10", only: [:dev, :test]},
-      {:ex_doc, "~> 0.28", only: :dev},
-      {:inch_ex, "~> 2.0", only: [:dev, :test]},
-      {:dialyxir, "~> 1.1", only: [:dev, :test], runtime: false}
+      {:ex_doc, "~> 0.28.0", only: :dev},
+      {:inch_ex, "~> 2.0.0", only: [:dev, :test]},
+      {:dialyxir, "~> 1.1.0", only: [:dev, :test], runtime: false}
```
wingyplus, line 45: Why we change the version constraint?

polvalente: These are only for the lib itself (only prod deps are propagated to the client libs). 

Therefore, I made this more strict because I always prefer to be as strict as possible with dependencies

wingyplus: 👍

lib/grpc/client/adapters/gun.ex
```diff
@@ -1,61 +1,56 @@
...
-    open_opts = %{transport: :ssl, protocols: [:http2]}
+    transport_opts = Map.get(opts, :transport_opts) || []
 
-    open_opts =
-      if gun_v2?() do
-        Map.put(open_opts, :tls_opts, transport_opts)
-      else
-        Map.put(open_opts, :transport_opts, transport_opts)
-      end
+    tls_opts = Keyword.merge(@default_transport_opts ++ ssl, transport_opts)
```
wingyplus, line 21: Just quick check from mobile.

which ssl configuration (via cred vs via transport_opts) is higher priority?

polvalente: Via transport options

There's a test that shows that

wingyplus: Do we need to document GRPC.Stub.connect about this behaviour? Because the library encourage user to use GRPC.Credential to set ssl credential but now user can set it via adapter option when calling GRPC.Stub.connect.

polvalente: I don't think so. This is a per-adapter specification

polvalente: We can decide to not let some overrides take place, though.

Although I wouldn't invest too much energy in the Gun adapter since we're planning on moving on to Mint

wingyplus: So how about 

```elixir
Keyword.merge(@default_transport_opts, transport_opts)
|> Keyword.merge(ssl)
```

i think we should clarify this behaviour before add mint adapter.

wingyplus: BTW. Which way we choose, i ok with it. :)

wingyplus:
Overall, LGTM. :)

lib/grpc/client/adapters/gun.ex
polvalente: I prefer the way it's currently coded, in which the `opts` values have precedence. Could you point to where in the docs it encourages the SSL values to be used? I don't think we'll have to change those other than adding a "opts can override this" warning

wingyplus: Here it is https://github.com/elixir-grpc/grpc/blob/master/lib/grpc/credential.ex#L3


Agree that if we can warning the opts can be override.

polvalente: I'll improve on that :)

polvalente: Done :) I'm merging this, but let me know if things are still unclear

wingyplus: That’s totally clear :)

PR #243, opened 2022-07-22
ci: remove cron_ci.yml and .ci

wingyplus:
Remove cron_ci.yml since we already run interop_tests every commit that
merge to master branch. And remove .ci since it's no uses. In Makefile,
remove cron-ci task since it's no use also.


Makefile
```diff
@@ -18,10 +18,4 @@ test-all:
 	mix test
 	cd interop && mix run script/run.exs
 
-# This is heavy
-ci-cron:
-	cd interop && mix deps.get && mix run script/run.exs --rounds 1000 --concurrency 30 && cd -
-	mix deps.get && bash .ci/build-plt-cache.sh && mix dialyzer
```
wingyplus, line 24: We may have a ci job to run dialyzer every commit that merge to master branch?

polvalente (approved):

wingyplus:
@polvalente So fast. 😂

Makefile
polvalente: Yeah, we can add a cached dialyzer step :)

PR #244, opened 2022-07-22
refactor: revise helloworld example

wingyplus:
* Revise children in HelloworldApp to use a new way of
  declaring supervisor.
* Generate proto code with elixir-protobuf 0.10.
* Add minimum proto file (copied from test/support) to under
  proto folder.

### DIFF
--- a/examples/helloworld/lib/helloworld.pb.ex
+++ b/examples/helloworld/lib/helloworld.pb.ex
@@ -1,31 +1,25 @@
 defmodule Helloworld.HelloRequest do
-  use Protobuf
+  @moduledoc false
+  use Protobuf, protoc_gen_elixir_version: "0.10.0", syntax: :proto3
 
-  @type t :: %__MODULE__{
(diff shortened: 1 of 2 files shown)


polvalente (approved):

examples/helloworld/lib/helloworld.pb.ex
```diff
@@ -1,31 +1,25 @@
 defmodule Helloworld.HelloRequest do
-  use Protobuf
+  @moduledoc false
+  use Protobuf, protoc_gen_elixir_version: "0.10.0", syntax: :proto3
```
sleipnir, line 3: Did you add this manually or did protoc generate this via the protobuf plugin? I don't think it's cool that we have to change these files after they've been created if that's the case here

wingyplus: @sleipnir i generate it with protoc. This change added since plugin version 0.10

PR #245, opened 2022-07-22
refactor: revise route_guide example

wingyplus:
* Change supervisor declration in tree.
* Cleanup dependencies.
* Use jason instead of poison.
* Auto start server.
* Move route_guide.proto to priv/protos.


examples/route_guide/config/config.exs
```diff
@@ -1,3 +1,3 @@
 import Config
 
-# config :grpc, start_server: true
+config :grpc, start_server: true
```
polvalente, line 3: This is something I wish to refactor. IMO We should move to a "add X to your supervision tree" instead of this global config

polvalente (approved):

examples/route_guide/config/config.exs
wingyplus: You means like:

```
children = [
  {GRPC.Server.Supervisor, {Endpoint, 50000}, start_server: true}
]
```

polvalente: Exactly! I've opened #246 for this

polvalente: This would enable multiple servers to run from the same application in different ports

wingyplus: Great! I will change the example to use the new way after #246 get merge into master. 🙇‍♂️

PR #250, opened 2022-07-23
chore: resolve some doc warnings

### LINKED ISSUE #249: GRPC.Server.Adapters.Cowboy.Handler.state document is missing
CI complaint about Handler.state documentation is missing

FA19CA9F-26F5-420A-B11A-2D31C713471B

found from PR #236

polvalente:
closes #249


lib/grpc/server/supervisor.ex
```diff
@@ -48,7 +48,8 @@ defmodule GRPC.Server.Supervisor do
     check_deps_version()
 
     children =
-      if Application.get_env(:grpc, :start_server, false) do
+      if (Keyword.has_key?(opts, :start_server) and opts[:start_server]) or
+           Application.get_env(:grpc, :start_server) do
```
wingyplus, line 52: ❤️

polvalente: Just caught a slight bug on the logic here haha!

The has_key check is slightly off

wingyplus, line 52: What is plan to remove global configuration?

polvalente, line 52:
```suggestion
      if opts[:start_server] or (not Keyword.has_key?(opts, :start_server) and Application.get_env(:grpc, :start_server) do
```

polvalente: This change removes the need to deprecate the global config :)

If the key is present in the options, it will have precedence (regardless of the value)

wingyplus: Got it. 🙇

lib/grpc/server/supervisor.ex
```diff
@@ -48,7 +48,7 @@ defmodule GRPC.Server.Supervisor do
     check_deps_version()
 
     children =
-      if Application.get_env(:grpc, :start_server, false) do
+      if opts[:start_server] or (not Keyword.has_key?(opts, :start_server) and Application.get_env(:grpc, :start_server) do
```
wingyplus, line 51: End parenthesis is missing? 

```suggestion
      if opts[:start_server] or (not Keyword.has_key?(opts, :start_server) and Application.get_env(:grpc, :start_server)) do
```

wingyplus:
@polvalente you may need to add start_server to this line https://github.com/elixir-grpc/grpc/blob/master/interop/script/run.exs#L17 since interop test does not starting a server

polvalente:
> @polvalente you may need to add start_server to this line https://github.com/elixir-grpc/grpc/blob/master/interop/script/run.exs#L17 since interop test does not starting a server

The problem was that `nil or ...` doesn't work. It should have been `opts[:start_server] || ...`, but I refactored into a cond instead

wingyplus (approved):
LGTM

Issue #229, opened 2022-07-19
Telemetry support


wingyplus:
It would be good if have telemetry support for this library. I'm not sure if interceptor is a good fit but we can start from that.

polvalente:
I think some default interceptors would do just fine, although perhaps we can have events being published directly by the core server and client functions.

I would have something like `[:grpc, :server, ...]` and `[:grpc, :client, ...]` for the event names.
What do you propose we publish through telemetry? I think we should decide what events we want and then we can look at how to implement them

wingyplus:
I want to start something simple like execution time such as `[:grpc, :server, :start]`, `[:grpc, :server, :stop]`, `[:grpc, :client, :start]` and `[:grpc, :client, :stop]`. (Event prefix can be change). This could use `:telemetry.span` to wrap this process. 

Do you have any metric that you interest?

polvalente:
Those make sense, although I would use [:grpc, :server, :request] as the name of the event (to which telemetry span would append start and stop and exception).

I think we should include the stream as part of the metadata. I think it's basically a mixture of what Tesla and Phoenix do with Telemetry.

Since client-server connections aren't necessarily related to each request, we should perhaps add separate events for connect and disconnect on both sides as well

polvalente:
@wingyplus to be clear, I think we're ready for a PR on these events :)
Ideally we don't do it using the adapters so that we can ensure the events are published regardless of the adapter used

wingyplus:
Yeah. I’ll looking into it after setup ci is done. :)

polvalente:
As a point of reference, I'll work to publish events so we can export all metrics that grpc_prometheus currently exports

PR #252, opened 2022-07-23
feat: implement telemetry interceptor on client side

wingyplus:
This is a draft PR to add telemetry to the client side. It's just interceptor that call span that capture `next.(stream, req)`. Currently, the interceptor attach `stream` to metadata when span is start, and add `reply` when the span stop. I've modify the helloworld example to show how implement logger with telemetry, its print all event and information of that event. 

Screen Shot 2565-07-23 at 20 59 08

Please note that it support only unary operation call. The event metadata can be change during discussion. 🙇 

This is work in progress since it's lack of unit tests.

Updates #229

### DIFF
--- a/lib/grpc/telemetry/client.ex
+++ b/lib/grpc/telemetry/client.ex
@@ -1,0 +1,36 @@
+defmodule GRPC.Telemetry.Client do
+  @moduledoc """
+  Telemetry interceptor for GRPC client.
+
+  ## Events
+
+  TBD.
+  """
+
+  @behaviour GRPC.ClientInterceptor
+
+  @prefix [:grpc, :client, :request]
+  @default_metadata %{}
+
+  @impl true
+  def init(opts) do
+    # TODO: clarify opts for this interceptor.
+    opts
+  end
+
+  @impl true
+  def call(%{grpc_type: :unary} = stream, req, next, _opts) do
+    # TODO: attach stream information to metadata.
+    metadata = Map.merge(@default_metadata, %{stream: stream})
+
+    :telemetry.span(@prefix, metadata, fn ->
+      reply = next.(stream, req)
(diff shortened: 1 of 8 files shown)


lib/grpc/telemetry/client.ex
```diff
@@ -0,0 +1,36 @@
+defmodule GRPC.Telemetry.Client do
+  @moduledoc """
+  Telemetry interceptor for GRPC client.
```
polvalente, line 3: I don't think we should go the interceptor route for telemetry. I think we should add the span as a wrapper for the actual adapter.send_request call.

Also, perhaps we should use execute instead of span, so we can add :headers_received and other intermediate steps instead of just :start and :stop

wingyplus: I see. Let’s try

wingyplus: > I don't think we should go the interceptor route for telemetry. I think we should add the span as a wrapper for the actual adapter.send_request call.
> 
> Also, perhaps we should use execute instead of span, so we can add :headers_received and other intermediate steps instead of just :start and :stop

What kind of value you expected from :headers_received? Is it headers when call recv_header from adapter?

polvalente: We can add the time that a given request took so far. If we introduce the request id as a first-class metadatum for the metrics, a monitoring system can easily show how much time it took to process each step (before headers, processing headers, total time since the beginning...)

lib/grpc/telemetry/client.ex
```diff
@@ -0,0 +1,36 @@
+defmodule GRPC.Telemetry.Client do
```
wingyplus, line 1: We'll not continue on this route.

PR #254, opened 2022-07-27
Server Reflection support

sleipnir:
This adds Server Reflection support


lib/grpc/endpoint.ex
```diff
@@ -94,12 +94,29 @@ defmodule GRPC.Endpoint do
...
     |> List.flatten()
   end
 
+  defp parse_servers(servers, true) do
+    servers
+    |> Enum.map(fn {server, _} -> server end)
+    |> List.flatten()
+  end
+
+  defp has_reflection_disabled?() do
```
beligante, line 111: IMO, I think it's more common/easy to read if you check for the feature to be enabled. 
Also, the convention in elixir is to have the bolean funcs suffixed with `?` and guards using `is_something`. Thoughts?
```suggestion
  defp reflection_enabled?() do
```

lib/grpc/endpoint.ex
```diff
@@ -94,12 +94,29 @@ defmodule GRPC.Endpoint do
...
     servers
     |> Enum.map(fn {server, _} -> server end)
     |> List.flatten()
   end
 
+  defp parse_servers(servers, true) do
+    servers
+    |> Enum.map(fn {server, _} -> server end)
+    |> List.flatten()
+  end
```
beligante, line 109: To avoid duplication, WDYT?
```suggestion
  defp parse_servers(servers) do
    case reflection_enabled?() do
      true -> [{GRPC.Reflection.Service, %{}} | servers]
      false -> servers
    end
    |> Enum.map(fn {server, _} -> server end)
    |> List.flatten()
  end
```

beligante (commented):
Good Job! Looks like a promising feature. Looking forward to see more from it

lib/grpc/endpoint.ex
```diff
@@ -94,12 +94,29 @@ defmodule GRPC.Endpoint do
...
+
+  defp has_reflection_disabled?() do
+    Application.get_env(:grpc, :reflection_disabled, false) ||
+      convert!(System.get_env("GRPC_REFLECTION_DISABLED"))
```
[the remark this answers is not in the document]
sleipnir: I agree

sleipnir:
Thanks for review @beligante 
I hope to be able to come back soon to develop this feature. Just a little busy yet to be able to continue, I'll soon have time to get on with it.
Thanks for your great suggestions

.gitignore
```diff
@@ -25,3 +25,5 @@ erl_crash.dump
 .elixir_ls
 
 grpc-*.tar
+
+.vscode
```
polvalente, line 29: this should be in you global gitignore instead

lib/grpc/endpoint.ex
```diff
@@ -95,11 +95,23 @@ defmodule GRPC.Endpoint do
   end
 
   defp parse_servers(servers) do
-    servers
+    case reflection_enabled?() do
+      true -> [{GRPC.Reflection.Service, %{}} | servers]
+      false -> servers
+    end
```
polvalente, line 101: if-else

lib/grpc/endpoint.ex
```diff
@@ -95,11 +95,23 @@ defmodule GRPC.Endpoint do
...
     |> Enum.map(fn {server, _} -> server end)
     |> List.flatten()
```
polvalente, line 103: Isn't this a flat list given you already extracted `server`?

lib/grpc/endpoint.ex
```diff
@@ -95,11 +95,23 @@ defmodule GRPC.Endpoint do
...
   end
 
+  defp reflection_enabled?() do
+    Application.get_env(:grpc, :reflection_disabled, false) ||
```
polvalente, line 107: avoid global configuration. This should be configured through `opts` instead

lib/grpc/endpoint.ex
```diff
@@ -95,11 +95,23 @@ defmodule GRPC.Endpoint do
...
+      convert!(System.get_env("GRPC_REFLECTION_DISABLED"))
+  end
+
+  defp convert!("true"), do: true
+  defp convert!("false"), do: false
+  defp convert!(op) when is_nil(op), do: false
```
polvalente, line 113: This kind of conversion is best done close to the config files. However, I don't think we'll move forward with this

lib/grpc/endpoint.ex
polvalente: also, let's make it opt-in instead

lib/grpc/reflection/reflection_service.ex
```diff
@@ -0,0 +1,45 @@
...
+  use GRPC.Server, service: Grpc.Reflection.V1alpha.ServerReflection.Service
+  require Logger
+
+  alias GRPC.Reflection
+  alias Grpc.Reflection.V1alpha.{ServerReflectionRequest, ServerReflectionResponse, ErrorResponse}
+
+  alias GRPC.Server
+
+  @spec server_reflection_info(ServerReflectionRequest.t(), GRPC.Server.Stream.t()) ::
+          ServerReflectionResponse.t()
```
polvalente, line 14: this currently only returns `:ok` instead of that custom type

lib/grpc/reflection/v1alpha/reflection.pb.ex
```diff
@@ -0,0 +1,586 @@
+defmodule Grpc.Reflection.V1alpha.ServerReflectionRequest do
```
polvalente, line 1:
```suggestion
defmodule Grpc.Reflection.ServerReflectionRequest do
```

Why add a version to the name? If we decide to keep that, we should perhaps use just `V1` instead. Remember to change the file path as well

lib/grpc/reflection/v1alpha/reflection.pb.ex
```diff
@@ -0,0 +1,586 @@
...
+  @moduledoc false
+  use Protobuf, protoc_gen_elixir_version: "0.10.0", syntax: :proto3
+
+  def descriptor do
+    # credo:disable-for-next-line
```
polvalente, line 6: Was this autogenerated? We don't have credo in this project

lib/grpc/reflection/v1alpha/reflection.pb.ex
sleipnir: because this is generated based on the protobuf that comes directly from google's protobufs schemas. I find it interesting to keep since this represents a formal specification

lib/grpc/reflection/v1alpha/reflection.pb.ex
sleipnir: Don't worry like I said before I'm just bringing what makes sense in the Falco environment here first and then refactoring to what makes sense in elixir-grpc. This will come out in the final version.

sleipnir: But I didn't look at it in depth and if this code is auto generated then the correct thing in my view would be not to mess with it since any build would bring back the comment regardless of having the creed or not. I will take a look and see what it is and what this comment is for more calmly when I resume PR

sleipnir:
Thanks for the suggestions @polvalente

lib/grpc/endpoint.ex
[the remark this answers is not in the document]
polvalente: IMO we should avoid relying on any sort of global application environment

polvalente:
@sleipnir what's the status on this PR?

sleipnir:
> @sleipnir what's the status on this PR?

Unfortunately I couldn't finish the implementation yet. But I should have availability in a couple of weeks. Two or less.

conradwt:
Are there any updates for this feature support?

sleipnir:
> Are there any updates for this feature support?

I've been very busy at work and my personal project for the last few months, so I haven't been able to pay attention to this.
But now I'm going to start a quieter phase and I believe I'll be able to look at it this week.

PR #259, opened 2022-08-01
feat: add Mint adapter

### LINKED ISSUE #242: Add Mint client adapter
After #199, we now have a contract defined for the HTTP client, so we can now implement an adapter using the Mint library and use that as a default instead of the Gun adapter

polvalente:
closes #242


polvalente:
Still really WIP. I'm basing the work on the client [as suggested here](https://github.com/elixir-grpc/grpc/issues/242#issuecomment-1191801711), but since there's lots of process wrapping around it, perhaps this is not the best path.

polvalente:
Someone has pinged me regarding taking this on, so this PR will be put on hold temporarily. I'll take this over if something else happens

PR #262, opened 2022-08-05
Feat/timestamp example

angelo-moreira:
An example of how to work with Google Timestamp Protobuf type.


examples/helloworld/lib/server.ex
```diff
@@ -4,6 +4,12 @@ defmodule Helloworld.Greeter.Server do
   @spec say_hello(Helloworld.HelloRequest.t(), GRPC.Server.Stream.t()) ::
           Helloworld.HelloReply.t()
   def say_hello(request, _stream) do
-    Helloworld.HelloReply.new(message: "Hello #{request.name}")
+    today = DateTime.utc_now()
+    seconds = today |> DateTime.truncate(:second) |> DateTime.to_unix(:second)
+
+    Helloworld.HelloReply.new(
+      message: "Hello #{request.name}",
+      today: %Google.Protobuf.Timestamp{seconds: seconds}
```
sleipnir, line 12: I would like to make a small suggestion.

Generally, when we want to convey the idea of time, we do it in a way that the client receiving the message can decode the values in the best possible way. The timestamp type was designed for these cases and for this reason it has attributes to show the time both in seconds and in nano. As we see in the original proto documentation:

https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/timestamp.proto:
```java
 Example 4: Compute Timestamp from Java `System.currentTimeMillis()`.

     long millis = System.currentTimeMillis();

     Timestamp timestamp = Timestamp.newBuilder().setSeconds(millis / 1000)
         .setNanos((int) ((millis % 1000) * 1000000)).build();
```
Note that in the original examples both attributes (seconds and nanos) are demonstrated.

I think it's interesting to add the values in nanos too since as an example it should be as complete as possible.

sleipnir (changes requested):

examples/helloworld/lib/server.ex
sleipnir: Hi @angelo-moreira I believe that the correct thing in Elixir to give us the Epoch Timestamp in nanoseconds would be if we used ```System.monotonic_time(:nanosecond)```.

sleipnir: Once we have the time in nano we can convert it to seconds using ```time |> System.convert_time_unit(:nanosecond, :second)```

polvalente: The rem calculation would be correct if you used microseconds instead of milliseconds.

However, using System to get the time seems simpler

polvalente: I would wrap the code to yield seconds directly from System.monotonic_time instead of converting from nanoseconds

sleipnir: @polvalente I wanted to avoid two calls to monotonic_time and so I would convert from nano to seconds, but your suggestion works too

polvalente: Right, that makes sense. I think calculating from a single System call might be better indeed, because the actual second can increment between calls

examples/helloworld/.gitignore
```diff
@@ -21,3 +21,5 @@ erl_crash.dump
 /tmp
 
 /log
+
+.iex.exs
```
polvalente, line 25:
```suggestion
```

I think this kind of ignored files should be added to your own global gitignore

examples/helloworld/mix.exs
```diff
@@ -2,24 +2,26 @@ defmodule Helloworld.Mixfile do
...
-     applications: [:logger, :grpc]]
+    [mod: {HelloworldApp, []}, applications: [:logger, :grpc]]
   end
 
   defp deps do
     [
       {:grpc, path: "../../"},
       {:protobuf, "~> 0.10"},
-      {:dialyxir, "~> 1.1", only: [:dev, :test], runtime: false},
+      {:google_protos, "~> 0.1"},
```
polvalente, line 23:
```suggestion
      {:google_protos, "~> 0.3.0"},
```

polvalente (approved):

examples/helloworld/lib/server.ex
angelo-moreira: For what I've read so far it seems that monotic time seems to be a measure to calculate the difference between 2 points in time without relying in the system clock, instead relying in the BEAM.

I might be wrong and I'm happy to be corrected, in the meantime I have updated the example to include nanos based in the Java example at https://github.com/protocolbuffers/protobuf/blob/79c3d07c63e711255e0d566c9b876706e978686a/src/google/protobuf/timestamp.proto#L90

Please let me know if you find this an acceptable solution.

polvalente (approved):

polvalente (approved):

Issue #265, opened 2022-08-21
Use newer type Supervisor.child_spec() instead of Supervisor.Spec.spec() for cowboy adapter


prihandi:
I've just tried to run GRPC server as a plugin in VerneMQ. Then I found an error pointing to 

https://github.com/elixir-grpc/grpc/blob/c7ee0c11ad9eb95a8925a342af8e2d5b1f082fee/lib/grpc/server/adapters/cowboy.ex#L45-L46

turns out current VerneMQ using newer `ranch` version (`2.1.0`) instead of old (`1.8.0`), and the pattern match failed because newer versin of `ranch` using new map version of `:supervisor.child_spec` instead of the old  5/6 elements tuple.

Then I seek Phoenix/Plug implementation, because phoenix runs well on my plugin. I found they're using below approach that returning map version instead of the tuple version 

https://github.com/elixir-plug/plug_cowboy/blob/d5b54031fb7fe98b8f2025e51587db988e4dabf0/lib/plug/cowboy.ex#L257-L270

```elixir
    case :ranch.child_spec(ref, ranch_module, transport_opts, cowboy_protocol, proto_opts) do
      {id, start, restart, shutdown, type, modules} ->
        %{
          id: id,
          start: start,
          restart: restart,
          shutdown: shutdown,
          type: type,
          modules: modules
        }

      child_spec when is_map(child_spec) ->
        child_spec
    end
```

I think it's better for this lib do similar approach, for the sake of compatibility

polvalente:
PRs are welcome!

PR #266, opened 2022-08-22
Replace tuple style supervisor child_spec with map on cowboy adapter

prihandi:
PR for #265  @polvalente 

- Changing child_spec format/type from Supervisor.Spec.spec() (a tuple) to Supervisor.child_spec() (a map)
- Updating test case


polvalente (approved):
Thanks!

prihandi:
I see CI pipeline failed on dialyzer check because current `:ranch.child_spec` will never return a map, 

Should I suppress dialyzer warning for this?

polvalente:
> I see CI pipeline failed on dialyzer check because current `:ranch.child_spec` will never return a map, 
> 
> Should I suppress dialyzer warning for this?

Does the code allow for returning a map, though?

prihandi:
Yes, for newer version of `:ranch`

polvalente:
Looks like this PR should also update Ranch

lib/grpc/server/adapters/cowboy.ex
```diff
@@ -42,13 +42,26 @@ defmodule GRPC.Server.Adapters.Cowboy do
         {:ranch_tcp, :cowboy_clear}
       end
 
-    {ref, mfa, type, timeout, kind, modules} =
-      :ranch.child_spec(ref, transport, trans_opts, protocol, proto_opts)
+    case :ranch.child_spec(ref, transport, trans_opts, protocol, proto_opts) do
```
polvalente, line 45: If we update ranch, we won't need this `case`

prihandi: If we update `:ranch`, then the tuple version would never match afterward, the dialyzer would see it as warning

polvalente: Yeah, because we won't need to support the tuple version :)

mix.exs
```diff
@@ -44,6 +44,7 @@ defmodule GRPC.Mixfile do
       # but we can't depend on an RC for releases
       {:gun, "~> 2.0.1", hex: :grpc_gun},
       {:cowlib, "~> 2.11"},
+      {:ranch, "~> 2.0", override: true},
```
polvalente, line 47: Why do we need the override?
We can't publish the package with the override here :(

prihandi: because cowlib still reference to `"1.8.0"`, this means this approach need to suppress warning from dialyzer on no_match for `:ranch.child_spec`

prihandi: I'll revert it back

polvalente: Ok, so we can't update ranch after all.
Best approach is to revert to the case implementation and ignore the dialyzer warning for that specific line.

polvalente: Hex blocks packages that have override: true for any dependencies, because override should only be used at the top level

lib/grpc/server/adapters/cowboy.ex
```diff
@@ -42,13 +44,24 @@ defmodule GRPC.Server.Adapters.Cowboy do
...
```
polvalente, line 47: Could you add a comment here and on the dialyzer warning explaining why these pieces of code are here?

lib/grpc/server/adapters/cowboy.ex
prihandi: we can't update ranch because cowlib still dependent on version 1.8.0 :(

lib/grpc/server/adapters/cowboy.ex
prihandi: It's just a no_match/never match warning since `:ranch` version 1.8.0 will always return a tuple lilke current implementation, while new version (2.x.x) will always return map.
So, even though we update ranch the warning still exist because it'll just return one type of child_spec, a tuper or a map
And turned out we can't update `:ranch` because `:cowlib` still dependent on version 1.8.0

I think I'll close my PR, since it'd not be generic use cases for having support for both of supervisor child_spec

polvalente (approved):

PR #267, opened 2022-08-24
Remove gun-specific code from stub to improve client adapter flexibility

beligante:
I'm working on this issue: https://github.com/elixir-grpc/grpc/issues/242

And while working on that, I've found that `stub.ex` has a lot of `gun` specific code and that should live inside the adapter. 

This doesn't contains drastic code changes, I'm simply moving a small piece of code around to keep on working on adapter code.

My Idea here is to create an layer of abstraction, where each adapter implements it's own version of receiving data. For `gun` we have the mechanics of receive the headers first and then the payload. 

When working with Mint that's not the case. See [`stream/2` docs](https://hexdocs.pm/mint/Mint.HTTP.html#stream/2) 
TL;DR: The message we receive in the process might contain headers and parts of the payload (in the same message). So having a function that receives the headers and the body doesn't make to much sense.


lib/grpc/client/adapter.ex
```diff
@@ -17,13 +17,10 @@ defmodule GRPC.Client.Adapter do
...
-
-  @callback recv_data_or_trailers(
-              stream :: map(),
-              trailers_or_metadata :: map(),
-              opts :: keyword()
-            ) ::
-              {:data, binary()} | {:trailers, binary()} | {:error, GRPC.RPCError.t()}
+  @callback receive_data(stream :: Stream.t(), opts :: keyword()) ::
+              {:ok, struct()}
+              | {:ok, struct(), map()}
```
polvalente, line 22: what struct can be returned here?

lib/grpc/client/adapter.ex
```diff
@@ -17,13 +17,10 @@ defmodule GRPC.Client.Adapter do
...
+              | {:ok, Enumerable.t()}
+              | {:ok, Enumerable.t(), map()}
```
polvalente, line 24: what is the enumerable here? `name :: type` notation could be useful

lib/grpc/client/adapters/gun.ex
```diff
@@ -135,7 +135,42 @@ defmodule GRPC.Client.Adapters.Gun do
...
+        %{server_stream: true, channel: %{adapter_payload: adapter_payload}, payload: payload} =
+          stream,
+        opts
+      ) do
+    with {:ok, headers, is_fin} <- recv_headers(adapter_payload, payload, opts) do
+      case is_fin do
+        :fin -> []
+        :nofin -> response_stream(stream, opts)
+      end
+      |> then(&{:ok, &1})
```
polvalente, line 148:
```suggestion
        :fin -> {:ok, []}
        :nofin -> {:ok, response_stream(stream, opts)}
      end
```

for readability

lib/grpc/client/adapters/gun.ex
```diff
@@ -135,7 +135,42 @@ defmodule GRPC.Client.Adapters.Gun do
...
+      |> maybe_return_headers(opts[:return_headers], %{headers: headers})
```
polvalente, line 149: I don't quite like `maybe_*` naming. I'd rather use plain `return_headers` or avoid the function altogether in favor of an in-place if

polvalente: It also doesn't just return headers because you use it to return a map which can also contain trailers

lib/grpc/client/adapters/gun.ex
```diff
@@ -287,4 +321,146 @@ defmodule GRPC.Client.Adapters.Gun do
...
+        end
+
+      error = {:error, _} ->
+        {error, %{buffer: <<>>, fin: true, fin_resp: nil}}
+    end
+  end
+
+  defp read_stream(%{buffer: buffer, need_more: false, response_mod: res_mod, codec: codec} = s) do
+    case GRPC.Message.get_message(buffer) do
+      # TODO
```
polvalente, line 409: what's "TODO" here?

lib/grpc/client/adapters/gun.ex
```diff
@@ -287,4 +321,146 @@ defmodule GRPC.Client.Adapters.Gun do
...
+          end
+
+        case GRPC.Message.from_data(%{compressor: compressor}, body) do
+          {:ok, msg} ->
+            {:ok, codec.decode(msg, res_mod)}
+
+          err ->
+            err
+        end
+
```
polvalente, line 451: the nesting here seems too much. Perhaps this branch can be a separate defp?

polvalente (commented):
This looks good! Some minor nitpicks and we're good to go

lib/grpc/client/adapter.ex
beligante: In this case the enumerable refers to an `Elixir.Stream` mostly used for streamed RPCs

lib/grpc/client/adapters/gun.ex
beligante: IDK to be honest lol
This part was a copy and paste from `stub.ex` code. It's comes from the original author. Seems unnecessary. I'll remove this one

lib/grpc/client/adapters/gun.ex
beligante: yeah! Agreed! I'll refactor this

lib/grpc/client/adapter.ex
beligante: Added a minor doc and a new type in `GRPC.Stub` to improve this.

lib/grpc/client/adapters/gun.ex
```diff
@@ -135,7 +135,37 @@ defmodule GRPC.Client.Adapters.Gun do
   end
 
   @impl true
-  def recv_headers(%{conn_pid: conn_pid}, %{stream_ref: stream_ref}, opts) do
+  def receive_data(
+        %{server_stream: true, channel: %{adapter_payload: adapter_payload}, payload: payload} =
+          stream,
+        opts
+      ) do
```
polvalente, line 142: Let's split this match, leaving only the flow-control at the function head
```suggestion
        %{server_stream: true} = stream,
        opts
      ) do
```

And then match to extract the data below. Same for the other clause

polvalente (approved):

PR #271, opened 2022-09-06
url encode/decode status-message header

### LINKED ISSUE #269: `grpc-message` is not percent-encoded as specified in the gRPC HTTP/2 spec
**Describe the bug**

In [gRPC over HTTP/2 spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#responses), it says that the `grpc-message` should be percent-encoded.

When sending `grpc-message` containing unencoded char `%`, this will cause some gRPC clients (e.g. in my case, grpc-js) to throw an exception.

Even though the spec also says (the client) MUST NOT error or throw away the message, it's still suggested to add the encoding on the server side (I'm not sure whether there is a corresponding decoding part on the elixir-grpc client side). Meanwhile, we also created an issue in grpc-js and it has been addressed.

**Expected behavior**

`grpc-message` should be correctly encoded as the spec.

drowzy:
Closes #269. As mentioned in the issue uses `URI.encode/1` to encode status message trailer `grpc-message` on the server and decodes the value on the client using `URI.decode`.


beligante (approved):
Nice

polvalente (approved):

PR #272, opened 2022-09-10
feat: Add Mint adapter

beligante:
Closes: https://github.com/elixir-grpc/grpc/issues/242
(sorry for the huge PR - tried to make everything well tested + make interop tests work with mint)

## Covered in this PR

Cover all possible use cases for a gRPC (client-server):
- Unary -> Unary
- Unary -> Stream
- Stream -> Unary
- Steam -> Stream

There is also several E2E test cases [here](https://github.com/beligante/grpc/blob/add-mint-adapter/interop/script/run.exs#L20), which includes:

- Canceling requests
- Compressing headers
- Large requests
- Timeout scenarios

## Suggestion on how to review

We have two main modules that handles the interactions with the server. 

#### Connection Process
This process is spawned every time we attempt to connect to a gRPC server. It does strictly three things:
- Open and store the Mint connection struct. 
- Deals with `tcp_messages` . And that includes
  - Handling window_size frames
  - pass the headers/trailers/data to the response process
- Handling the requests 

Because of [HTTP2 flow control](https://medium.com/coderscorner/http-2-flow-control-77e54f7fd518) mechanics I've designed the request handling of all requests to behave as a streamed request, this way it's easier to check for the window size and chunk the request if necessary. Because of that, you might see that I'm treating even unary requests, as streams - but that behavior is transparent to the user of the lib. 

I'm also handling connection drops with the following strategy:
- I close/end all pending requests as [suggested here](https://github.com/elixir-mint/mint/issues/211#issuecomment-529832679)
- I'm also sending an message to the parent process to inform that the connection droped. with the following format:
```elixir
{:elixir_grpc, :connection_down, self_pid}
```

#### Response Process
This process is spawned every time a request is made. It's responsible to receive cast calls from the `ConnectionProcess` and produce the messages to an Elixir.Stream. It dies when the request is finished - so it should be a short lived process.


lib/grpc/client/adapters/gun.ex
```diff
@@ -135,7 +137,37 @@ defmodule GRPC.Client.Adapters.Gun do
...
 
   @impl true
-  def recv_headers(%{conn_pid: conn_pid}, %{stream_ref: stream_ref}, opts) do
+  def receive_data(
+        %{server_stream: true, channel: %{adapter_payload: adapter_payload}, payload: payload} =
+          stream,
+        opts
+      ) do
+    with {:ok, headers, is_fin} <- recv_headers(adapter_payload, payload, opts),
+         response <- response_stream(is_fin, stream, opts) do
```
polvalente, line 146: since you don't match on anything specific for `response`, I think it's better to move it inside the `do` block

lib/grpc/client/adapters/mint.ex
```diff
@@ -0,0 +1,182 @@
...
+  def connect(%{host: host, port: port} = channel, opts \\ []) do
+    opts = Keyword.merge(@default_connect_opts, connect_opts(channel, opts))
+
+    channel
+    |> mint_scheme()
+    |> ConnectionProcess.start_link(host, port, opts)
+    |> case do
+      {:ok, pid} -> {:ok, %{channel | adapter_payload: %{conn_pid: pid}}}
+      # TODO add proper error handling
+      _error -> {:ok, %{channel | adapter_payload: %{conn_pid: nil}}}
```
polvalente, line 24: I would raise here instead, for now

lib/grpc/client/adapters/mint.ex
```diff
@@ -0,0 +1,182 @@
...
+    end
+  end
+
+  @impl true
+  def disconnect(%{adapter_payload: %{conn_pid: pid}} = channel)
+      when is_pid(pid) do
+    ConnectionProcess.disconnect(pid)
```
polvalente, line 31: I think at least a hard-match would be better here

lib/grpc/client/adapters/mint.ex
```diff
@@ -0,0 +1,182 @@
...
+    {:ok, %{channel | adapter_payload: %{conn_pid: nil}}}
```
polvalente, line 32: This throws away the previous adapter payload instead of just setting the conn_pid to nil. Is this intentional? Why not just set the adapter_payload to nil directly?

lib/grpc/client/adapters/mint.ex
```diff
@@ -0,0 +1,182 @@
...
+    end
+  end
+
+  # for streamed requests
+  def receive_data(
+        %{payload: %{response: {:ok, %{request_ref: _ref}}, stream_response_pid: pid}},
+        opts
+      ) do
+    with stream <- StreamResponseProcess.build_stream(pid),
+         responses <- Enum.into(stream, []),
```
polvalente, line 83:
```suggestion
         responses <- Enum.to_list(stream),
```

assuming this is an Elixir.Stream

lib/grpc/client/adapters/mint.ex
```diff
@@ -0,0 +1,182 @@
...
+      case opts[:return_headers] do
+        true -> {:ok, stream, headers}
+        _any -> {:ok, stream}
+      end
+    end
+  end
+
+  # for streamed requests
+  def receive_data(
+        %{payload: %{response: {:ok, %{request_ref: _ref}}, stream_response_pid: pid}},
```
polvalente, line 79: why match on the ref with an empty variable?

lib/grpc/client/adapters/mint.ex
```diff
@@ -0,0 +1,182 @@
...
+        opts
+      ) do
+    with stream <- StreamResponseProcess.build_stream(pid),
+         responses <- Enum.into(stream, []),
+         :ok <- check_for_error(responses) do
+      {:ok, data} = Enum.find(responses, fn {status, _data} -> status == :ok end)
```
polvalente, line 85:
```suggestion
      data = Keyword.fetch!(responses, :ok)
```

lib/grpc/client/adapters/mint.ex
```diff
@@ -0,0 +1,182 @@
...
+      end
+    end
+  end
+
+  def receive_data(%{payload: %{response: response, stream_response_pid: pid}}, opts) do
+    with {:ok, %{headers: headers}} <- response,
+         stream <- StreamResponseProcess.build_stream(pid),
+         responses <- Enum.into(stream, []),
+         :ok <- check_for_error(responses) do
+      {:ok, data} = Enum.find(responses, fn {status, _data} -> status == :ok end)
```
polvalente, line 99:
```suggestion
      data = Keyword.fetch!(responses, :ok)
```

lib/grpc/client/adapters/mint.ex
```diff
@@ -0,0 +1,182 @@
...
+        true -> {:ok, data, get_headers(responses) |> append_trailers(responses)}
+        _any -> {:ok, data}
+      end
+    end
+  end
+
+  def receive_data(%{payload: %{response: response, stream_response_pid: pid}}, opts) do
+    with {:ok, %{headers: headers}} <- response,
+         stream <- StreamResponseProcess.build_stream(pid),
+         responses <- Enum.into(stream, []),
```
polvalente, line 97:
```suggestion
         responses <- Enum.to_list(stream),
```

lib/grpc/client/adapters/mint.ex
```diff
@@ -0,0 +1,182 @@
...
+    transport_opts = Keyword.get(opts, :transport_opts, [])
+    [transport_opts: Keyword.merge(@default_transport_opts, transport_opts)]
+  end
+
+  defp mint_scheme(%Channel{scheme: "https"} = _channel), do: :https
+  defp mint_scheme(_channel), do: :http
+
+  def check_for_error(responses) do
+    error = Enum.find(responses, fn {status, _data} -> status == :error end)
+    if error == nil, do: :ok, else: error
```
polvalente, line 162:
```suggestion
    if error, do: error, else: :ok
```

polvalente, line 162: could also be the following, albeit it's less legible:
```suggestion
    Keyword.get(responses, :error) || :ok
```

lib/grpc/client/adapters/mint.ex
```diff
@@ -0,0 +1,182 @@
...
+    end
+  end
+
+  defp get_headers(responses) do
+    responses
+    |> Enum.find(fn {status, _data} -> status == :headers end)
+    |> case do
+      nil -> nil
+      {:headers, headers} -> headers
+    end
```
polvalente, line 180:
```suggestion
    Enum.find_value(responses, fn 
      {:headers, headers} -> headers
      _ -> nil
    end)
```

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -0,0 +1,177 @@
...
+  def start_link(scheme, host, port, opts \\ []) do
+    GenServer.start_link(__MODULE__, {scheme, host, port, opts})
+  end
+
+  def request(pid, method, path, headers, body, opts \\ []) do
+    GenServer.call(pid, {:request, method, path, headers, body, opts})
+  end
+
+  def disconnect(pid) do
+    GenServer.call(pid, {:disconnect, :brutal})
```
polvalente, line 18: why `:brutal`?

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -0,0 +1,177 @@
...
+  def stream_request_body(pid, request_ref, body) do
+    GenServer.call(pid, {:stream_body, request_ref, body})
+  end
+
+  ## Callbacks
+
+  @impl true
+  def init({scheme, host, port, opts}) do
+    # The current behavior in gun is return error if the connection wasn't successful
+    # Should we do the same here?
```
polvalente, line 30: We can design a separate UX if we want. What would be the alternative?

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -0,0 +1,177 @@
...
+    case Mint.HTTP.connect(scheme, host, port, opts) do
+      {:ok, conn} ->
+        {:ok, State.new(conn)}
+
+      {:error, reason} ->
+        # TODO check what's better: add to state map if connection is alive?
+        # TODO Or simply stop the process and handle the error on caller?
```
polvalente, line 37: Since we have a process per connection, I think killing it might represent better what's happening

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -0,0 +1,177 @@
...
+  end
+
+  defp process_response({:status, request_ref, status}, state) do
+    State.update_response_status(state, request_ref, status)
+  end
+
+  defp process_response({:headers, request_ref, headers}, state) do
+    empty_headers? = State.empty_headers?(state, request_ref)
+
+    case empty_headers? do
```
polvalente, line 128: this is an `if-else`

lib/grpc/client/adapters/mint/stream_response_process.ex
```diff
@@ -0,0 +1,156 @@
...
+  def done(pid) do
+    GenServer.cast(pid, {:consume_response, :done})
+  end
+
+  @doc """
+  Cast a message to process to consume an incoming data or trailers
+  """
+  @spec consume(pid(), :data | :trailers | :headers, binary() | Mint.Types.headers()) :: :ok
+  def consume(pid, :data, data) do
+    GenServer.cast(pid, {:consume_response, {:data, data}})
```
polvalente, line 35: I don't like these `cast` calls because they'll ignore any errors or missed messages. Shouldn't they all be either a `send` or a GenServer.call?

polvalente: all handle_cast below would change to handle_call or handle_info depending on what we choose here

lib/grpc/client/adapters/mint/stream_response_process.ex
```diff
@@ -0,0 +1,156 @@
...
+    status = String.to_integer(decoded_trailers["grpc-status"])
+
+    case status == GRPC.Status.ok() do
+      true ->
+        {:trailers, decoded_trailers}
+
+      false ->
+        rpc_error = %GRPC.RPCError{status: status, message: decoded_trailers["grpc-message"]}
+        {:error, rpc_error}
+    end
```
polvalente, line 150: if-else

polvalente (commented):
The overall work looks great!

lib/grpc/client/adapters/mint.ex
beligante: Yeah! that makes sense

lib/grpc/client/adapters/mint.ex
beligante: There are two usages for `receive_data/2`: 
- One is for server stream
- Other is for unary response

In this first one I'm patter matching for the server stream case. But it's a bit confusing in fact. I'll refactor this to return an atom, just to be clear what each match is solving. WDYT?

polvalente: I see! Perhaps you could use a `defguardp` to validate the `{:ok, value}`, like `defguardp is_client_stream_response(value) when is_ref(value.request_ref)` and another for `is_unary_request_response`

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
beligante: Ideally in the future we could add logic here to wait for pending request to finish before killing the connection. `brutal` here refers to the fact that I'm going to kill the connection without wait for pending requests. thoughts?

lib/grpc/client/adapters/mint/stream_response_process.ex
beligante: So, in this case I'm treating as a cast call, cause I'm this process to be just about perform request and receive tcp messages. process the response (event the error ones) should be done on the response process. 

Cause this way I don't block the connection process from receiving other messages and also in case of an error, I would like to propagate them to the response process and put them in the Elixir.Strem. So the caller is responsible to handle errors. 

IMO I don't see lot's of benefits on using a call here, since the error treatment is done on the other hand of the processing.

lib/grpc/client/adapters/mint.ex
beligante: WDYT of a guard for each case?

- `is_bidirectional_stream`
- `is_unary_request_stream_response`
- `is_stream_request_unary_response`
- `is_unary_request_response`

I think it would make it easier to understand responsibility of each case  + I could add a doc for each guard

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
beligante: Currently I'm spawning the connection async for mint. So, if the connection doesn't succeeds I need to perform a request to know that it wasn't successful. 

To match the behavior we have for GUN I would need to make connection spawn be sync OR create a `call` for the gen_server to check if the connection was successful

lib/grpc/client/adapters/mint.ex
beligante: Actually I just refactor this, to use a cond. Way more cleaner than before.

mix.exs
```diff
@@ -43,10 +43,13 @@ defmodule GRPC.Mixfile do
       # This is the same as :gun 2.0.0-rc.2,
       # but we can't depend on an RC for releases
       {:gun, "~> 2.0.1", hex: :grpc_gun},
+      {:mint, "~> 1.4.2"},
       {:cowlib, "~> 2.11"},
       {:protobuf, "~> 0.10", only: [:dev, :test]},
       {:ex_doc, "~> 0.28.0", only: :dev},
-      {:dialyxir, "~> 1.1.0", only: [:dev, :test], runtime: false}
+      {:dialyxir, "~> 1.1.0", only: [:dev, :test], runtime: false},
+      {:ex_machina, "~> 2.7.0", only: :test},
```
polvalente, line 51: I'd rather not add ExMachina as a dependency here

beligante: Out of curiosity, this is just to avoid having the dependency on the repo. Right?

interop/script/run.exs
```diff
@@ -13,49 +13,39 @@ Logger.configure(level: level)
...
+                            Client.empty_stream!(ch)
+                            Client.custom_metadata!(ch)
+                            Client.status_code_and_message!(ch)
+                            Client.unimplemented_service!(ch)
+                            Client.cancel_after_begin!(ch)
+                            Client.cancel_after_first_response!(ch)
+                            Client.timeout_on_sleeping_server!(ch)
+                          end
+                          :ok
+  end, max_concurrency: concurrency, ordered: false, timeout: :infinity)
```
wingyplus, line 46: How about extracting the anonymous function to a variable?

lib/grpc/client/adapters/mint/stream_response_process.ex
```diff
@@ -0,0 +1,178 @@
...
+        nil -> nil
+        response -> {response, pid}
+      end
+    end)
+  end
+
+  @doc """
+  Cast a message to process to inform that the stream has finished
+    once all messages are produced. This process will automatically
+    be killed.
```
wingyplus, line 40:
```suggestion
  once all messages are produced. This process will automatically
  be killed.
```

polvalente:
Is this ready for review?

beligante:
> Is this ready for review?

M'bad! Forgot to ping you! Yes! It's ready

One thing that I might need your help here is: How to make mint optional here? 

I tried to add the check using the `Code.ensure_compiled?`, but it stops to run the tests

interop/script/run.exs
beligante: https://github.com/elixir-grpc/grpc/pull/272/commits/d3e026f6bbdb462d044247f94a5a1104866e1b23

polvalente:
> > Is this ready for review?
> 
> M'bad! Forgot to ping you! Yes! It's ready
> 
> One thing that I might need your help here is: How to make mint optional here?
> 
> I tried to add the check using the `Code.ensure_compiled?`, but it stops to run the tests

You can try using `Code.ensure_loaded`, which is a blocking call. Also, be aware that tests will always see an optional dependency as available, so it might be hard to check that it is indeed optional.

interop/lib/interop/client.ex
```diff
@@ -120,12 +117,9 @@ defmodule Interop.Client do
         size: 92653}
     ])
     {:ok, res_enum} = ch |> Grpc.Testing.TestService.Stub.streaming_output_call(req)
-    result = Enum.map([31415, 92653], &String.duplicate(<<0>>, &1))
+    result = Enum.map([31415, 92653], &String.duplicate(<<0>>, &1)) |> Enum.sort()
```
polvalente, line 120: Is this sort necessary?

interop/lib/interop/client.ex
```diff
@@ -191,19 +185,31 @@ defmodule Interop.Client do
         payload: payload(271_828)
       )
 
-    {:ok, res_enum, %{headers: new_headers}} =
+    {headers, data, trailers} =
       ch
-      |> Grpc.Testing.TestService.Stub.full_duplex_call(metadata: metadata)
+      |> Grpc.Testing.TestService.Stub.full_duplex_call(metadata: metadata, return_headers: true)
```
polvalente, line 190: I'm not sure if we should add `return_headers` here. Perhaps we should keep a version that has `return_headers` and one that does not.

interop/lib/interop/client.ex
```diff
@@ -191,19 +185,31 @@ defmodule Interop.Client do
...
     {:trailers, new_trailers} = Stream.take(res_enum, 1) |> Enum.to_list() |> List.first()
-    validate_headers!(new_headers, new_trailers)
+    {new_headers, data, new_trailers}
+  end
+
+  defp process_full_duplex_response({:ok, res_enum}) do
+    {:headers, headers} = Stream.take(res_enum, 1) |> Enum.to_list() |> List.first()
+    {:ok, data} = Stream.take(res_enum, 1) |> Enum.to_list() |> List.first()
+    {:trailers, trailers} = Stream.take(res_enum, 1) |> Enum.to_list() |> List.first()
+    {headers, data, trailers}
```
polvalente, line 212: I don't get how this works. It seems to me that you're always obtaining the same value from the stream and matching into 3 different formats.

```elixir
iex(1)> s = Stream.cycle([1, 2, 3, 4])
#Function<63.6935098/2 in Stream.unfold/2>
iex(2)> s |> Stream.take(1) |> Enum.to_list()
[1]
iex(3)> s |> Stream.take(1) |> Enum.to_list()
[1]
iex(4)> s |> Stream.take(3) |> Enum.to_list()
[1, 2, 3]
```

I'd expect something closer to the last line instead.

interop/lib/interop/client.ex
```diff
@@ -226,6 +232,10 @@ defmodule Interop.Client do
       |> Grpc.Testing.TestService.Stub.full_duplex_call()
       |> GRPC.Stub.send_request(req, end_stream: true)
       |> GRPC.Stub.recv()
+      |> case do
+        {:ok, stream} -> Stream.take(stream, 1) |> Enum.to_list() |> List.first()
```
polvalente, line 236:
```suggestion
        {:ok, stream} -> Enum.at(stream, 0)
```

interop/script/run.exs
```diff
@@ -13,49 +13,46 @@ Logger.configure(level: level)
 Logger.info("Rounds: #{rounds}; concurrency: #{concurrency}; port: #{port}")
 
 alias Interop.Client
+alias GRPC.Client.Adapters.{Mint, Gun}
```
polvalente, line 16:
```suggestion
alias GRPC.Client.Adapters.Mint
alias GRPC.Client.Adapters.Gun
```

I prefer this style because it makes for easier refactoring and find-all

lib/grpc/client/adapters/mint.ex
```diff
@@ -0,0 +1,227 @@
+defmodule GRPC.Client.Adapters.Mint do
+  @moduledoc """
+  A client adapter using mint
+  """
+
+  alias GRPC.{Channel, Credential}
+  alias GRPC.Client.Adapters.Mint.{ConnectionProcess, StreamResponseProcess}
```
polvalente, line 7: Please separate these into different `alias` calls as well

lib/grpc/client/adapters/mint.ex
```diff
@@ -0,0 +1,227 @@
...
+    %{
+      channel: %{adapter_payload: %{conn_pid: conn_pid}},
+      payload: %{response: {:ok, %{request_ref: request_ref}}}
+    } = stream
+
+    ConnectionProcess.cancel(conn_pid, request_ref)
+  end
+
+  defp connect_opts(%Channel{scheme: "https"} = channel, opts) do
+    %Credential{ssl: ssl} = Map.get(channel, :cred, %Credential{})
```
polvalente, line 119:
```suggestion
    %Credential{ssl: ssl} = Map.get(channel, :cred) || %Credential{}
```

lib/grpc/client/adapters/mint.ex
```diff
@@ -0,0 +1,227 @@
...
+    {:ok, stream}
+  end
+
+  defp do_receive_data(
+         %{payload: %{stream_response_pid: pid}},
+         request_type,
+         opts
+       )
+       when request_type in [:stream_request_unary_response, :unary_request_response] do
+    with stream <- StreamResponseProcess.build_stream(pid),
```
polvalente, line 149: This `with` seems unnecessary because both the first and second clauses always match. This could be refactored into a `case`

lib/grpc/client/adapters/mint.ex
```diff
@@ -0,0 +1,227 @@
...
+        {:ok, data}
+      end
+    end
+  end
+
+  def handle_errors_receive_data(%GRPC.Client.Stream{payload: %{response: response}}, _opts) do
+    {:error, "an error occurred while when receiving data: error=#{inspect(response)}"}
+  end
+
+  defp success_bidi_stream?(%GRPC.Client.Stream{
```
polvalente, line 166: I think it would be more clear if you only passed the payload to these `success_*` functions, with an outer case selecting based on the grpc_type value

polvalente: Also, don't all of them work the same way, returning true upon {ok, _} and false otherwise?

lib/grpc/client/adapters/mint.ex
```diff
@@ -0,0 +1,227 @@
...
+        do_receive_data(stream, :bidirectional_stream, opts)
+
+      success_server_stream?(stream) ->
+        do_receive_data(stream, :unary_request_stream_response, opts)
+
+      success_client_stream?(stream) ->
+        do_receive_data(stream, :stream_request_unary_response, opts)
+
+      success_unary_request?(stream) ->
+        do_receive_data(stream, :unary_request_response, opts)
```
polvalente, line 67: Ok, so I now get what the success_ functions mean.
I'd rather have a `success_response?` function and evaluate explicitly the grpc_type field here on the cond

lib/grpc/client/adapters/mint.ex
```diff
@@ -0,0 +1,227 @@
...
+      end
+    end
+  end
+
+  def handle_errors_receive_data(%GRPC.Client.Stream{payload: %{response: response}}, _opts) do
+    {:error, "an error occurred while when receiving data: error=#{inspect(response)}"}
+  end
+
+  defp success_bidi_stream?(%GRPC.Client.Stream{
+         grpc_type: :bidi_stream,
```
polvalente, line 167: can we refactor `bidi_stream` into `bidirectional_stream` for clarity?

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -0,0 +1,390 @@
+defmodule GRPC.Client.Adapters.Mint.ConnectionProcess do
+  @moduledoc """
+  This module is responsible for manage a connection with a grpc server.
```
polvalente, line 3:
```suggestion
  This module is responsible for managing a connection with a gRPC server.
```

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -0,0 +1,390 @@
...
+  It's also responsible for manage requests, which also includes check for the
```
polvalente, line 4:
```suggestion
  It's also responsible for managing requests, which also includes checks for the
```

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -0,0 +1,390 @@
...
+  connection/request window size, split a given payload into appropriate sized chunks
```
polvalente, line 5:
```suggestion
  connection/request window size, splitting a given payload into appropriate sized chunks
```

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -0,0 +1,390 @@
...
+  and stream those to the server using an internal queue.
```
polvalente, line 6:
```suggestion
  and streaming those to the server using an internal queue.
```

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
polvalente: Perhaps just use `:disconnect` for now. Any refactors in the future can think of separate terminology depending on the use case.

polvalente: brutal/friendly doesn't seem like the right terminology for this

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -0,0 +1,390 @@
...
+        {:noreply, State.update_conn(state, conn)}
+    end
+  end
+
+  defp stream_body_and_reply({request_ref, body, from}, state) do
+    send_eof? = from == nil
+
+    case stream_body(state.conn, request_ref, body, send_eof?) do
+      {:ok, conn} ->
+        if from != nil do
```
polvalente, line 287:
```suggestion
        if not is_nil(from) do
```

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -0,0 +1,390 @@
...
+          GenServer.reply(from, :ok)
+        end
+
+        check_request_stream_queue(State.update_conn(state, conn))
+
+      {:error, conn, error} ->
+        if from != nil do
```
polvalente, line 294:
```suggestion
        if not is_nil(from) do
```

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -0,0 +1,390 @@
...
+    with {:ok, conn} <- Mint.HTTP.stream_request_body(conn, request_ref, body),
+         {:ok, conn} <- Mint.HTTP.stream_request_body(conn, request_ref, :eof) do
+      {:ok, conn}
+    end
+  end
+
+  defp stream_body(conn, request_ref, body, false = _stream_eof?) do
+    with {:ok, conn} <- Mint.HTTP.stream_request_body(conn, request_ref, body) do
+      {:ok, conn}
+    end
```
polvalente, line 316:
```suggestion
    Mint.HTTP.stream_request_body(conn, request_ref, body)
```

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -0,0 +1,390 @@
...
+  def check_request_stream_queue(state) do
+    if :queue.is_empty(state.request_stream_queue) do
+      {:noreply, state}
+    else
+      {:noreply, state, {:continue, :process_request_stream_queue}}
+    end
+  end
+
+  defp chunk_body(body, bytes_length) do
+    <<head::binary-size(bytes_length), tail::binary>> = body
```
polvalente, line 328: This breaks if body has less bytes than bytes_length. Is this intended? You can use byte_size as a guard to protect against this if it's not intended

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -0,0 +1,390 @@
...
+
+          {ref, _body, from} ->
+            acc_state
+            |> State.stream_response_pid(ref)
+            |> send_connection_close_and_end_stream_response()
+
+            GenServer.reply(from, {:error, @connection_closed_error})
+        end
+
+        {ref, _, _} = request
```
polvalente, line 358: I'd rather this match be on the function head instead

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -0,0 +1,390 @@
...
+        {_ref, new_state} = State.pop_ref(acc_state, ref)
```
polvalente, line 359:
```suggestion
        {^ref, new_state} = State.pop_ref(acc_state, ref)
```

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -0,0 +1,390 @@
...
+
+        new_state
+      end)
+
+    # Inform the parent that the connection is down
+    send(new_state.parent, {:elixir_grpc, :connection_down, self()})
+
+    new_state.requests
+    |> Map.keys()
```
polvalente, line 368: You should use Enum.each on the map directly to avoid creating the intermediate list

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -0,0 +1,390 @@
...
+      |> State.stream_response_pid(ref)
+      |> send_connection_close_and_end_stream_response()
+    end)
+
+    {:noreply, State.update_request_stream_queue(%{new_state | requests: %{}}, :queue.new())}
+  end
+
+  defp send_connection_close_and_end_stream_response(pid) do
+    StreamResponseProcess.consume(pid, :error, @connection_closed_error)
+    StreamResponseProcess.done(pid)
```
polvalente, line 380: I think a match on the consume result would be nice, to ensure that it  succeeded

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -0,0 +1,390 @@
...
+  end
+
+  def check_connection_status(state) do
+    if(Mint.HTTP.open?(state.conn)) do
```
polvalente, line 384:
```suggestion
    if Mint.HTTP.open?(state.conn) do
```

lib/grpc/client/adapters/mint/stream_response_process.ex
```diff
@@ -0,0 +1,178 @@
+defmodule GRPC.Client.Adapters.Mint.StreamResponseProcess do
+  @moduledoc """
```
polvalente, line 2: Is this a public module that people should be using?

lib/grpc/client/adapters/mint/stream_response_process.ex
```diff
@@ -0,0 +1,178 @@
...
+    GenServer.start_link(__MODULE__, {stream, send_headers_or_trailers?})
+  end
+
+  @doc """
+  Given a pid from this process, build an Elixir.Stream that will consume the accumulated
+    data inside this process
+  """
+  @spec build_stream(pid()) :: Enumerable.t()
+  def build_stream(pid) do
+    Elixir.Stream.unfold(pid, fn pid ->
```
polvalente, line 29:
```suggestion
    Stream.unfold(pid, fn pid ->
```

There's no alias shadowing the module name

lib/grpc/client/adapters/mint/stream_response_process.ex
```diff
@@ -0,0 +1,178 @@
...
+  def done(pid) do
+    GenServer.cast(pid, {:consume_response, :done})
+  end
+
+  @doc """
+  Consume an incoming data or trailers/headers
+  """
+  @spec consume(pid(), type :: accepted_types, data :: data_types) :: :ok
+  def consume(pid, type, data) when type in @accepted_types do
+    GenServer.cast(pid, {:consume_response, {type, data}})
```
polvalente, line 52: Does this _have_ to be a cast? Casts are non-blocking and behave in a fire-and-forget way.

lib/grpc/message.ex
```diff
@@ -172,4 +172,28 @@ defmodule GRPC.Message do
...
+        data
+    end
+  end
+
+  def get_message(data, compressor) do
+    case data do
+      <<1, length::unsigned-integer-size(32), message::bytes-size(length), rest::binary>> ->
+        {{1, compressor.decompress(message)}, rest}
+
+      <<0, length::unsigned-integer-size(32), message::bytes-size(length), rest::binary>> ->
```
polvalente, line 192:
```suggestion
      <<1::8, length::unsigned-integer-32, message::bytes-size(length), rest::binary>> ->
        {{1, compressor.decompress(message)}, rest}

      <<0::8, length::unsigned-integer-32, message::bytes-size(length), rest::binary>> ->
```

polvalente: adding ::8 to be more explicit about it

lib/grpc/server/adapters/cowboy.ex
```diff
@@ -203,7 +203,8 @@ defmodule GRPC.Server.Adapters.Cowboy do
           # https://github.com/ninenines/cowboy/issues/1398
           # If there are 1000 streams in one connection, then 1000/s frames per stream.
           max_received_frame_rate: {10_000_000, 10_000},
-          max_reset_stream_rate: {10_000, 10_000}
+          max_reset_stream_rate: {10_000, 10_000},
+          stream_window_update_threshold: 65_500
```
polvalente, line 207: Why do we have to change the cowboy adapter here?

test/support/factory.ex
```diff
@@ -1,12 +1,11 @@
...
-  alias GRPC.Channel
-  alias GRPC.Credential
+  use GRPC.Factories.Channel
+  use GRPC.Factories.Client.Stream
 
-  @cert_path Path.expand("./tls/server1.pem", :code.priv_dir(:grpc))
-  @key_path Path.expand("./tls/server1.key", :code.priv_dir(:grpc))
-  @ca_path Path.expand("./tls/ca.pem", :code.priv_dir(:grpc))
+  # Protobuf factories
+  use GRPC.Factories.Proto.HelloWorld
```
polvalente, line 8: I'd keep things simple and keep all factories here for now. I know it seems cleaner to split up modules, but the fact that we have to encapsulate everything in macros smells bad for me.

There aren't too many factories yet to warrant this split IMO

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
wingyplus: Fixed in https://github.com/beligante/grpc/pull/1

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
wingyplus: Fixed in https://github.com/beligante/grpc/pull/1

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
wingyplus: Fixed in https://github.com/beligante/grpc/pull/1

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
wingyplus: Fixed in https://github.com/beligante/grpc/pull/1

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
wingyplus: Fixed in https://github.com/beligante/grpc/pull/1

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
wingyplus: Fixed in https://github.com/beligante/grpc/pull/1

wingyplus:
@beligante  Hi, I've some fix from code review  suggestions here https://github.com/beligante/grpc/pull/1. Please kindly review. 🙇

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
wingyplus: This function calls `GenServer.cast`, always returns `:ok`.

polvalente: Yeah, but I think that GenServer.cast should be avoided either way, if we can

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
wingyplus: We can't use `^ref` here the `ref` on the right-hand side is Reference but the left-hand side is a map.

polvalente: Perhaps use another name for the underscored variable then?

beligante:
Hey Guys! Tks for the review. I'm on vacations for one week. I promise to carefully check your comments once I'm back.

Hope you guys understand!

interop/lib/interop/client.ex
```diff
@@ -103,12 +103,9 @@ defmodule Interop.Client do
     params = Enum.map([31415, 9, 2653, 58979], &res_param(&1))
     req = Grpc.Testing.StreamingOutputCallRequest.new(response_parameters: params)
     {:ok, res_enum} = ch |> Grpc.Testing.TestService.Stub.streaming_output_call(req)
-    result = Enum.map([31415, 9, 2653, 58979], &String.duplicate(<<0>>, &1))
+    result = Enum.map([31415, 9, 2653, 58979], &String.duplicate(<<0>>, &1)) |> Enum.sort()
```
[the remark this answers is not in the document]
beligante: ohh yeah! I copy/paste the above list. I'll sort that

interop/lib/interop/client.ex
beligante: same as above

interop/lib/interop/client.ex
beligante: About this one. When using gun adapter, the adapter always returns the headers. :/

I could do as you asked, tho, I might have to write a match case for gun and another for mint. WDYT?

polvalente: Why do we have the `return_headers: option` then?
Is it new to the Mint adapter?

If so, I think it's important to add it on a separate, specific test.

interop/lib/interop/client.ex
beligante: So, I kept the behavior similarly to the gun adapter. A few things to make this easier to understand. 

1. We build the Stream struct using [Stream.unfold/2](https://hexdocs.pm/elixir/1.12/Stream.html#unfold/2)
2. the `unfold` function is built in such a way - for both adapters - that the `acc` is map used to find a connection_stream process
3. and the `next_fun` arg is a function that reads directly from the connection_stream that is producing data. 
4. Every time we execute the `next_fun` we read a chunk of data and remove that from the buffer. That action by itself will generate a side effect that will update the state of the connection_stream by removing the chunk of data we just read. 
5. An easier way to visualize that is. Take the code and the execution bellow as an example

### Code 

```elixir
# I used http terms here to make it easier to see the big picture on how this haves 

defmodule MyRequestProcess do
  use GenServer

  def init(_) do
    {:ok, [1, 2, 3]}
  end

  def handle_call(:get, _from, []) do
    {:reply, nil, []}
  end

  def handle_call(:get, _from, [head | tail]) do
    {:reply, head, tail}
  end
end

{:ok, pid} = GenServer.start_link(MyRequestProcess, [:ok])

ex_stream = Stream.unfold(pid, fn request_pid ->
  case GenServer.call(request_pid, :get) do
    nil -> nil
    value -> {value, request_pid}
  end
end)
```

### Execution

```bash
iex(4)> ex_stream |> Stream.take(1) |> Enum.to_list()
[1]
iex(5)> ex_stream |> Enum.to_list()
[2, 3]
iex(6)> ex_stream |> Enum.to_list()
[]
```
----
Final considerations

To make this behave as you pictured on the above comment, for both adapters I would have to somehow store chunks of data that I read from the connection and keep that in memory. 

Also, that would drastically change the current behavior of the lib. If you want to follow that path, I would recommend think in an new option to make the application behave like that and avoid add a breaking change

interop/script/run.exs
beligante: That makes sense. Tks

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
beligante: Sure! I'll do that update

interop/lib/interop/client.ex
beligante: That option is not new, its present on the original implementation. 

On gun adapter it's used to know if the trailer headers should be returned. 
On mint if the flag is off I don't return both (the response headers and the trailer headers)

beligante: I'll work on have a test of this one tho

interop/lib/interop/client.ex
polvalente: Right! I didn't realize that the stream was a gen server being wrapped. This clarifies things.

Perhaps a comment in the code highlighting this would be good.

polvalente: Mainly, the genserver makes the variable stateful instead of a common stream that we normally encounter

beligante: Sure! Maybe also add that in the docs about using the streaming?

polvalente: Sounds good!

lib/grpc/client/adapters/mint.ex
beligante: Ahh, right. This can definitely be simplified.

lib/grpc/client/adapters/mint.ex
beligante: Sure! This change might add some more changes on the Gun adapter. Since both use the same struct

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
beligante: the left one is an ref for `gen_server` and the right one is a `request_ref` used by mint internally. I'll update the terms

lib/grpc/client/adapters/mint/stream_response_process.ex
beligante: Nope! This is an internal module. You think I should declare this inside another module?

lib/grpc/client/adapters/mint/stream_response_process.ex
beligante: Yup! in this case this one makes sense to be a cast. Cause this process will be responsible (and should be the one responsible) for handling errors related to a single request. 

 I pictured this to to keep the connection process responsible for only deal with connection stuff (aka: manage the connection state and properly handle the the low level part of the requests and responses - chunk requests and distribute responses). 

If I use `call` here there's not much I can do to recover from errors from the connection process. Since at the end I designed the adapter to get responses from this module.

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
beligante: yeah...add a guard would be more safe. I'll make that change

lib/grpc/server/adapters/cowboy.ex
beligante: We don't need. I'll remove this. I was testing something and probably commit the change. M'bad

test/support/factory.ex
beligante: Yeah! That makes sense. I started using this and I thought I would be adding more stuff, but it's not the case. I'll make that change

lib/grpc/client/adapters/mint/stream_response_process.ex
polvalente: Just set moduledoc false and turn the docstring into comments!

lib/grpc/client/adapters/mint/stream_response_process.ex
polvalente: The problem with cast is that it returns ok even for dead processes. I don't think we should be using it. I'd rather have this be a call even if just for us to send a telemetry event on the failure case

lib/grpc/client/adapters/mint/stream_response_process.ex
beligante: You made a similar comment bellow. I'll reply there and mark this one as solved.

lib/grpc/client/adapters/mint/stream_response_process.ex
beligante: ahhh, right. I didn't thought about that. Makes sense. I'll do the change

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
beligante: https://github.com/elixir-grpc/grpc/pull/272#discussion_r1057331090 I'll reply in this thread

beligante: Mark this one as fixed.

lib/grpc/client/adapters/mint/stream_response_process.ex
beligante: Hey @polvalente just checking this. When you talk about telemetry you're talking about a general telemetry (aka: `Logger.error(...something)` ) or you're thinking about use the telemetry lib to emit such events?

polvalente: It would be ctual telemetry events. But for this PR we can just add a TODO to deal with the failure case there.

The only issue with the current code is that a dead stream process would kill the entire connection

interop/lib/interop/client.ex
beligante: Solved in: https://github.com/elixir-grpc/grpc/pull/272/commits/fe380234597b700f03560b31245455dff3a367fe

interop/lib/interop/client.ex
beligante: Hey @polvalente! 
I've changed this to behave exactly as Gun adapter. Aka: we don't need that option anymore. Can I mark this one as solved? I'm asking that, cause I have an possible refactor in mind - but it's for after this PR

Tho, I found out that that the `return_headers` option is kinda messy. Basically we have two  [1](https://github.com/elixir-grpc/grpc/blob/master/lib/grpc/stub.ex#L236) [2](https://github.com/elixir-grpc/grpc/blob/master/lib/grpc/stub.ex#L367) places where we can pass that each one is used to return headers for (1) unary / client stream (2) bidirectional_stream / server stream

But the (2) also changes the behavior for trailers. IMO we should have only one of them defining this behavior.

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
beligante: But for safety I've added a case for safety

beligante:
@polvalente I think this one is good to go. There is one this one comment that I've solved, but I like your input about what I saw: https://github.com/elixir-grpc/grpc/pull/272#discussion_r1059002072

Also, dialyzer is breaking for 24.x because of cache. Since my last [successful build](https://github.com/elixir-grpc/grpc/actions/runs/3787496201/jobs/6439353368) til today the latest OTP 24 version was updated from `OTP-24.3.4.6` to `OTP-24.3.4.7` thoughts on pin that version to use an exact match?

polvalente:
> @polvalente I think this one is good to go. There is one this one comment that I've solved, but I like your input about what I saw: [#272 (comment)](https://github.com/elixir-grpc/grpc/pull/272#discussion_r1059002072)
> 
> Also, dialyzer is breaking for 24.x because of cache. Since my last [successful build](https://github.com/elixir-grpc/grpc/actions/runs/3787496201/jobs/6439353368) til today the latest OTP 24 version was updated from `OTP-24.3.4.6` to `OTP-24.3.4.7` thoughts on pin that version to use an exact match?
> 
> Edit: Found out that you can also delete the cache manually: https://github.com/elixir-grpc/grpc/actions/caches lol

TIL about manually deleting cache entries!

interop/lib/interop/client.ex
polvalente: Cool! We can discuss improvements on a separate issue/PR for sure!

Regarding the `return_headers` option, it seems that the only problem might be the naming. Trailers are returned in both places whenever the option is activated.

Also, I think that [1](https://github.com/elixir-grpc/grpc/blob/master/lib/grpc/stub.ex#L236) shouldn't be a doc-exposed function anyway. Perhaps turn that doc to a comment and set `@doc false` for the `def call` function. It's only public because `use GRPC.Stub` injects a call to it.

interop/lib/interop/client.ex
```diff
@@ -169,6 +161,25 @@ defmodule Interop.Client do
...
+  @doc """
+  We build the Stream struct (res_enum) using Stream.unfold/2
+  the unfold function is built in such a way - for both adapters - that the acc is map used to find a
+  connection_stream process and the next_fun arg is a function that reads directly from the connection_stream
+  that is producing data.
+  Every time we execute the next_fun we read a chunk of data and remove that from the buffer.
+  That action by itself will generate a side effect that will update the state of the connection_stream by removing the chunk of data we just read.
+  An easier way to visualize that is. Take the code and the execution bellow as an example
+
+
```
polvalente, line 173:
```suggestion
  We build the Stream struct (`res_enum` in the code) using `Stream.unfold/2`.
  
  The unfold function is built in such a way that  - for both adapters - the accumulator is a map used to find the
  `connection_stream`process and the `next_fun` argument is a function that reads directly from the `connection_stream`
  that is producing data.
  Every time we execute `next_fun` we read a chunk of data. This means that `next_fun` will have the side effect of updating the state of the `connection_stream` process, removing the chunk of data that's being read from the underlying `GenServer`'s state.
```

interop/lib/interop/client.ex
```diff
@@ -169,6 +161,25 @@ defmodule Interop.Client do
...
+  ```
+  iex(4)> ex_stream |> Stream.take(1) |> Enum.to_list()
+  [1]
+  iex(5)> ex_stream |> Enum.to_list()
+  [2, 3]
+  iex(6)> ex_stream |> Enum.to_list()
+  []
+  ```
```
polvalente, line 181:
```suggestion
  ```

  ## Examples

      iex> ex_stream |> Stream.take(1) |> Enum.to_list()
      [1]
      iex> ex_stream |> Enum.to_list()
      [2, 3]
      iex> ex_stream |> Enum.to_list()
      []
  ```
```

polvalente: By the way, this example actually got me thinking if there's a way to make the stream read-only by default and have it become stateful only by doing something more explicit to it.

lib/grpc/stub.ex
```diff
@@ -373,14 +373,38 @@ defmodule GRPC.Stub do
...
+      {:ok, ex_stream} = GRPC.Stub.recv(stream)
+      replies = Enum.map(ex_stream, fn({:ok, reply}) -> reply end)
 
   ## Options
 
     * `:timeout` - request timeout
     * `:deadline` - when the request is timeout, will override timeout
     * `:return_headers` - when true, headers will be returned.
+
+  ## Stream behavior
```
polvalente, line 385: Perhaps move that suggestion+doc from `interop` into here and just point here from there.

polvalente (commented):
I think we're getting to the final rounds here! I just left some comments improving docs and a last question regarding the stream.

After that's answered I'll do a final pass through the whole PR. I don't think we missed anything, but since this PR went through many iterations, it's best to be safe.

interop/lib/interop/client.ex
beligante: For both adapters yes, tho, I would need to decouple the connection stream from the unfold function. That by itself would require huge refactor :/ We could add address that in a incoming PR

lib/grpc/stub.ex
beligante: Sure! I'll make that change

beligante: Added that doc to this function and added a general comment on the Interop.Client module . Since streams are broadly used on that module. 

https://github.com/elixir-grpc/grpc/pull/272/commits/55cbc28e3a9903ecd313e7dbb3919359c9a29ca0

interop/lib/interop/client.ex
beligante: yeah! Agreed! We can include the option renaming when addressing these changes

beligante: I'll do the doc change as you suggested

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -0,0 +1,395 @@
...
+        Logger.debug(fn -> "Received unknown message: " <> inspect(message) end)
+        {:noreply, state}
+
+      {:ok, conn, responses} ->
+        state = State.update_conn(state, conn)
+
+        state =
+          if state.requests == %{},
+            do: state,
+            else: Enum.reduce(responses, state, &process_response/2)
```
polvalente, line 187: Just to avoid map-equality, which can be a bit misleading:

```suggestion
          case state.requests do
            requests when map_size(requests) == 0 ->
              state
            _ ->
              Enum.reduce(responses, state, &process_response/2)
          end
```

Also, I wonder if this isn't a race condition for a response arriving before the corresponding request is saved.
Just to confirm it isn't, a request is sent by this same process, right? If so, this isn't a race condition because the process will process messages sequentially anyway.

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -0,0 +1,395 @@
...
+    window_size = get_window_size(state.conn, ref)
+    body_size = IO.iodata_length(body)
+    dequeued_state = State.update_request_stream_queue(state, queue)
+
+    cond do
+      # Do nothing, wait for server (on stream/2) to give us more window size
+      window_size == 0 -> {:noreply, state}
+      body_size > window_size -> chunk_body_and_enqueue_rest(request, dequeued_state)
+      true -> stream_body_and_reply(request, dequeued_state)
+    end
```
polvalente, line 210:
```suggestion
    window_size = get_window_size(state.conn, ref)
    dequeued_state = State.update_request_stream_queue(state, queue)

    cond do
      # Do nothing, wait for server (on stream/2) to give us more window size
      window_size == 0 -> {:noreply, state}
      IO.iodata_length(body) > window_size -> chunk_body_and_enqueue_rest(request, dequeued_state)
      true -> stream_body_and_reply(request, dequeued_state)
    end
```

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -0,0 +1,395 @@
...
+    :ok =
+      state
+      |> State.stream_response_pid(request_ref)
+      |> StreamResponseProcess.consume(:data, new_data)
+
+    state
+  end
+
+  defp process_response({:done, request_ref}, state) do
+    :ok =
```
polvalente, line 252: A non-blocking aside: I wonder if all of these hard-matches might cause one request in the connection to kill all other requests if something goes wrong. Really depends on what the pipe functions do.

Any insights on this @beligante?

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -0,0 +1,395 @@
...
+        new_state =
+          state
+          |> State.update_conn(conn)
+          |> State.update_request_stream_queue(queue)
+
+        {:noreply, new_state}
+
+      {:error, conn, error} ->
+        if from != nil do
+          GenServer.reply(from, {:error, error})
```
polvalente, line 277:
```suggestion
          # We need an explicit reply here because the process that called this GenServer
          # isn't the same one that's expecting the reply.
          GenServer.reply(from, {:error, error})
```

If it is, we can refactor this into two separate clauses of the `case` to remove some indentation and use `{:reply, ...` for one of the clauses instead.

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -0,0 +1,395 @@
...
+            |> State.stream_response_pid(request_ref)
+            |> StreamResponseProcess.consume(:error, error)
+        end
+
+        {:noreply, State.update_conn(state, conn)}
+    end
+  end
+
+  defp stream_body_and_reply({request_ref, body, from}, state) do
+    send_eof? = from == nil
```
polvalente, line 290:
```suggestion
    send_eof? = is_nil(from)
```

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -0,0 +1,395 @@
...
+
+    case stream_body(state.conn, request_ref, body, send_eof?) do
+      {:ok, conn} ->
+        if not is_nil(from) do
```
polvalente, line 294:
```suggestion
        if not send_eof? do
```

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -0,0 +1,395 @@
...
+          GenServer.reply(from, :ok)
+        end
+
+        check_request_stream_queue(State.update_conn(state, conn))
+
+      {:error, conn, error} ->
+        if not is_nil(from) do
```
polvalente, line 301:
```suggestion
        if not send_eof? do
```

polvalente: Same question about who is expecting the reply as well

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -0,0 +1,395 @@
...
+
+        check_request_stream_queue(State.update_conn(state, conn))
+    end
+  end
+
+  defp stream_body(conn, request_ref, body, true = _stream_eof?) do
+    with {:ok, conn} <- Mint.HTTP.stream_request_body(conn, request_ref, body),
+         {:ok, conn} <- Mint.HTTP.stream_request_body(conn, request_ref, :eof) do
+      {:ok, conn}
+    end
```
polvalente, line 318:
```suggestion
    case Mint.HTTP.stream_request_body(conn, request_ref, body) do
      {:ok, conn} -> Mint.HTTP.stream_request_body(conn, request_ref, :eof)
      error -> error
    end
```

polvalente: This is cosmetic only

lib/grpc/client/adapters/mint/stream_response_process.ex
```diff
@@ -0,0 +1,199 @@
...
+
+  @spec start_link(GRPC.Client.Stream.t(), send_headers_or_trailers? :: boolean()) ::
+          GenServer.on_start()
+  def start_link(stream, send_headers_or_trailers?) do
+    GenServer.start_link(__MODULE__, {stream, send_headers_or_trailers?})
+  end
+
+  @doc """
+  Given a pid from this process, build an Elixir.Stream that will consume the accumulated
+    data inside this process
```
polvalente, line 27:
```suggestion
  data inside this process
```

lib/grpc/client/adapters/mint/stream_response_process.ex
```diff
@@ -0,0 +1,199 @@
+defmodule GRPC.Client.Adapters.Mint.StreamResponseProcess do
+  @moduledoc false
+  # This module represents the process responsible for consuming the
+  # incoming messages from a connection. For each request, there will be
+  # a process responsible for consuming its messages. At the end of a stream
+  # this process will automatically be killed.
+
+  # TODO: Refactor the GenServer.call/3 occurrences on this module to produce
+  # telemetry errors in case of failures
```
polvalente, line 9:
```suggestion
  # telemetry events and log entries in case of failures
```

lib/grpc/client/adapters/mint/stream_response_process.ex
```diff
@@ -0,0 +1,199 @@
...
+        %{send_headers_or_trailers: false, responses: responses} = state
+      )
+      when type in @header_types do
+    state = update_compressor({type, headers}, state)
+
+    with {:error, _rpc_error} = error <- get_headers_response(headers, type) do
+      {:reply, :ok, %{state | responses: [error | responses]}, {:continue, :produce_response}}
+    else
+      _any -> {:reply, :ok, state, {:continue, :produce_response}}
+    end
```
polvalente, line 134:
```suggestion
    case get_headers_response(headers, type) do
      {:error, _rpc_error} = error ->
        {:reply, :ok, %{state | responses: [error | responses]}, {:continue, :produce_response}}
      _any ->
        {:reply, :ok, state, {:continue, :produce_response}}
    end
```

lib/grpc/client/adapters/mint/stream_response_process.ex
```diff
@@ -0,0 +1,199 @@
...
+  # TODO: Refactor the GenServer.call/3 occurrences on this module to produce
+  # telemetry errors in case of failures
+
+  @typep accepted_types :: :data | :trailers | :headers | :error
+  @typep data_types :: binary() | Mint.Types.headers() | Mint.Types.error()
+
+  @accepted_types [:data, :trailers, :headers, :error]
+  @header_types [:headers, :trailers]
+
+  use GenServer
```
polvalente, line 17: Aren't we missing a `@behaviour GenServer` here?

I'd expect `handle_continue` & friends to warn on missing `@impl true` below

lib/grpc/message.ex
```diff
@@ -172,4 +172,28 @@ defmodule GRPC.Message do
...
+    end
+  end
+
+  def get_message(data, compressor) do
+    case data do
+      <<1::8, length::unsigned-integer-32, message::bytes-size(length), rest::binary>> ->
+        {{1, compressor.decompress(message)}, rest}
+
+      <<0::8, length::unsigned-integer-32, message::bytes-size(length), rest::binary>> ->
+        {{0, message}, rest}
```
polvalente, line 193: just to make sure, we don't need to decompress this message here, right? It's what the first bit is saying

lib/grpc/stub.ex
```diff
@@ -218,25 +218,24 @@ defmodule GRPC.Stub do
...
-    * `:timeout` - request timeout. Default is 10s for unary calls and `:infinity` for
-      client or server streaming calls
-    * `:deadline` - when the request is timeout, will override timeout
-    * `:metadata` - a map, your custom metadata
-    * `:return_headers` - default is false. When it's true, a three elem tuple will be returned
-      with the last elem being a map of headers `%{headers: headers, trailers: trailers}`(unary) or
-      `%{headers: headers}`(server streaming)
-  """
+  @doc false
+  #  The actual function invoked when invoking a rpc function.
```
polvalente, line 222:
```suggestion
  #  The actual function invoked when invoking an RPC function.
```

lib/grpc/stub.ex
```diff
@@ -218,25 +218,24 @@ defmodule GRPC.Stub do
...
+  #  Options
+  #
+  #    * `:timeout` - request timeout. Default is 10s for unary calls and `:infinity` for
+  #      client or server streaming calls
+  #    * `:deadline` - when the request is timeout, will override timeout
+  #    * `:metadata` - a map, your custom metadata
+  #    * `:return_headers` - default is false. When it's true, a three elem tuple will be returned
+  #      with the last elem being a map of headers `%{headers: headers, trailers: trailers}`(unary) or
+  #      `%{headers: headers}`(server streaming)
   @spec call(atom(), tuple(), GRPC.Client.Stream.t(), struct() | nil, keyword()) :: rpc_return
```
polvalente, line 239:
```suggestion
  # @spec call(atom(), tuple(), GRPC.Client.Stream.t(), struct() | nil, keyword()) :: rpc_return
```

Let's hide the spec as well, or even remove it

lib/grpc/stub.ex
```diff
@@ -373,14 +372,43 @@ defmodule GRPC.Stub do
...
+  ## Examples
+
+      iex> ex_stream |> Stream.take(1) |> Enum.to_list()
+      [1]
+      iex> ex_stream |> Enum.to_list()
+      [2, 3]
+      iex> ex_stream |> Enum.to_list()
+      []
+
+  Unlikely the usual stream behavior you can see bellow.
```
polvalente, line 400: I couldn't understand this sentence

lib/grpc/stub.ex
```diff
@@ -373,14 +372,43 @@ defmodule GRPC.Stub do
...
     * `:deadline` - when the request is timeout, will override timeout
     * `:return_headers` - when true, headers will be returned.
+
+  ## Stream behavior
+  We build the Stream struct using `Stream.unfold/2`.
+
+  The unfold function is built in such a way that  - for both adapters - the accumulator is a map used to find the
+  `connection_stream`process and the `next_fun` argument is a function that reads directly from the `connection_stream`
+  that is producing data.
+  Every time we execute `next_fun` we read a chunk of data. This means that `next_fun` will have the side effect of updating the state of the `connection_stream` process, removing the chunk of data that's being read from the underlying `GenServer`'s state.
```
polvalente, line 390:
```suggestion
  Every time we execute `next_fun` we read a chunk of data. This means that `next_fun` will have the side effect of updating the state of the `connection_stream` process, removing the chunk of data that's being read from the underlying `GenServer`'s state.

```

polvalente (commented):
Awesome work! I left a final batch of comments.
Most are stylistic, but there are some questions as well (which don't seem to block this PR further!)

Once we resolve this last batch we can merge and open an issue to tackle all of the TODOs we left.

interop/lib/interop/client.ex
polvalente: If we haven't already, let's leave a to-do and resolve this convo :)

lib/grpc/client/adapters/mint.ex
```diff
@@ -0,0 +1,212 @@
...
+  end
+
+  defp return_headers_for_request?(%GRPC.Client.Stream{grpc_type: type}, _opts)
+       when type in [:bidirectional_stream, :server_stream] do
+    true
+  end
+
+  defp return_headers_for_request?(_stream, opts) do
+    # Explicitly check for true to ensure the boolean type here
+    opts[:return_headers] == true or false
```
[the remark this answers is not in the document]
beligante: In my defense I had two beers when I did that hahahaha

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
beligante: Gotcha! (for the map - I just read about that on elixir forum - TIL) 
For the race condition: yeah. That's not a problem we send the request by this same process indeed.

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
beligante: I could - if that's okay with out - remove these and leave the telemetry stuff only for the response process.  This way the connection process will simply be responsible for distribute the data and ignore errors related to process it

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
beligante: For RPCs that has any sort of client streaming, when this [function is called](https://github.com/elixir-grpc/grpc/blob/master/lib/grpc/client/stream.ex#L91) we need to synchronously wait for that piece of request to finish it's stream before we can continue with the execution. So, I need to reply the caller process so that it can continue with the execution.

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
beligante: The way I modeled this unfortunately we can't. 

Since I had to came up with the client stream queue to avoid blow the window size error. I can't reply the caller process directly - that means reply directly on the `handle_call` ) because I need to always check the stream queue (which I use a `handle_continue` ) before I send the data to the server. 

I could refactor to a private function if you think that would be cleaner

lib/grpc/client/adapters/mint/stream_response_process.ex
beligante: yeah! that's strange indeed hahaha. I'll annotate the `@impl` on the bellow  tho

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
sleipnir: > The way I modeled this unfortunately we can't.
> 
> Since I had to came up with the client stream queue to avoid blow the window size error. I can't reply the caller process directly - that means reply directly on the `handle_call` ) because I need to always check the stream queue (which I use a `handle_continue` ) before I send the data to the server.
> 
> I could refactor to a private function if you think that would be cleaner

Perhaps you could return :noreply and handle what another process needs via spawn instead of a handle call. Maybe?

lib/grpc/message.ex
beligante: yup! that's what the first byte is about. I've looked at this doc for ref: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md

lib/grpc/stub.ex
beligante: I think I'll remove this.

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
beligante: not sure if I followed your suggestion.
but in here I'm returning `:noreply` and [en-queuing the request](https://github.com/beligante/grpc/blob/add-mint-adapter/lib/grpc/client/adapters/mint/connection_process/connection_process.ex#L158)

sleipnir: I've looked at the code more closely now and it would be hard to follow my suggestion without having to change a lot.

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
polvalente: I'd rather have them explicitly match if we ensure this function only returns `:ok`
If errors should be ignored, they shouldn't be returned.
If this is a refactor, we can leave a to-do for a future PR

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
polvalente: My suggestion up there still holds, though :)

polvalente: (the actual code snippet, I mean)

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
beligante: Gotcha! Then we're good here. This will always return `:ok`

lib/grpc/client/adapters/mint.ex
```diff
@@ -26,11 +26,11 @@ defmodule GRPC.Client.Adapters.Mint do
         {:ok, %{channel | adapter_payload: %{conn_pid: pid}}}
 
       error ->
-        {:error, "An error happened while trying to opening the connection: #{inspect(error)}"}
+        {:error, "Error when opening connection: #{inspect(error)}"}
```
polvalente, line 29:
```suggestion
        {:error, "Error while opening connection: #{inspect(error)}"}
```

lib/grpc/client/adapters/mint.ex
```diff
@@ -26,11 +26,11 @@ defmodule GRPC.Client.Adapters.Mint do
...
     end
   catch
     :exit, reason ->
-      {:error, "An error happened while trying to opening the connection: #{inspect(reason)}"}
+      {:error, "Error when opening connection: #{inspect(reason)}"}
```
polvalente, line 33:
```suggestion
      {:error, "Error while opening connection: #{inspect(reason)}"}
```

lib/grpc/client/adapters/mint.ex
```diff
@@ -160,7 +160,7 @@ defmodule GRPC.Client.Adapters.Mint do
   end
 
   def handle_errors_receive_data(%GRPC.Client.Stream{payload: %{response: response}}, _opts) do
-    {:error, "an error occurred while when receiving data: error=#{inspect(response)}"}
+    {:error, "Error occurred when receiving data: #{inspect(response)}"}
```
polvalente, line 163:
```suggestion
    {:error, "Error occurred while receiving data: #{inspect(response)}"}
```

test/grpc/client/adapters/mint_test.exs
```diff
@@ -0,0 +1,36 @@
...
+    test "connects insecurely (custom options)", %{port: port} do
+      channel = build(:channel, port: port, host: "localhost")
+
+      assert {:ok, result} = Mint.connect(channel, transport_opts: [ip: :loopback])
+      assert %{channel | adapter_payload: %{conn_pid: result.adapter_payload.conn_pid}} == result
+
+      # Ensure that changing one of the options breaks things
+      assert {:error, message} = Mint.connect(channel, transport_opts: [ip: "256.0.0.0"])
+
+      assert message == "Error when opening connection: {:error, :badarg}"
```
polvalente, line 33:
```suggestion
      assert message == "Error while opening connection: {:error, :badarg}"
```

polvalente (approved):

polvalente:
@tony612 I'm merging this so we can have the current state of the adapter as a clean slate. We can discuss next iterations on a future issue/PR if needed.

There are already some TODO comments that will be addressed

beligante:
Opened an Issue for one of the TODO's: https://github.com/elixir-grpc/grpc/issues/291

tony612:
@polvalente  Got it. Please ping me if there's anything needed from me.

Nice to see this got merged!

PR #285, opened 2022-11-16
Upgrade to protobuf 0.11

08525182818:
This upgrades the protobuf dependency to 0.11 which is notable as it added support for proto3 `optional` fields.


08525182818:
Sorry, used wrong account to make the PR.

PR #292, opened 2023-01-11
fix: restore deprecated interceptor modules

wingyplus:
I found the compilation warning in interop tests since PR #289 was merged. Because of GRPC Prometheus didn't change to use the new interceptor module. And it's suddenly breaking changes. So I propose to bring it back but add the deprecated module to it. The rest topic is discussing when to remove it. :)

### DIFF
--- a/lib/grpc/client/interceptor.ex
+++ b/lib/grpc/client/interceptor.ex
@@ -1,3 +1,20 @@
+defmodule GRPC.ClientInterceptor do
+  @moduledoc """
+  Interceptor on client side. See `GRPC.Stub.connect/2`.
+  """
+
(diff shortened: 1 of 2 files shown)


polvalente:
Although this makes sense, I'm not sure if the overhead of maintaining both versions is worth it.
I think a PR for grpc-prometheus updating it instead is the better approach.

polvalente:
We'll probably deprecate the Prometheus library :)

I'll merge this. I think it's nice to warn for at least 1 version on the deprecation.

wingyplus:
> We'll probably deprecate the Prometheus library :)
> 
> I'll merge this. I think it's nice to warn for at least 1 version on the deprecation.

Agreed. :)

PR #300, opened 2023-02-07
Use original `gun` dependency instead of `grpc_gun`

general-CbIC:
I noticed that gun 2.0.0 [was released](https://hex.pm/packages/gun) in January. Maybe we want to use it again?


mix.exs
```diff
@@ -40,11 +40,9 @@ defmodule GRPC.Mixfile do
   defp deps do
     [
       {:cowboy, "~> 2.9"},
-      # This is the same as :gun 2.0.0-rc.2,
-      # but we can't depend on an RC for releases
-      {:gun, "~> 2.0.1", hex: :grpc_gun},
+      {:gun, "~> 2.0.0"},
       {:mint, "~> 1.4.2"},
-      {:cowlib, "~> 2.11"},
+      {:cowlib, "~> 2.12", override: true},
```
polvalente, line 45: Why do we need to set `override: true` here?

When I tried updating I ran into an incompatibility with cowboy which you seem to have worked around. So I don't see the need for the override.

general-CbIC: Gun has dependency version 2.12.0: https://github.com/ninenines/gun/blob/master/rebar.config#L2
Cowboy depends on 2.11.0: https://github.com/ninenines/gun/blob/master/rebar.config#L2

When I remove `override`, I have:

```
Resolving Hex dependencies...
Resolution completed in 0.024s
Because your app depends on cowboy ~> 2.9 which depends on cowlib 2.11.0, cowlib 2.11.0 is required.
So, because your app depends on cowlib ~> 2.12, version solving failed.
** (Mix) Hex dependency resolution failed
```

polvalente: Yeah, I don't think we should update this now then.

Next release will include the new Mint adapter, so that's more of a reason to not worry about this.

I opened an issue about this on the Gun repo nevertheless.

polvalente:
Thanks for the PR! Closing this due to cowboy+gun incompatibility

mix.exs
general-CbIC: Ok, thanks!

PR #301, opened 2023-02-10
Update gun to 2.0

carrascoacd:
## Context

Gun 2.0 has been released. As a consequence, the dependency resolution when using https://github.com/elixir-grpc/grpc is broken.

In order to remove the temporal fork of `gun` and use the official `gun` library, we want to make sure that:
1. `gun` depends on `cowlib` 2.12.0
2. `cowboy` depends on `cowlib` 2.12.0
3. `grpc` depends indirectly on on `cowlib` 2.12.0

I'm also taking the chance for releasing the minor version of the dependency so we are not so strict.

Related to https://github.com/ninenines/cowboy/pull/1601 @essen @polvalente


mix.exs
```diff
@@ -40,11 +40,10 @@ defmodule GRPC.Mixfile do
   defp deps do
     [
       {:cowboy, "~> 2.9"},
-      # This is the same as :gun 2.0.0-rc.2,
-      # but we can't depend on an RC for releases
-      {:gun, "~> 2.0.1", hex: :grpc_gun},
+      {:gun, "~> 2.0"},
       {:mint, "~> 1.4.2"},
-      {:cowlib, "~> 2.11"},
+      # This is temporal only for code review until https://github.com/ninenines/cowboy/pull/1601 is merged
```
carrascoacd, line 45: Please, do not merge until `cowboy` is released, this is for the sake of the PR review.

polvalente: We can't actually release anything with `override: true` in Hex, so I'll leave a "request changes" on this :)

polvalente (changes requested):

mix.exs
carrascoacd: Sure! This change is on purpose for the shake of code review until the depending library is released. Then, we just need to change this. As you said this is a request change :)

nulian:
cowboy with cowlib is finally updated so this can be merged.

polvalente:
@carrascoacd (cc @nulian) The tests are currently failing on CI due to some SSL validation issues. Any ideas on how to fix that? Locally everything works.

carrascoacd:
I'm afraid it is because `certs_keys` were added in OTP 25 so probably the current tests are not working. I've added a commit to try to solve it but now the previous from 25 failed. I think we will need to create different tests to distinguish it and change the CI configuration to run them depending on the version.

carrascoacd:
Yes, it is quite weird. Maybe I'm wrong and it is backwards compatible. Can you check the runtime GH uses in the tests to use exactly that instead of asdf based?

polvalente:
Great news, it seems that the SSL issues appear on Erlang 25.3, so using 25.1.x (which is the same I had locally) solves this for now, so we'll be able to merge

mix.exs
```diff
@@ -39,14 +39,13 @@ defmodule GRPC.Mixfile do
...
     [
-      {:cowboy, "~> 2.9"},
-      # This is the same as :gun 2.0.0-rc.2,
-      # but we can't depend on an RC for releases
-      {:gun, "~> 2.0.1", hex: :grpc_gun},
+      {:cowboy, "~> 2.10"},
+      {:gun, "~> 2.0"},
       {:mint, "~> 1.4.2"},
-      {:cowlib, "~> 2.11"},
+      # This is temporal only for code review until https://github.com/ninenines/cowboy/pull/1601 is merged
```
polvalente, line 45:
```suggestion
```

README.md
```diff
@@ -11,17 +11,13 @@ An Elixir implementation of [gRPC](http://www.grpc.io/).
...
-> __Note__
-> The [Gun](https://github.com/ninenines/gun) library doesn't have a full 2.0 release yet, so we depend on `:grcp_gun 2.0.1` for now.
-This is the same as `:gun 2.0.0-rc.2`, but [Hex](https://hex.pm/) doesn't let us depend on RC versions for releases.
+- [gRPC Elixir](#grpc-elixir)
+  - [Table of contents](#table-of-contents)
+  - [Installation](#installation)
+  - [Usage](#usage)
+  - [Features](#features)
+  - [Benchmark](#benchmark)
+  - [Contributing](#contributing)
```
polvalente, line 20:
```suggestion
- [Notice](#notice)
- [Installation](#installation)
- [Usage](#usage)
- [Features](#features)
- [Benchmark](#benchmark)
- [Contributing](#contributing)
```

carrascoacd:
Great! Good job

Issue #305, opened 2023-03-10
Support `cowboy`/`ranch_tcp` opts: `inet`, `inet6`, `ipv6_only`


Nezteb:
**Is your feature request related to a problem? Please describe.**
`elixir-grpc` uses `cowboy`, which uses `ranch` for listening on TCP sockets. `ranch` supports several config options like `inet`, `inet6`, and `ipv6_only`. https://ninenines.eu/docs/en/ranch/1.7/manual/ranch_tcp/

From what I can tell, it seems like `elixir-grpc` strips out most `cowboy` configuration options here: https://github.com/elixir-grpc/grpc/blob/f21e2e1bcf48c3a2991b61ed8675e9159fa5872c/lib/grpc/server/adapters/cowboy.ex#L222-L239

I'm working on a service that uses `elixir-grpc` that we want to listen with both IPv4 and IPv6, and eventually move to only supporting IPv6 communication. Given the above, I don't think that's currently possible with `elixir-grpc`.

**Describe the solution you'd like**
I might be wrong; there might be a way to support both IPv4 and IPv6 (as well as only IPv6) already in `elixir-grpc` with some configuration tweaking. I just can't find any obvious way to do that.

`plug_cowboy` supports these options: https://hexdocs.pm/plug_cowboy/Plug.Cowboy.html#module-options

The difference seems to be that `plug_cowboy` doesn't filter out options: https://github.com/elixir-plug/plug_cowboy/blob/1796e77aa42c2284f3667c7aeac7074236f47c07/lib/plug/cowboy.ex#L254-L302

**Describe alternatives you've considered**
N/A.

**Additional context**
N/A.

polvalente:
A PR extending the current behavior is welcome!

Nezteb:
@polvalente I've pulled in the `master` branch of `grpc` as a dependency to my project and realized: I don't know if I can configure the Cowboy adapter from the project-level.

Previously I had:
```elixir
children = [
  # ...
  {GRPC.Server.Supervisor, endpoint: Helloworld.Endpoint, port: 50051}
  # ...
]
```

I think I just need to do the following?
```elixir
children = [
  # ...
  {GRPC.Server.Supervisor, endpoint: Helloworld.Endpoint, port: 50051, net: :inet6, ipv6_v6only: true}
  # ...
]
```

Does that seem right?

polvalente:
@Nezteb yup, looks like it's right!

[...]

PR #308, opened 2023-03-13
Add support for ranch_tcp opts: ip, ipv6_v6only, inet, inet6

Nezteb:
Addresses https://github.com/elixir-grpc/grpc/issues/305

It's a bit awkward because `ranch_tcp`'s listen options are all keyword tuples _except_ for `:inet` and `:inet6`, so I borrowed `plug_cowboy`'s method of passing the network atoms: https://github.com/elixir-plug/plug_cowboy/blob/v2.6.0/lib/plug/cowboy.ex#L360-L361

### DIFF
--- a/lib/grpc/server/adapters/cowboy.ex
+++ b/lib/grpc/server/adapters/cowboy.ex
@@ -224,7 +224,7 @@ defmodule GRPC.Server.Adapters.Cowboy do
 
     # https://ninenines.eu/docs/en/ranch/1.7/manual/ranch_tcp/
     socket_opts =
-      Enum.reduce(opts, socket_opts, fn 
+      Enum.reduce(opts, socket_opts, fn
         {k, v}, acc when k in [:ip, :ipv6_v6only] and not is_nil(v) -> [{k, v} | acc]
         {:net, v}, acc when not is_nil(v) -> [v | acc]
         _, acc -> acc
--- a/lib/grpc/server/adapters/cowboy.ex
+++ b/lib/grpc/server/adapters/cowboy.ex
@@ -221,7 +221,22 @@ defmodule GRPC.Server.Adapters.Cowboy do
 
   defp socket_opts(port, opts) do
     socket_opts = [port: port]
-    socket_opts = if opts[:ip], do: [{:ip, opts[:ip]} | socket_opts], else: socket_opts
+
+    # https://ninenines.eu/docs/en/ranch/1.7/manual/ranch_tcp/
+    allowed_ranch_opts = %{
+      ip: {:ip, opts[:ip]},
+      ipv6_v6only: {:ipv6_v6only, opts[:ipv6_v6only]},
+      net: opts[:net]
+    }
+
+    socket_opts =
+      Enum.reduce(allowed_ranch_opts, socket_opts, fn {key, value}, acc ->
+        if opts[key] != nil do
+          [value | acc]
+        else
+          acc
+        end
+      end)
 
     if opts[:cred] do
       opts[:cred].ssl ++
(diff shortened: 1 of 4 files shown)


lib/grpc/server/adapters/cowboy.ex
```diff
@@ -221,7 +221,22 @@ defmodule GRPC.Server.Adapters.Cowboy do
...
+    }
+
+    socket_opts =
+      Enum.reduce(allowed_ranch_opts, socket_opts, fn {key, value}, acc ->
+        if opts[key] != nil do
+          [value | acc]
+        else
+          acc
+        end
+      end)
```
polvalente, line 239:
```suggestion
    socket_opts = 
      Enum.reduce(opts, socket_opts, fn 
        {k, v}, acc when k in [:ip, :ipv6_v6only] and not is_nil(v) -> [{k, v} | acc]
        {:net, v}, acc when not is_nil(v) -> [v | acc]
        _, acc -> acc
      end)
```

Your code is not wrong, but this suggestion does it in a single pass. Note that yours does `opts[key]`, which is slow because of list lookup, for each entry.

lib/grpc/server/adapters/cowboy.ex
```diff
@@ -221,7 +221,14 @@ defmodule GRPC.Server.Adapters.Cowboy do
 
   defp socket_opts(port, opts) do
     socket_opts = [port: port]
-    socket_opts = if opts[:ip], do: [{:ip, opts[:ip]} | socket_opts], else: socket_opts
+
+    # https://ninenines.eu/docs/en/ranch/1.7/manual/ranch_tcp/
+    socket_opts =
```
polvalente, line 226:
```suggestion
    socket_opts =
```

polvalente:
Please run the formatter so we can merge :)

Nezteb:
Screenshot 2023-03-14 at 12 50 54 AM
@polvalente That's odd, the formatter passes on my machine. 😅

polvalente:
> Screenshot 2023-03-14 at 12 50 54 AM
> @polvalente That's odd, the formatter passes on my machine. 😅 

I think you need to run with Elixir 1.14 (the same as CI)

Nezteb:
> I think you need to run with Elixir 1.14 (the same as CI)

Oh, the `.tool-versions` file specified 1.13 so that's what my `asdf` install was using. 😄

mix.exs
```diff
@@ -7,7 +7,7 @@ defmodule GRPC.Mixfile do
     [
       app: :grpc,
       version: @version,
-      elixir: "~> 1.11",
+      elixir: "~> 1.14",
```
polvalente, line 10: We still support Elixir 1.11, it's just the formatter that changed a few things in 1.14 :)

polvalente, line 10:
```suggestion
      elixir: "~> 1.11",
```

lib/grpc/server/adapters/cowboy.ex
```diff
@@ -221,7 +221,14 @@ defmodule GRPC.Server.Adapters.Cowboy do
...
+      Enum.reduce(opts, socket_opts, fn
+        {k, v}, acc when k in [:ip, :ipv6_v6only] and not is_nil(v) -> [{k, v} | acc]
+        {:net, v}, acc when not is_nil(v) -> [v | acc]
```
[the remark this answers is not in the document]
polvalente: I think you could document on `start` and point `child_spec` to `start`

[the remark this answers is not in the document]
Nezteb: How's this? https://github.com/elixir-grpc/grpc/pull/309 😄

PR #320, opened 2023-06-14
Add start_server opt on README server example

avillen:
This is already [documented](https://github.com/elixir-grpc/grpc/blob/3cbd100c5aaf49720b36b67064e9e38530cacaea/lib/grpc/server/supervisor.ex#L6), but just to avoid some headache copy/pasting from the README.md 😅


polvalente (approved):
Nice

Issue #316, opened 2023-06-14
Bubble errors up from server (and client?) pipelines


polvalente:
From PR #218, we should add a capability for both server (and client?) pipelines to bubble errors up instead of needing to raise for flow control.

avillen:
Do you have something in mind about this? taking a look at the code I think that it may be "as easy" as the following on this [part](https://github.com/elixir-grpc/grpc/blob/master/lib/grpc/server/adapters/cowboy/handler.ex#L378):
```
     result = server.__call_rpc__(path, stream)

     case result do
+      {:ok, _stream, error = %GRPC.RPCError{message: nil, status: status}} ->
+        {:error, %{error | message: GRPC.Status.status_message(status)}}
+
+      {:ok, _stream, error = %GRPC.RPCError{}} ->
+        {:error, error}
+
       {:ok, stream, response} ->
         stream
         |> GRPC.Server.send_reply(response)
```

The first match is because, if I understand correctly, right now the server creates a `GRPC.RPCError` exception that already generates the `:message` if [necessary](https://github.com/elixir-grpc/grpc/blob/master/lib/grpc/rpc_error.ex#L71), so we need to generate it somehow.

I think that this will be retrocompatible, so both exception and "handled" errors will be allowed. WDYT? I can submit a quick PR if you agree 😄 or if you have something different in mind or I'm wrong, please, ignore me 😂

polvalente:
Yeah, it's something along those lines!

I don't worry too much about retrocompatibility in the sense that this is a new code path that didn't exist before, but we should ensure the test coverage is good for the modified code path.

I'm happy to review a PR on this

sleipnir:
@polvalente is still relevant?

polvalente:
Not sure. Probably not given how old the PR is. We can recover the code if needed

PR #321, opened 2023-06-14
Handle not exception errors from the server

avillen:
Related to https://github.com/elixir-grpc/grpc/issues/316#issuecomment-1592115328


lib/grpc/server/adapters/cowboy/handler.ex
```diff
@@ -376,6 +376,12 @@ defmodule GRPC.Server.Adapters.Cowboy.Handler do
     result = server.__call_rpc__(path, stream)
 
     case result do
+      {:ok, _stream, error = %GRPC.RPCError{message: nil, status: status}} ->
+        {:error, %{error | message: GRPC.Status.status_message(status)}}
+
+      {:ok, _stream, error = %GRPC.RPCError{}} ->
+        {:error, error}
+
```
polvalente, line 384: I believe more changes are needed so that we get this errors to bubble up without raising. Currently they still raise on like 364.

The idea is that if we manage to bubble them up, then they'll bypass any exception-handling workflow and they'll also follow all interceptors along instead of cutting the workflow dry after the call

avillen: Surely I'm wrong, but I'm understanding that those exits are because that part of the code is executed with a `spawn_link + Process.flag(:trap_exit, true)`. I have updated the Server module to return an error tuple that in fact is what I think the logger interceptor expects. I'm missing something no?

Issue #325, opened 2023-07-05
Default `accepted_comparators` option of logger interceptors doesn't match Elixir's default behavior


guisehn:
**Describe the bug**

`elixir-grpc`'s interceptors `GRPC.Server.Interceptors.Logger` and `GRPC.Client.Interceptors.Logger` are doing the opposite check of Elixir's default Logger behavior by default, this seems to be a mistake.

**To Reproduce**

Set your application level to `:debug`:

```elixir
config :logger, level: :debug
```

Add the `GRPC.Server.Interceptors.Logger` to your application:

```elixir
defmodule MyApp.Endpoint do
  use GRPC.Endpoint

  intercept GRPC.Server.Interceptors.Logger # `level` option defaults to `:info`, so elixir-grpc will call Logger.info()

  run MyApp.GRPC.Health.Server
end
```

Then, do a gRPC call to your server and you'll notice it will not log anything.

**Expected behavior**

When I set my application log level to `:debug`, I expect it to print all levels of logs, because `:debug` is the lowest log priority possible.

This is how Elixir's `Logger` works by default:

Example:

```elixir
iex(1)> require Logger
Logger
iex(2)> Logger.configure(level: :debug)
:ok
iex(3)> Logger.info("I am being logged because debug < info")

[info]  I am being logged because debug < info
:ok
iex(4)> Logger.configure(level: :warn) 
:ok
iex(5)> Logger.info("I am not being logged because warn > info")
:ok
```

**Additional context**

Both `GRPC.Server.Interceptors.Logger` and `GRPC.Client.Interceptors.Logger` compare the `level` option with `Log.level()`. I assume this is to prevent unnecessary processing. [See code](https://github.com/elixir-grpc/grpc/blob/691ac2146eac1691e703e31985765f042ec5e91a/lib/grpc/server/interceptors/logger.ex#L53).

However, it seems that the comparison is backwards. To match [Elixir's default behavior](https://github.com/elixir-lang/elixir/blob/7f7a8bca99fa306a41a985df0018ba642e577d4d/lib/logger/lib/logger.ex#L1112), we should either 1) reverse the order params are passed to `Logger.compare_levels`, or 2) set `accepted_comparators` default to `[:gt, :eq]`.

Doing this change will restore the default behavior before https://github.com/elixir-grpc/grpc/pull/227 was merged.

PR #326, opened 2023-07-05
Remove `accepted_comparators` option from logger interceptors

guisehn:
Remove `accepted_comparators` option from logger interceptors.

Fixes https://github.com/elixir-grpc/grpc/issues/325

Restores the default behavior before https://github.com/elixir-grpc/grpc/pull/227 was merged

### DIFF
--- a/lib/grpc/client/interceptors/logger.ex
+++ b/lib/grpc/client/interceptors/logger.ex
@@ -14,11 +14,11 @@ defmodule GRPC.Client.Interceptors.Logger do
   ## Usage
 
       {:ok, channel} = GRPC.Stub.connect("localhost:50051", interceptors: [GRPC.Client.Interceptors.Logger])
-      # This will log on `:info` and lower priority
+      # This will log on `:info` and lower priority (debug)
       {:ok, channel} = GRPC.Stub.connect("localhost:50051", interceptors: [{GRPC.Client.Interceptors.Logger, level: :info}])
       # This will log only on `:info`
       {:ok, channel} = GRPC.Stub.connect("localhost:50051", interceptors: [{GRPC.Client.Interceptors.Logger, level: :info, accepted_comparators: [:eq]}])
-      # This will log on `:info` and higher priority
+      # This will log on `:info` and higher priority (warn, error...)
       {:ok, channel} = GRPC.Stub.connect("localhost:50051", interceptors: [{GRPC.Client.Interceptors.Logger, level: :info, accepted_comparators: [:eq, :gt]}])
   """
 
--- a/lib/grpc/client/interceptors/logger.ex
+++ b/lib/grpc/client/interceptors/logger.ex
@@ -14,11 +14,11 @@ defmodule GRPC.Client.Interceptors.Logger do
   ## Usage
 
       {:ok, channel} = GRPC.Stub.connect("localhost:50051", interceptors: [GRPC.Client.Interceptors.Logger])
-      # This will log on `:info` and greater priority
+      # This will log on `:info` and lower priority
       {:ok, channel} = GRPC.Stub.connect("localhost:50051", interceptors: [{GRPC.Client.Interceptors.Logger, level: :info}])
       # This will log only on `:info`
       {:ok, channel} = GRPC.Stub.connect("localhost:50051", interceptors: [{GRPC.Client.Interceptors.Logger, level: :info, accepted_comparators: [:eq]}])
-      # This will log on `:info` and lower priority
+      # This will log on `:info` and higher priority
       {:ok, channel} = GRPC.Stub.connect("localhost:50051", interceptors: [{GRPC.Client.Interceptors.Logger, level: :info, accepted_comparators: [:eq, :gt]}])
   """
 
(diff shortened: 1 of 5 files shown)


lib/grpc/client/interceptors/logger.ex
```diff
@@ -14,11 +14,11 @@ defmodule GRPC.Client.Interceptors.Logger do
   ## Usage
 
       {:ok, channel} = GRPC.Stub.connect("localhost:50051", interceptors: [GRPC.Client.Interceptors.Logger])
-      # This will log on `:info` and greater priority
+      # This will log on `:info` and lower priority (debug)
```
polvalente, line 17: I think we should change the default to be `[:gt, :eq]` to keep with the default Logger behavior

lib/grpc/server/interceptors/logger.ex
```diff
@@ -23,7 +23,7 @@ defmodule GRPC.Server.Interceptors.Logger do
         use GRPC.Endpoint
 
         # logs on :info and higher priority (warn, error...)
-        intercept GRPC.Server.Interceptors.Logger, level: :info, accepted_comparators: [:lt, :eq]
+        intercept GRPC.Server.Interceptors.Logger, level: :info, accepted_comparators: [:gt, :eq]
```
polvalente, line 26: In changing the default, we'll be able to remove the `accepted_comparators` altogether here

lib/grpc/client/interceptors/logger.ex
guisehn: That's a valid approach too. My main concern is how the API would read to users.

With the change I did in this PR, the API reads like this:

- `[:lt, :eq]` means it will log on `:info` and lower priority (`lt` = `lower`)
- `[:gt, :eq]` means it will log on `:info` and higher priority (`gt` = `higher`)

Whereas if we just change the default to `[:gt, :eq]` and keep the order on `Logger.compare_levels`, the API reads like:

- `[:gt, :eq]` means it will log on `:info` and lower priority (but `gt` means higher)
- `[:lt, :eq]` means it will log on `:info` and higher priority (but `lt` means higher)

I think the first keeps things more clear. WDYT?

polvalente: I don't mean changing the default _instead_ of your changes.

I mean in addition to them. I don't think it makes sense to have the default logging info+debug instead of info+warn+error+...

guisehn: The default behavior of Logger is `[:lt, :eq]`, not `[:gt, :eq]`.

Example:

```elixir
iex(1)> require Logger
Logger
iex(2)> Logger.configure(level: :debug) # should print all level of logs
:ok
iex(3)> Logger.info("I am being logged because debug < info")

[info]  I am being logged because debug < info
:ok
iex(4)> Logger.configure(level: :warn) 
:ok
iex(5)> Logger.info("I am not being logged because warn > info")
:ok
```

If we do `[:gt, :eq]` and I have my application log level set to `debug`, elixir-grpc logs (that use `:info` by default) won't be printed, not matching Logger's behavior. That's the issue described on https://github.com/elixir-grpc/grpc/issues/325

guisehn: > I might have mixed up the comparators, but the point that we should change the default so that by passing info you log everything but debug still stands.

I think I'm missing something here.

Here's my thought process: The interceptor `level` configuration determines what log level the interceptor itself will use to log the requests coming in and out. If you set it to `:warn`, `GRPC.Client.Interceptors.Logger` will call `Log.warn(...)`. If you set it to `:debug`, `GRPC.Client.Interceptors.Logger` will call `Log.debug(...)`.

Now, it's up to the `:logger` configuration to determine what level should be printed or not. If you configure `:logger` level to `:info` but the log level of `GRPC.Client.Interceptors.Logger` is `:debug`, then nothing should be logged (as long as default is `[:lt, :eq]`).

If we keep `[:gt, :eq]` as the default using the scenario above (logger level = `:info`, interceptor level = `:debug`), the interceptor will enter the `if Logger.compare_levels` condition and will attempt to log, using resources unnecessarily to calculate the time diffs, while Elixir will discard the log.

I'm missing the use case of `accepted_comparators` because I couldn't understand in which use case it should be customized to anything besides `[:lt, :eq]`. If you don't bother to explain, can you share a real use case of that? 😅

polvalente: > I'm missing the use case of accepted_comparators because I couldn't understand in which use case it should be customized to anything besides [:lt, :eq]. If you don't bother to explain, can you share a real use case of that? 😅

That wasn't something I decided on haha! I just refactored a bit on this code that was already present.

polvalente: tl; dr: I'll make the changes I suggested, and as I see it, we're in overall agreement on how the code should work :)

polvalente: registering what we talked on Slack:

Turns out that the `accepted_comparators` option is too confusing and we don't see an actual application in which the option makes sense. Therefore, we're gonna remove the option and revert to the previous behavior.

polvalente (approved):

PR #334, opened 2023-11-08
Add result to client telemetry spans

davydog187:
This PR unifies client spans with server spans so that the result is present.


lib/grpc/telemetry.ex
```diff
@@ -115,7 +115,8 @@ defmodule GRPC.Telemetry do
 
     :telemetry.span(@client_rpc, start_metadata, fn ->
       try do
-        {span_fn.(), start_metadata}
+        result = span_fn.()
+        {result, Map.put(start_metadata, :result, result)}
```
Nezteb, line 119: Could we add this new `:result` metadata key to the moduledoc in the `Client-side Events -> Metadata` section? Similar to how it is `Server-side Events -> Metadata` section. 😄

davydog187: Yep good call @Nezteb

Nezteb (approved):

PR #346, opened 2024-01-31
Expose Server Address via gRPC Stream

matthewberends:
- Adds exposure to the gRPC server address via the gRPC stream.

We are currently in the process of updating services to follow opentelemetry semantics and a required attribute is [server.address](https://opentelemetry.io/docs/specs/semconv/rpc/rpc-spans/#common-attributes). With the current exposed APIs we could not find a way to access the server address details or the underlying request object to get this information.

### DIFF
--- a/lib/grpc/server/adapters/cowboy.ex
+++ b/lib/grpc/server/adapters/cowboy.ex
@@ -174,6 +174,10 @@ defmodule GRPC.Server.Adapters.Cowboy do
     Handler.get_headers(pid)
   end
 
+  def get_sock(%{pid: pid}) do
+    Handler.get_sock(pid)
+  end
+
   def get_peer(%{pid: pid}) do
     Handler.get_peer(pid)
   end
--- a/lib/grpc/server/adapters/cowboy/handler.ex
+++ b/lib/grpc/server/adapters/cowboy/handler.ex
@@ -142,6 +142,10 @@ defmodule GRPC.Server.Adapters.Cowboy.Handler do
     sync_call(pid, :get_headers)
   end
 
+  def get_sock(pid) do
+    sync_call(pid, :get_sock)
+  end
+
   def get_peer(pid) do
     sync_call(pid, :get_peer)
   end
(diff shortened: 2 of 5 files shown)


polvalente (changes requested):
Thanks for the PR!

Did you try to use `set_headers` to return the necessary metadata through that?
I think that's a better path for OTel because then you can also pass an OTel trace that continues through from the client to the server and back.

If you did, what were the problems you found?

Finally, if we do move forward with this PR (i.e. questions above have been fully answered), I'd like to suggest a few changes so that we don't leak Cowboy terminology into the server-agnostic layer

matthewberends:
> Did you try to use `set_headers` to return the necessary metadata through that?
> I think that's a better path for OTel because then you can also pass an OTel trace that continues through from the client to the server and back.

Hey @polvalente 

I looked briefly at using headers to propagate the server details but ultimately was hesitant since it seemed like the info should be available from the request on the server side and so adding it headers may be a bit hacky. One other concern we was if the client side did not use our instrumentation library it would not have the headers injected and races would potentially be missing information (but this situation is not common and probably acceptable).

This being said using headers would definitely work if thats whats recommended as best practice here?

polvalente:
> > Did you try to use `set_headers` to return the necessary metadata through that?
> > I think that's a better path for OTel because then you can also pass an OTel trace that continues through from the client to the server and back.
> 
> Hey @polvalente 
> 
> I looked briefly at using headers to propagate the server details but ultimately was hesitant since it seemed like the info should be available from the request on the server side and so adding it headers may be a bit hacky. One other concern we was if the client side did not use our instrumentation library it would not have the headers injected and races would potentially be missing information (but this situation is not common and probably acceptable).
> 
> This being said using headers would definitely work if thats whats recommended as best practice here?

As I see it, if you need the client side to be agnostic to how the server is transmitting the OTel data, then you should be including this as metadata in your proto definitions.

Generally, OTel will transmit trace info using headers anyway, so you'd be looking at those to complete end-to-end traces anyway.

Issue #312, opened 2023-04-13
server_stream keeps disconnecting


nulian:
Found that connecting with a server with a server stream it keeps disconnecting.

We where trying to connect with the Watch request in spicedb but usually we get auto disconnected.

Also tried the mint adapter it has the same issue but found it on server_stream the open check in connection_process.ex is checked as `Mint.HTTP.open?(state.conn, :read)` it works correctly.

Not sure how to fix it in the gun adapter.

polvalente:
Also, if there's already a known fix for the Mint client, PR is welcome!

nulian:
The issue came forward when we had response streaming only

```
rpc :Watch,
      AuthorizationGrpc.Authorization.WatchRequest,
      stream(AuthorizationGrpc.Authorization.WatchResponse)
```

sleipnir:
Closed due to inactivity. Please reopen if this is still necessary.

PR #352, opened 2024-02-23
Make server stream work with mint client and add keepalive.

nulian:
This is what I been using for a while in our application.

Changed connection open? check to :read so server streaming is seen as a valid state for the connection.

And added a keepalive to make sure the connection is still open of the server stream because if a service like nginx breaks the connection you won't get any notification so you will never close the connection even if it's already broken.

Also added some error handling cases so it won't cause any server errors.

#312

### DIFF
--- a/lib/grpc/client/adapters/mint.ex
+++ b/lib/grpc/client/adapters/mint.ex
@@ -114,12 +114,16 @@ defmodule GRPC.Client.Adapters.Mint do
       |> Keyword.get(:transport_opts, [])
       |> Keyword.merge(ssl)
 
-    [transport_opts: Keyword.merge(@default_transport_opts, transport_opts)]
+    keep_alive = Keyword.get(opts, :keep_alive, false)
+
+    [transport_opts: Keyword.merge(@default_transport_opts, transport_opts), keep_alive: keep_alive]
   end
 
   defp connect_opts(_channel, opts) do
     transport_opts = Keyword.get(opts, :transport_opts, [])
-    [transport_opts: Keyword.merge(@default_transport_opts, transport_opts)]
+    keep_alive = Keyword.get(opts, :keep_alive, false)
+
+    [transport_opts: Keyword.merge(@default_transport_opts, transport_opts), keep_alive: keep_alive]
   end
 
   defp mint_scheme(%Channel{scheme: "https"} = _channel), do: :https
--- a/lib/grpc/client/adapters/mint/connection_process/connection_process.ex
+++ b/lib/grpc/client/adapters/mint/connection_process/connection_process.ex
@@ -11,6 +11,8 @@ defmodule GRPC.Client.Adapters.Mint.ConnectionProcess do
   alias GRPC.Client.Adapters.Mint.ConnectionProcess.State
   alias GRPC.Client.Adapters.Mint.StreamResponseProcess
 
+  @keep_alive_timeout 60_000
+
   require Logger
 
   @connection_closed_error "the connection is closed"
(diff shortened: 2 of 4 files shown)


lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -94,6 +106,9 @@ defmodule GRPC.Client.Adapters.Mint.ConnectionProcess do
   def handle_call(:disconnect, _from, state) do
     # TODO add a code to if disconnect is brutal we just  stop if is friendly we wait for pending requests
     {:ok, conn} = Mint.HTTP.close(state.conn)
+
+    finish_all_pending_requests(state)
```
beligante, line 110: Nice catch

lib/grpc/client/adapters/mint.ex
```diff
@@ -114,12 +114,16 @@ defmodule GRPC.Client.Adapters.Mint do
       |> Keyword.get(:transport_opts, [])
       |> Keyword.merge(ssl)
 
-    [transport_opts: Keyword.merge(@default_transport_opts, transport_opts)]
+    keep_alive = Keyword.get(opts, :keep_alive, false)
+
+    [transport_opts: Keyword.merge(@default_transport_opts, transport_opts), keep_alive: keep_alive]
```
beligante, line 119: For clarity, it's better if you add a question mark to this option, since, usually `keep_alive` is a provided as a number by convention. 


```suggestion
    keep_alive? = Keyword.get(opts, :keep_alive?, false)

    [transport_opts: Keyword.merge(@default_transport_opts, transport_opts), keep_alive?: keep_alive?]
```

beligante: I wonder if it's valuable to add a module doc describing this adapter-specific option. WDYT?

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -172,6 +187,34 @@ defmodule GRPC.Client.Adapters.Mint.ConnectionProcess do
...
+          state
+          |> State.update_conn(conn)
+          |> State.set_ping_ref(ref)
+
+        {:noreply, state}
+
+      {:error, conn, _error} ->
+        state = State.update_conn(state, conn)
+        check_connection_status(state)
+        {:noreply, state}
```
beligante, line 205: You want the `check_connection_status` to handle the GS reply for you
```suggestion
        check_connection_status(state)
```

lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -261,6 +308,14 @@ defmodule GRPC.Client.Adapters.Mint.ConnectionProcess do
     new_state
   end
 
+  defp process_response({:error, request_ref, error}, state) do
+    pid = State.stream_response_pid(state, request_ref)
+    :ok = StreamResponseProcess.consume(pid, :error, error)
+    :ok = StreamResponseProcess.done(pid)
```
beligante, line 314: Why you're sending `done/1` here? By default, even if a error happens, the connection will produce a `:done` message when the stream is closed by the server

beligante (commented):
Overall code LGTM, great job!
The connection process has a good test coverage, could you add some tests for this ?

polvalente:
@nulian did you get a chance to act on the reviews?

lib/grpc/client/adapters/mint.ex
gritsenkob: https://github.com/elixir-mint/mint/blob/main/lib/mint/http.ex#L276-L364
can't see any keep_alive option here

beligante: `transport_opts` is forwarded to `gen_tcp` or `ssl`. The Mint docs list the most common ones, but there are more of them.

https://www.erlang.org/doc/man/gen_tcp.html#type-option

Now that you mentioned, `keepalive` on gen_tcp doesn't have a underscore

PR #357, opened 2024-04-04
Accept URIs with scheme, add default SSL options for https, deprecate URLs without scheme

v0idpwn:
This commit changes how addresses are handled by GRPC.Stub.connect/2.

First, by accepting URIs with scheme. Before, `http://localhost:50051" would
cause a crash.

Second, by using the URI scheme to automatically set a GRPC credential if no
GRPC credentials were provided. If we were to remove this part, we could at
least validate that a GRPC credential was provided when the scheme is https.

~~Finally, by deprecating receiving URLs without scheme, such as `localhost:50051`.
This isn't strictly necessary, but in my opinion makes things more consistent when
we want to support both local sockets and remote urls.~~
Edit: the deprecation part was removed after review.

### DIFF
--- a/lib/grpc/stub.ex
+++ b/lib/grpc/stub.ex
@@ -108,10 +108,10 @@ defmodule GRPC.Stub do
 
   ## Examples
 
-      iex> GRPC.Stub.connect("http://localhost:50051")
+      iex> GRPC.Stub.connect("localhost:50051")
       {:ok, channel}
 
-      iex> GRPC.Stub.connect("http://localhost:50051", accepted_compressors: [GRPC.Compressor.Gzip])
+      iex> GRPC.Stub.connect("localhost:50051", accepted_compressors: [GRPC.Compressor.Gzip])
       {:ok, channel}
 
       iex> GRPC.Stub.connect("/paht/to/unix.sock")
--- a/lib/grpc/stub.ex
+++ b/lib/grpc/stub.ex
@@ -131,13 +131,47 @@ defmodule GRPC.Stub do
   """
   @spec connect(String.t(), keyword()) :: {:ok, Channel.t()} | {:error, any()}
   def connect(addr, opts \\ []) when is_binary(addr) and is_list(opts) do
-    {host, port} =
-      case String.split(addr, ":") do
-        [host, port] -> {host, port}
-        [socket_path] -> {{:local, socket_path}, 0}
-      end
+    case URI.parse(addr) do
+      %URI{scheme: @secure_scheme, host: host, port: port}
+      when is_binary(host) and is_integer(port) ->
+        opts = Keyword.put_new_lazy(opts, :cred, &default_ssl_option/0)
+        connect(host, port, opts)
 
-    connect(host, port, opts)
+      %URI{scheme: @insecure_scheme, host: host, port: port}
+      when is_binary(host) and is_integer(port) ->
+        connect(host, port, opts)
+
+      # For compatibility with previous versions, we accept URIs in the "#{address}:#{port}" format
+      _ ->
+        case String.split(addr, ":") do
+          [socket_path] ->
+            connect({:local, socket_path}, 0, opts)
+
+          [address, port] ->
+            Logger.warning("Usage of non-local URIs without scheme is deprecated")
+            port = String.to_integer(port)
+            connect(address, port, opts)
+        end
+    end
+  end
+
+  defp default_ssl_option do
+    if Code.ensure_loaded?(CAStore) do
+      %GRPC.Credential{
+        ssl: [
+          verify: :verify_peer,
+          depth: 99,
+          cacert_file: CAStore.file_path()
+        ]
+      }
+    else
+      raise """
+      no GRPC credentials provided. Please either:
+
+      - Pass the `:cred` option to `GRPC.Stub.connect/3`
+      - Add `:castore` to your dependencies
+      """
+    end
   end
 
   @spec connect(String.t(), binary() | non_neg_integer(), keyword()) ::
--- a/lib/grpc/stub.ex
+++ b/lib/grpc/stub.ex
@@ -155,7 +155,6 @@ defmodule GRPC.Stub do
             connect({:local, socket_path}, 0, opts)
 
           [address, port] ->
-            Logger.warning("Usage of non-local URIs without scheme is deprecated")
             port = String.to_integer(port)
             connect(address, port, opts)
         end
--- a/lib/grpc/stub.ex
+++ b/lib/grpc/stub.ex
@@ -132,16 +132,15 @@ defmodule GRPC.Stub do
   @spec connect(String.t(), keyword()) :: {:ok, Channel.t()} | {:error, any()}
   def connect(addr, opts \\ []) when is_binary(addr) and is_list(opts) do
     case URI.parse(addr) do
-      %URI{scheme: @secure_scheme, host: host, port: port}
-      when is_binary(host) and is_integer(port) ->
+      %URI{scheme: @secure_scheme, host: host, port: port} ->
         opts = Keyword.put_new_lazy(opts, :cred, &default_ssl_option/0)
         connect(host, port, opts)
 
-      %URI{scheme: @insecure_scheme, host: host, port: port}
-      when is_binary(host) and is_integer(port) ->
+      %URI{scheme: @insecure_scheme, host: host, port: port} ->
         connect(host, port, opts)
 
-      # For compatibility with previous versions, we accept URIs in the "#{address}:#{port}" format
+      # For compatibility with previous versions, we accept URIs in
+      # the "#{address}:#{port}" format
       _ ->
         case String.split(addr, ":") do
           [socket_path] ->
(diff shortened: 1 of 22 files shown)


lib/grpc/stub.ex
```diff
@@ -131,13 +131,46 @@ defmodule GRPC.Stub do
   """
   @spec connect(String.t(), keyword()) :: {:ok, Channel.t()} | {:error, any()}
   def connect(addr, opts \\ []) when is_binary(addr) and is_list(opts) do
-    {host, port} =
-      case String.split(addr, ":") do
-        [host, port] -> {host, port}
-        [socket_path] -> {{:local, socket_path}, 0}
-      end
+    case URI.parse(addr) do
```
polvalente, line 134:
```suggestion
    # This works because we only accept `http` and `https` schemes (allowlisted below explicitly)
    # addresses like "localhost:1234" parse as if `localhost` is the scheme for URI, and this falls through to
    # the base case. Accepting only `http/https` is a trait of `connect/3`.
    
    case URI.parse(addr) do
```

lib/grpc/stub.ex
```diff
@@ -131,13 +131,50 @@ defmodule GRPC.Stub do
...
+        ]
+      }
+    else
+      raise """
+      no GRPC credentials provided. Please either:
+
+      - Pass the `:cred` option to `GRPC.Stub.connect/2,3`
+      - Add `:castore` to your list of dependencies in `mix.exs`
+      """
+    end
```
polvalente, line 177: I'm not sure this is the best way to check this. Could we instead do the check at compile-time?

```suggestion
  if {:module, CAStore} == Code.ensure_loaded!(CAStore) do
    defp default_ssl_option do
      %GRPC.Credential{
        ssl: [
          verify: :verify_peer,
          depth: 99,
          cacert_file: CAStore.file_path()
        ]
      }
    end
  else
    defp default_ssl_option do
      raise """
      no GRPC credentials provided. Please either:

      - Pass the `:cred` option to `GRPC.Stub.connect/2,3`
      - Add `:castore` to your list of dependencies in `mix.exs`
      """
    end
  end
```

polvalente: Or even just bring castore as a full on dep

test/grpc/channel_test.exs
```diff
@@ -3,20 +3,51 @@ defmodule GRPC.ChannelTest do
...
+    test "works for insecure" do
+      {:ok, channel} = GRPC.Stub.connect("10.1.0.0:50051", adapter: ClientAdapter)
+      assert %Channel{host: "10.1.0.0", port: 50051, scheme: "http", cred: nil} = channel
+    end
+
+    test "works for ssl" do
+      cred = %{ssl: []}
+      {:ok, channel} = GRPC.Stub.connect("10.1.0.0:50051", adapter: ClientAdapter, cred: cred)
+      assert %Channel{host: "10.1.0.0", port: 50051, scheme: "https", cred: ^cred} = channel
+    end
```
polvalente, line 42:
```suggestion
    for {host, host_kind} <- [{"10.1.0.0", "ip"}, {"hostname", "hostname"}] do
      test "works for insecure - #{host_kind}" do
        {:ok, channel} = GRPC.Stub.connect("#{unquote(host}}:50051", adapter: ClientAdapter)
        assert %Channel{host: unquote(host}, port: 50051, scheme: "http", cred: nil} = channel
      end

      test "works for ssl - #{host_kind}" do
        cred = %{ssl: []}
        {:ok, channel} = GRPC.Stub.connect("#{unquote(host}}:50051", adapter: ClientAdapter, cred: cred)
        assert %Channel{host: unquote(host}, port: 50051, scheme: "https", cred: ^cred} = channel
      end
    end
```

polvalente: So that we test both ip and hostname here

lib/grpc/stub.ex
```diff
@@ -131,13 +131,50 @@ defmodule GRPC.Stub do
...
+    # This works because we only accept `http` and `https` schemes (allowlisted below explicitly)
+    # addresses like "localhost:1234" parse as if `localhost` is the scheme for URI, and this falls through to
+    # the base case. Accepting only `http/https` is a trait of `connect/3`.
+    
+    case URI.parse(addr) do
+      %URI{scheme: @secure_scheme, host: host, port: port} ->
+        opts = Keyword.put_new_lazy(opts, :cred, &default_ssl_option/0)
+        connect(host, port, opts)
+
+      %URI{scheme: @insecure_scheme, host: host, port: port} ->
```
polvalente, line 143: I think we need a test for when scheme is `http` and there are credentials in `opts`. I believe the current code will force the upgrade to https. We should raise an error in this case, because the user would be explicitly requesting http

v0idpwn: You're right it would automatically upgrade. Added an error

v0idpwn:
Pushed a few commits doing (I believe) all suggestions :)

lib/grpc/stub.ex
```diff
@@ -131,17 +131,62 @@ defmodule GRPC.Stub do
...
+    # the base case. Accepting only `http/https` is a trait of `connect/3`.
+
+    case URI.parse(addr) do
+      %URI{scheme: @secure_scheme, host: host, port: port} ->
+        opts = Keyword.put_new_lazy(opts, :cred, &default_ssl_option/0)
+        connect(host, port, opts)
+
+      %URI{scheme: @insecure_scheme, host: host, port: port} ->
+        if opts[:cred],
+          do: raise(ArgumentError, "invalid option for insecure (http) address: :cred")
```
polvalente, line 145:
```suggestion
        if opts[:cred] do
          raise ArgumentError, "invalid option for insecure (http) address: :cred"
        end
```

test/grpc/integration/stub_test.exs
```diff
@@ -78,7 +80,7 @@ defmodule GRPC.Integration.StubTest do
 
   test "returns error when timeout" do
     run_server(SlowServer, fn port ->
-      {:ok, channel} = GRPC.Stub.connect("localhost:#{port}")
+      {:ok, channel} = GRPC.Stub.connect("http://localhost:#{port}")
```
polvalente, line 83: Why did we have to change these to http?

polvalente (approved):
Only 1 question left :)

test/grpc/integration/stub_test.exs
v0idpwn: I was just updating it to avoid using the deprecated branch

polvalente: Not sure if we should really deprecate this. Any specific reason?

v0idpwn: I'm not 100% sure if it should be deprecated. I'm aiming for an API where the input address is always a valid URI. For example, this would be the code without backwards compatibility:

```elixir
case URI.parse(addr) do
   %URI{scheme: "http"} -> # insecure 
   %URI{scheme: "https"} -> # secure
   %URI{scheme: nil, path: path} -> # unix domain socket
   _ -> # error
end
```

v0idpwn: If you think deprecation isn't the way, I can revert ec02a949c173ca17f563e1504aa5078a700af94a and remove the warning log

polvalente: I see no need to deprecate since:

1. It's an unnecessary breaking change because your original issue was not being able to pass a scheme
2. The code to maintain this is quite small.

The third alternative where you only have the path but no port is already captured in the fallback clause.

polvalente: Let's revert for now

polvalente, line 83:
```suggestion
      {:ok, channel} = GRPC.Stub.connect("localhost:#{port}")
```

test/grpc/integration/stub_test.exs
```diff
@@ -59,7 +59,9 @@ defmodule GRPC.Integration.StubTest do
   test "body larger than 2^14 works" do
     run_server(HelloServer, fn port ->
       {:ok, channel} =
-        GRPC.Stub.connect("localhost:#{port}", interceptors: [GRPC.Client.Interceptors.Logger])
+        GRPC.Stub.connect("http://localhost:#{port}",
+          interceptors: [GRPC.Client.Interceptors.Logger]
+        )
```
polvalente, line 64:
```suggestion
        GRPC.Stub.connect("localhost:#{port}", interceptors: [GRPC.Client.Interceptors.Logger])
```

test/grpc/integration/stub_test.exs
```diff
@@ -32,7 +32,7 @@ defmodule GRPC.Integration.StubTest do
 
   test "you can disconnect stubs" do
     run_server(HelloServer, fn port ->
-      {:ok, channel} = GRPC.Stub.connect("localhost:#{port}")
+      {:ok, channel} = GRPC.Stub.connect("http://localhost:#{port}")
```
polvalente, line 35:
```suggestion
      {:ok, channel} = GRPC.Stub.connect("localhost:#{port}")
```

test/grpc/integration/stub_test.exs
```diff
@@ -50,7 +50,7 @@ defmodule GRPC.Integration.StubTest do
 
   test "disconnecting a disconnected channel is a no-op" do
     run_server(HelloServer, fn port ->
-      {:ok, channel} = GRPC.Stub.connect("localhost:#{port}")
+      {:ok, channel} = GRPC.Stub.connect("http://localhost:#{port}")
```
polvalente, line 53:
```suggestion
      {:ok, channel} = GRPC.Stub.connect("localhost:#{port}")
```

polvalente (approved):

PR #362, opened 2024-05-03
Translate errors from HTTP statuses

v0idpwn:
This is inspired in how GRPC go approaches HTTP errors when performing GRPC requests.

This provides meaningful errors in different situations. For example, AWS load balancer replies with a 504 when there are no live instances. There, a `unavailable` error is much more fitting than `internal`.

For go implementation: https://github.com/grpc/grpc-go/blob/273fe145d03df516018cf4642e6987e027ffc0f5/internal/transport/http_util.go#L68

### DIFF
--- a/lib/grpc/rpc_error.ex
+++ b/lib/grpc/rpc_error.ex
@@ -71,6 +71,23 @@ defmodule GRPC.RPCError do
     %{error | message: error.message || Status.status_message(error.status)}
   end
 
+  def from_status(401), do: exception(GRPC.Status.unauthorized(), status_message(401))
+  def from_status(403), do: exception(GRPC.Status.permission_denied(), status_message(403))
+  def from_status(404), do: exception(GRPC.Status.unimplemented(), status_message(404))
+  def from_status(429), do: exception(GRPC.Status.unavailable(), status_message(429))
+
+  def from_status(status) when status in [502, 503, 504] do
+    exception(GRPC.Status.unavailable(), status_message(status))
+  end
+
+  def from_status(status) do
+    exception(GRPC.Status.internal(), status_message(status))
+  end
+
+  defp status_message(status) do
+    "got http status #{status_code} instead of 200"
+  end
+
   defp parse_args([], acc), do: acc
 
   defp parse_args([{:status, status} | t], acc) when is_integer(status) do
--- a/lib/grpc/client/adapters/gun.ex
+++ b/lib/grpc/client/adapters/gun.ex
@@ -245,11 +245,7 @@ defmodule GRPC.Client.Adapters.Gun do
                )}
           end
         else
-          {:error,
-           GRPC.RPCError.exception(
-             GRPC.Status.internal(),
-             "status got is #{status} instead of 200"
-           )}
+          {:error, GRPC.RPCError.from_status(status)}
         end
 
       {:response, :nofin, status, headers} ->
@@ -266,11 +262,7 @@ defmodule GRPC.Client.Adapters.Gun do
             {:response, headers, :nofin}
           end
         else
-          {:error,
-           GRPC.RPCError.exception(
-             GRPC.Status.internal(),
-             "status got is #{status} instead of 200"
-           )}
+          {:error, GRPC.RPCError.from_status(status)}
         end
 
       {:data, :fin, data} ->


v0idpwn:
I didn't find where these errors are handled in Mint (if they are), so I didn't touch it. I'm not using Mint either.

lib/grpc/rpc_error.ex
```diff
@@ -71,7 +71,7 @@ defmodule GRPC.RPCError do
     %{error | message: error.message || Status.status_message(error.status)}
   end
 
-  def from_status(401), do: exception(GRPC.Status.unauthorized(), status_message(401))
+  def from_status(401), do: exception(GRPC.Status.unauthenticated(), status_message(401))
```
polvalente, line 74: 401 is unauthorized as per HTTP specs

polvalente: https://developer.mozilla.org/pt-BR/docs/Web/HTTP/Status/401

polvalente: I believe we should use the same verbs as HTTP if we are to do any kind of translation here

polvalente (changes requested):
Verbs should be changed and tests need to be added

lib/grpc/rpc_error.ex
v0idpwn: Hi, @polvalente. There's only so much we can do without adding new errors to the library. As the GRPC specification states only 16 error codes, what we do is only a best efforth. I copied all the matching pairs from the official implementation in Go

polvalente: Right! In this case, I believe a comment should be added noting this :)

v0idpwn: WRT to the mint implementation, I can add, but I wanted to get some validation on this first :)

polvalente: Nice :)
Go ahead with Mint and tests then! Looks good to me now, with the clarification above

v0idpwn: Hi, @polvalente. I managed to write tests, but found that adding the status handling to the mint adapter is way more complex due to the way it's structured. Do they really need to be feature-equivalent? Can't we perhaps open an issue for a mint implementation of this feature in the future?

test/grpc/integration/stub_test.exs
```diff
@@ -30,6 +30,30 @@ defmodule GRPC.Integration.StubTest do
...
+    {503, GRPC.Status.unavailable()},
+    {504, GRPC.Status.unavailable()},
+    # General case
+    {518, GRPC.Status.internal()}
+  ]
+
+  client_adapters = [GRPC.Client.Adapters.Gun, GRPC.Client.Adapters.Mint]
+
+  for {http_code, expected_error_code} <- error_table, client_adapter <- client_adapters  do
+    test "#{client_adapter} returns RPC Error when getting HTTP #{http_code}" do
```
v0idpwn, line 48: Not sure where I should put these tests

lib/grpc/rpc_error.ex
polvalente: I'd rather have them being feature-equivalent as one is meant to be a drop-in replacement of the other.

Can you elaborate on what made it difficult to add this kind of handling?

v0idpwn: Essentially, the status is received at `ConnectionProcess`, which is called to start the request, but the calling process reads the response from the `StreamResponseProcess`. Meaning I'd need to write some process coordination or rework current coordination logic. Since I don't fully understand the current architecture decisions (yet?), I found this a bit challenging.

polvalente: I believe you need to change things here https://github.com/elixir-grpc/grpc/blob/b3a7ff627ea013907a78c693016325a023a540b8/lib/grpc/client/adapters/mint.ex#L162C3-L164C6

polvalente: And maybe here: https://github.com/elixir-grpc/grpc/blob/b3a7ff627ea013907a78c693016325a023a540b8/lib/grpc/client/adapters/mint.ex#L162C3-L164C6

test/grpc/integration/stub_test.exs
polvalente: This seems fine

sleipnir:
@v0idpwn any progress here?

v0idpwn:
Hi, @sleipnir. I did an attempt at implementing it on the mint adapter, but ended up hitting a wall. I'll give it another stab soonish.

sleipnir:
ping @v0idpwn

polvalente:
Closing due to inactivity.

PR #367, opened 2024-06-03
Remove deprecated functions

sleipnir:
This PR just removes some deprecated functions


davydog187:
I'm ok with this change as long as we're planning on doing a minor release afterwards

sleipnir:
> I'm ok with this change as long as we're planning on doing a minor release afterwards

@polvalente wdyt?

PR #368, opened 2024-06-09
Fix elixir 1.17 warnings

hypno2000:
Fix these warnings in elixir 1.17:

```
warning: using map.field notation (without parentheses) to invoke function Module.__rpc_calls__() is deprecated, you must add parentheses instead: remote.function()
```

Right now all projects that depend on `grpc` generate warnings on user project (because the warnings are generated in the macros that user projects are calling), which means it breaks user projects CI's in elixir 1.17 that use `--warnings-as-errors`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,8 +33,8 @@ jobs:
     name: OTP ${{matrix.otp}} / Elixir ${{matrix.elixir}}
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


hypno2000:
Is there anything i can do from my side to help it getting merged?

sleipnir (approved):

sleipnir:
Thank you @hypno2000 sorry for the late answer.

PR #370, opened 2024-06-24
Allow configurable Mint adapter

caike:
## Problem

The use case for this came from attempting to connect to a particular gRPC server which was rejecting the connection due to the default `transport_options` being used:

```
[
    timeout: :infinity,
    verify: :verify_peer,
    depth: 99,
    cacert_file: "...." # redacted
  ]
```

## Solution

After overriding `transport_options` I was able to connect to said server. I believe the ability to override these options could also be useful for others. This PR allows the adapter options to be set from the client application via:

```
config :grpc, GRPC.Client.Adapters.Mint, transport_opts: []
```

### DIFF
--- a/lib/grpc/client/adapters/mint.ex
+++ b/lib/grpc/client/adapters/mint.ex
@@ -15,7 +15,8 @@ defmodule GRPC.Client.Adapters.Mint do
 
   @impl true
   def connect(%{host: host, port: port} = channel, opts \\ []) do
-    opts = Keyword.merge(@default_connect_opts, connect_opts(channel, opts))
+    opts = connect_opts(channel, opts) |> merge_opts()
+
     Process.flag(:trap_exit, true)
 
     channel
@@ -120,6 +121,14 @@ defmodule GRPC.Client.Adapters.Mint do
   defp connect_opts(_channel, opts) do
     transport_opts = Keyword.get(opts, :transport_opts, [])
     [transport_opts: Keyword.merge(@default_transport_opts, transport_opts)]
+  end
+
+  # Merges default opts with application specific opts
+  # set via `config :grpc, GRPC.Client.Adapters.Mint, custom_opts`
+  defp merge_opts(opts) do
+    module_opts = Application.get_env(:grpc, __MODULE__, [])
+    opts = Keyword.merge(opts, module_opts)
+    Keyword.merge(@default_connect_opts, opts)
   end
 
   defp mint_scheme(%Channel{scheme: "https"} = _channel), do: :https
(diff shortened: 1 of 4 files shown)


sleipnir:
Hi @caike! First of all, thank you for taking the time for us.
If I'm not wrong we deprecated options via configuration in the past in favor of passing options as options directly at GRPC.Server.Supervisor initialization.

Am I right on this @polvalente ?

caike:
> Hi @caike! First of all, thank you for taking the time for us. If I'm not wrong we deprecated options via configuration in the past in favor of passing configurations as options directly at `GRPC.Server.Supervisor` initialization.
> 
> Am I right on this @polvalente ?

Thanks for the feedback and for maintaining this project!

How about when the application is simply a client to an external grpc server ? I have a client Stub that connects to an external server and I'm not using `GRPC.Server.Supervisor` at all. For more context, [this my current use case](https://github.com/wowica/elixir-sdk/blob/main/lib/chain_sync.ex#L29-L48)

sleipnir:
> 
> Thanks for the feedback and for maintaining this project!
> 

Thank you very much!

> How about when the application is simply a client to an external grpc server ? I have a client Stub that connects to an external server and I'm not using `GRPC.Server.Supervisor` at all. For more context, [this my current use case](https://github.com/wowica/elixir-sdk/blob/main/lib/chain_sync.ex#L29-L48)

Sorry for my inattention, for customers you are correct in going through the application configuration. Although for me it might be interesting to add this as options for the connect function of the Stub module. If we wanted to have both options available (via options and via config)
If you want to continue working on this, we'll wait until it's ready. However, if you want to continue with this PR this way, that's fine with me. Be as you prefer.

lib/grpc/client/adapters/mint.ex
```diff
@@ -15,7 +15,14 @@ defmodule GRPC.Client.Adapters.Mint do
 
   @impl true
   def connect(%{host: host, port: port} = channel, opts \\ []) do
-    opts = Keyword.merge(@default_connect_opts, connect_opts(channel, opts))
+    # Added :config_options to facilitate testing.
+    {config_opts, opts} = Keyword.pop(opts, :config_options, [])
+    # Merges default opts with application specific opts set
```
sleipnir, line 20: I think this comment is not necessary. The code is self explanatory

sleipnir (commented):
Just a small comment

lib/grpc/client/adapters/mint.ex
sleipnir: This comment about how pass the options to the adapter would make more sense to be in the documentation via README or something like that

caike: Good call. Added suggested changes.

sleipnir (approved):

sleipnir:
thank you for this work @caike

Issue #184, opened 2021-01-20
The GRPC.Adapter.Cowboy.Handler does not handle :EXIT message


bszaf:
**Describe the bug**
The `GRPC.Adapter.Cowboy.Handler` does not handle `:shutdown` message from Cowboy.

**To Reproduce**
 - Open connection to GRPC Server
 - Open a stream in the context of connection. Possibly long lived
 - Close connection while stream is open: ` GRPC.Stub.disconnect(channel)`

**Expected behavior**
1. Let stream handling process know that connection is closed, so it can terminate gracefully.
2. Optionally log an error/warn. Ideally it could be silenced. 

**Logs**

```
time=2021-01-20T06:21:01.504 node=backend@10.68.143.164 pid=<0.5844.0> [error] Ranch listener "GRPC.Endpoint", connection process #PID<0.5844.0>, stream 3 had its request process #PID<0.6180.0> exit with reason :function_clause and stacktrace [{GRPC.Adapter.Cowboy.Handler, :info, [{:EXIT, #PID<0.6182.0>, :shutdown}, %{bindings: %{}, body_length: :undefined, cert: <<stripped>>, has_body: true, headers: %{"content-type" => "application/grpc+proto", "te" => "trailers", "user-agent" => "grpc-elixir/0.5.0-beta"}, host: "stripped", host_info: :undefined, method: "POST", path: "/stripped", path_info: :undefined, peer: {{127, 0, 0, 1}, 50090}, pid: #PID<0.5844.0>, port: 443, qs: "", ref: "GRPC.Endpoint", resp_headers: %{"content-type" => "application/grpc+proto"}, scheme: "https", sock: {{127, 0, 0, 1}, 50056}, streamid: 3, version: :"HTTP/2"}, %{handling_timer: nil, pid: #PID<0.6181.0>}], []}, {:cowboy_loop, :call, 5, [file: 'deps/cowboy/src/cowboy_loop.erl', line: 77]}, {:grpc_stream_h, :execute, 3, [file: 'src/grpc_stream_h.erl', line: 348]}, {:grpc_stream_h, :request_process, 3, [file: 'src/grpc_stream_h.erl', line: 330]}, {:proc_lib, :init_p_do_apply, 3, [file: 'proc_lib.erl', line: 249]}]
```

**Versions:**
- OS: 
- Erlang:
- Elixir:
- mix.lock(grpc, gun, cowboy, cowlib):

```
[...]

PR #375, opened 2024-07-01
[#184] gracefully handles server process shutdown

beligante:
### Fixes https://github.com/elixir-grpc/grpc/issues/184

Add pattern match to handle server process shutdown. It's very common to see these error happen when we manually setup gRPC server in our test cases. ExUnit gracefully shutdown the applications (for example), thus causing the server processes to exit with `:shutdown` reason.

Without this handler, a lot of error logs are produced, causing to misleading logs.


lib/grpc/server/adapters/cowboy/handler.ex
```diff
@@ -449,8 +449,8 @@ defmodule GRPC.Server.Adapters.Cowboy.Handler do
     end
   end
 
-  def info({:EXIT, pid, :normal}, req, state = %{pid: pid}) do
-    exit_handler(pid, :normal)
```
beligante, line 453: removed this call to `exit_handler` -- cause this function is redundant for this callback.

test/support/google/api/annotations.pb.ex
beligante, line 1: I removed these two files, cause they're already defined in https://github.com/elixir-grpc/grpc/tree/master/lib/google/api

beligante: This was causing the test to throw some warnings

sleipnir (approved):

test/grpc/integration/server_test.exs
```diff
@@ -7,6 +7,12 @@ defmodule GRPC.Integration.ServerTest do
     def get_feature(point, _stream) do
       %Routeguide.Feature{location: point, name: "#{point.latitude},#{point.longitude}"}
     end
+
+    def route_chat(_ex_stream, stream) do
+      GRPC.Server.send_headers(stream, %{})
```
zhihuizhang17, line 12: Hi, I'm investigating a Cowboy shutdown issue for my gRPC service running version v0.9.0 in production. My client receives `{:error, %GRPC.RPCError{status: 13, message: "status got is 204 instead of 200"}}` when Cowboy shuts down.

In my unit test, if I add this line, my client receives `{:error, %GRPC.RPCError{status: 13, message: ":stream_error: {:stream_error, :no_error, :\"Stream reset by server.\"}"}}`.

How can I gracefully handle the service's endpoint shutdown from the client side? Should I send headers like retry or goaway to prompt the client to retry?

PR #351, opened 2024-02-16
Feat/http transcoding (Continuation)

sleipnir:
This PR continues the excellent work started by @drowzy in his PR https://github.com/elixir-grpc/grpc/pull/273. The merit of this work must be given entirely to @drowzy.

I just made some small fixes and eliminated all the deprecated code from the Protobuf api, as well as merging it with the recent codebase and fixing some tests.

I hope @drowzy doesn't mind that I picked up where he left off. But I accept it if you don't want to continue with this PR here. Again I reiterate that he did an excellent job in his previous PR.

closes #273 

***Below is the text copied from the original PR (https://github.com/elixir-grpc/grpc/pull/273):***

Support for [gRPC transcoding](https://cloud.google.com/endpoints/docs/grpc/transcoding).

* Closes https://github.com/elixir-grpc/grpc/issues/192
* Closes https://github.com/elixir-grpc/grpc/issues/274

This feature is opt-in using the http_transcode option in the server definition:

```elixir
defmodule Helloworld.Greeter.Server do
  use GRPC.Server,
    service: Helloworld.Greeter.Service,
    http_transcode: true
```
Request/response mapping can be controlled by adding grpc-gateway annotations to your proto definition:

```protobuf
import "google/api/annotations.proto";
import "google/protobuf/timestamp.proto";

package helloworld;

// The greeting service definition.
service Greeter {
  // Sends a greeting
  rpc SayHello (HelloRequest) returns (HelloReply) {
    option (google.api.http) = {
      get: "/v1/greeter/{name}"
    };
  }

  rpc SayHelloFrom (HelloRequestFrom) returns (HelloReply) {
    option (google.api.http) = {
      post: "/v1/greeter"
      body: "*"
    };
  }
}
```
This PR does not require any changes to user code.

For a full list of features see https://github.com/elixir-grpc/grpc/pull/273

PR #376, opened 2024-07-07
Require protobuf only in dev/test

v0idpwn:
This fixes a potential regression on #351. 

On README it's stated:
```
    # We don't force protobuf as a dependency for more
    # flexibility on which protobuf library is used,
    # but you probably want to use it as well
```

At @coingaming we rely on this characteristic of `:grpc` to be able to use an internal fork of  `:protobuf`. Due to this change, we get a dependency conflict and aren't able to use `:grpc` v0.8+.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -40,7 +40,7 @@ defmodule GRPC.Mixfile do
       {:jason, ">= 0.0.0", optional: true},
       {:cowlib, "~> 2.12"},
       {:castore, "~> 0.1 or ~> 1.0", optional: true},
-      {:protobuf, "~> 0.11"},
+      {:protobuf, "~> 0.11", only: [:dev, :test]},
       {:protobuf_generate, "~> 0.1.1", only: [:dev, :test]},
       {:googleapis,
        github: "googleapis/googleapis",


mix.exs
```diff
@@ -40,7 +40,7 @@ defmodule GRPC.Mixfile do
       {:jason, ">= 0.0.0", optional: true},
       {:cowlib, "~> 2.12"},
       {:castore, "~> 0.1 or ~> 1.0", optional: true},
-      {:protobuf, "~> 0.11"},
+      {:protobuf, "~> 0.11", only: [:dev, :test]},
```
polvalente, line 43: I think this has already been changed before.

Can you check the history to confirm, please?

Simplest solution for you is to override: true on this dependency

v0idpwn: I think can't `override: true` because I'm a package too :(

Will double check the history

I'd understand if its not accepted (but then the readme would need to be updated)

v0idpwn: @polvalente #233

v0idpwn: @sleipnir was the change intentional? 👀

v0idpwn: Looks like it breaks without this dependency, because of some new files that were added that use `:protobuf` in comp time...

v0idpwn:
Closing as it doesn't seem possible anymore. Will follow up with a PR removing the statement from the readme.

mix.exs
sleipnir: @v0idpwn Yes for implementing http transcoding we had to add some things, I'm not sure if it's possible to do without this. I may look into this in the future and try to reopen this issue if I see a solution.

Issue #307, opened 2023-03-13
Performance issues with `Mint` adapter 


beligante:
**Describe the bug**
There are some issues with the current implementation for mint adapter from a performance standpoint. 

**To Reproduce**
Easy way to see is In [Interop tests script](https://github.com/elixir-grpc/grpc/blob/master/interop/script/run.exs)

If you wrap the runner with `:timer.tc` and run ` mix run script/run.exs --rounds 1` it's possible to see that the amount of time that Gun adapter takes to run all test cases vs Mint is quite different. On my machine the avg for both was:
- 800ms for Gun
- 5.6s for Mint

With the numbers above it's possible to assume for sure that there are some performance improvements that needs to be done for Mint adapter.

**Expected behavior**
Similar times for both adapters 

**Versions:**
- OS: Linux / Mint
- Elixir: 1.13.3-otp-25
- Erlang: 25.0.3

**Additional context**
A few things that I could observer from the current implementation that could be better:
- [Accessing/Updating state using `Accesss`](https://github.com/elixir-grpc/grpc/blob/master/lib/grpc/client/adapters/mint/connection_process/state.ex#L33) - According to this article this action is quite slow: https://medium.com/learn-elixir/speed-up-data-access-in-elixir-842617030514
- Storing and updating data on state that I'm not using. 

I'm open for investigate and fix this issue and also I'm quite open for suggestions.

polvalente:
I'd look for a simpler script that benchmarks with Benchee so that things are more easily reproducible.

It could be the case that the Mint code is as fast as Gun when there's no concurrency. If this is the case, then it's a problem with process message serialization.

If even with minimal concurrency it's still much slower, then it's a matter of finding the bottleneck in the code.

beligante:
Gotcha! I'll work on the script for Benchee (which - TIL about the lib) and see what are the results. Tks Paulo!

Issue #377, opened 2024-07-07
Add new option to server to report crashes/exceptions 


beligante:
**Is your feature request related to a problem? Please describe.**

Hello Team! 

Currently at my work, the team face a challenge when comes to the topic of reporting crashes to tools such as Sentry or Bugsnag. Internally we use a fork of this project where we add some implementation details to report crashes from inside `Cowboy.Handler` 

So, I thought on a proposal where we could introduce something generic and flexible enough so that anyone can use for their own purposes. 

My Idea here is to introduce an option to the GRPC Server startup where you can pass a callback to be invoked whenever an exception (or what the maintainers consider an exception worthy to be reported), that callback expects an exception to be provided and internally any implementation can be used.

**Describe the solution you'd like**

The new option would be passed on the server startup. Eg:

```elixir

# bugsnag (no sanitization)
Supervisor.init(endpoint: MyEnpoint, port: 1234, start_server: true, crash_report: &Bugsnag.report/1)

#sentry
Supervisor.init(endpoint: MyEnpoint, port: 1234, start_server: true, crash_report: &Sentry.capture_exception/1)

# own impl
Supervisor.init(endpoint: MyEnpoint, port: 1234, start_server: true, crash_report: &MyOwnReport.capture/1)
```

**Describe alternatives you've considered**

We can alternatively add these report calls on the service level implementations. Such as

```elixir
defmodule FeatureServer do
  use GRPC.Server, service: Routeguide.RouteGuide.Service

  def get_feature(point, _stream) do
    # something that can generate exceptions
  catch
    kind, e -> Bugsnag.report(e, __STACKTRACE__)  
  end
end
```

Tho, add the report on that level has it's limitations. For example

1. We can only capture exceptions related to the function used to resolve this request (which is usually what we're interested), but with this level, we don't have access to the request ( `:cowboy.req` ) to add another layer of information about that execution
2. We can't capture exceptions related to issues on the grpc connection level that can at some point terminate the connection (such as client canceling request or closing the connection abrupt abruptly)  

**Additional context**
I made a WIP PR to illustrate better the idea, there's some punishment required, but I think I can add more depth to my request. It's on my fork for fow. Good part of the work there I got inspired by @darrenclark 

https://github.com/beligante/grpc/pull/3

beligante:
Another interesting option was presented to me today.  [Logger Metadata](https://hexdocs.pm/logger/main/Logger.html#module-metadata) -- In specific the `crash_reason` metadata

[...]

PR #378, opened 2024-07-07
Allow client settings opt on mint adapter

beligante:
Hey Team, I've been actively working on this: https://github.com/elixir-grpc/grpc/issues/307 but so far no good results.

Tho, I've noticed that with a few tweaks in the configs, it's possible to improve the Mint adapter performance by 2x (which still slow if compared with gun, but it's a quick win)

This PR aims to accept a new option when starting the connection where it's possible to pass [Mint's client settings options](https://hexdocs.pm/mint/Mint.HTTP2.html#t:setting/0) and also setting a bigger window size as default (I got these values from [Gun](https://github.com/ninenines/gun/blob/master/src/gun_http2.erl#L187) )

I also have a branch where , with the supposer of benchee I can compare the improvement on speed 

https://github.com/beligante/grpc/pull/4

### DIFF
--- a/lib/grpc/client/adapters/mint.ex
+++ b/lib/grpc/client/adapters/mint.ex
@@ -1,6 +1,6 @@
 defmodule GRPC.Client.Adapters.Mint do
   @moduledoc """
-  A client adapter using Mint
+  A client adapter using Mint.
   """
 
   alias GRPC.Channel
--- a/lib/grpc/client/adapters/mint.ex
+++ b/lib/grpc/client/adapters/mint.ex
@@ -10,7 +10,13 @@ defmodule GRPC.Client.Adapters.Mint do
 
   @behaviour GRPC.Client.Adapter
 
-  @default_connect_opts [protocols: [:http2]]
+  @default_connect_opts [
+    protocols: [:http2]
+  ]
+  @default_client_settings [
+    initial_window_size: 8_000_000,
+    max_frame_size: 8_000_000
+  ]
   @default_transport_opts [timeout: :infinity]
 
   @impl true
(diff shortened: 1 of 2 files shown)


lib/grpc/client/adapters/mint.ex
```diff
@@ -119,12 +125,22 @@ defmodule GRPC.Client.Adapters.Mint do
       |> Keyword.get(:transport_opts, [])
       |> Keyword.merge(ssl)
 
-    [transport_opts: Keyword.merge(@default_transport_opts, transport_opts)]
+    client_settings = Keyword.get(opts, :client_settings, @default_client_settings)
+
+    [
+      transport_opts: Keyword.merge(@default_transport_opts, transport_opts),
+      client_settings: client_settings
```
sleipnir, line 132: It might be useful to update the module documentation to mention the new client_settings option and its defaults. This will help users to understand the new configurations they can use and modify if needed.

sleipnir (changes requested):

lib/grpc/client/adapters/mint.ex
beligante: Ohhh yeah -- M'bad. I'll work on this later today

beligante: Check [2f4623a](https://github.com/elixir-grpc/grpc/pull/378/commits/2f4623a86e5b8ece04ec9e032a63c1f3e1347f05) and let me know if this is helpful enough. I tried to keep short, but still informative. I can expand if needed

lib/grpc/client/adapters/mint.ex
```diff
@@ -1,6 +1,6 @@
 defmodule GRPC.Client.Adapters.Mint do
   @moduledoc """
-  A client adapter using mint
+  A client adapter using Mint
```
sleipnir, line 3:
```suggestion
  A client adapter using Mint.
```

sleipnir (approved):

sleipnir:
Thank you @beligante 
You pretend to follow soon with https://github.com/elixir-grpc/grpc/issues/377 ?
I think we are accumulating enough in main to generate a new version. But it's very interesting and if we work on it quickly we can wait.

beligante:
> Thank you @beligante 
> You pretend to follow soon with https://github.com/elixir-grpc/grpc/issues/377 ?
> I think we are accumulating enough in main to generate a new version. But it's very interesting and if we work on it quickly we can wait.

Yup! I've finished the code changes already, I'm working on some tests. Probably later today

PR #395, opened 2024-12-28
Encode protobuf to iodata for better performance

v0idpwn:


v0idpwn:
We can do the same for JSON once https://github.com/elixir-protobuf/protobuf/pull/388 is released.

lib/grpc/client/adapter.ex
```diff
@@ -14,7 +14,7 @@ defmodule GRPC.Client.Adapter do
 
   @callback disconnect(channel :: Channel.t()) :: {:ok, Channel.t()} | {:error, any()}
 
-  @callback send_request(stream :: Stream.t(), contents :: binary(), opts :: keyword()) ::
```
polvalente, line 17: This is technically a breaking change, so let's hold off until we're ready for a new minor release

v0idpwn: Agreed :)

lib/grpc/codec/proto.ex
```diff
@@ -6,7 +6,7 @@ defmodule GRPC.Codec.Proto do
   end
 
   def encode(%mod{} = struct) do
```
v0idpwn, line 8:
```suggestion
  def encode(struct) do
```

sleipnir:
@polvalente I think it's time for us to move on with this. Wdyt?

polvalente (approved):

polvalente:
> @polvalente I think it's time for us to move on with this. Wdyt?

We gotta bump to 0.10 upon merging this (which we can already).

 In that case, I think we could also start using Keyword.validate! throughout the lib and drop support from Elixir 1.12 in the process.

sleipnir:
> > @polvalente I think it's time for us to move on with this. Wdyt?
> 
> We gotta bump to 0.10 upon merging this (which we can already).
> 
> In that case, I think we could also start using Keyword.validate! throughout the lib and drop support from Elixir 1.12 in the process.

I agree

v0idpwn:
If you're interested, I can send a PR for JSON as well, but it would require bumping protobuf to 0.14

sleipnir:
> If you're interested, I can send a PR for JSON as well, but it would require bumping protobuf to 0.14

Of course it would be very welcome.

PR #402, opened 2025-01-29
fix: Use atom in pattern matching of http version

JohanKarlbergg:
cowboy:http_version seems to be an atom

### DIFF
--- a/lib/grpc/server/adapters/cowboy/handler.ex
+++ b/lib/grpc/server/adapters/cowboy/handler.ex
@@ -619,7 +619,7 @@ defmodule GRPC.Server.Adapters.Cowboy.Handler do
     {:ok, "proto"}
   end
 
-  defp transcode?(%{version: "HTTP/1.1"}), do: true
+  defp transcode?(%{version: :"HTTP/1.1"}), do: true
 
   defp transcode?(req) do
     case find_content_type_subtype(req) do


polvalente:
Could you add a test that requires this fix to pass?

JohanKarlbergg:
> Could you add a test that requires this fix to pass?

I was debugging something and saw that the type was incorrect, that is how I found it.
I've looked at the code a bit more and I can't imagine a case when this code is required. It is maybe a better idea to remove it rather than introducing a new behavior.

The history on this piece of code is limited, so I don't know what the intent of it is.

polvalente:
I'd rather only commit this change if we have a case that shows the value as it is right now breaks something.

Unfortunately, I don't know this part of the code well enough to say whether this is correct or not.

JohanKarlbergg:
> I'd rather only commit this change if we have a case that shows the value as it is right now breaks something.
> 
> Unfortunately, I don't know this part of the code well enough to say whether this is correct or not.

Agreed, I'll close this PR.

sleipnir:
@polvalente in cowboy version is an atom https://github.com/ninenines/cowboy/blob/fcab905ecac3adc77348880c9702e53d65681344/src/cowboy_req.erl#L122 
but I think we can simply ensure that the value is properly sanitized and treated as both an atom and a string. I may do that at some point...

PR #404, opened 2025-02-10
allow-local-unix-socket

nicksanford:


lib/grpc/server/adapters/cowboy.ex
```diff
@@ -293,6 +293,9 @@ defmodule GRPC.Server.Adapters.Cowboy do
         :undefined ->
           raise "undefined address for ranch server"
 
+        :local ->
```
polvalente, line 296: As with the other PR, please add tests for the change.

polvalente: I believe when using `:local`, the handling of the address would be a bit different as it needs the socket file returned in `port` to decide which is the unix file descriptor handling the connection

polvalente: Actually, seeing as this is only used in logs, we can perhaps add that info to `Logger.metadata` and return `"local"` as it is now.

PR #405, opened 2025-02-10
support-empty-requests-and-responses

nicksanford:


lib/grpc/server/transcode.ex
```diff
@@ -70,6 +69,8 @@ defmodule GRPC.Server.Transcode do
     Map.get(response_body, key)
   end
 
+  def map_response_body(%{}, response_body), do: response_body
```
polvalente, line 72: This doesn't match on an empty map. It matches on any map. You need to use when map_size(arg) == 0

polvalente: Also, please add tests for this

sleipnir:
@nicksanford @polvalente I was wondering why this was necessary. Can you explain it to me?

polvalente:
> @nicksanford @polvalente I was wondering why this was necessary. Can you explain it to me?

I was hoping to discover when reading the test case I requested 😅

PR #410, opened 2025-03-21
chore: update protobuf library version

polvalente:
As suggested by @jeregrine


mix.exs
```diff
@@ -40,14 +40,8 @@ defmodule GRPC.Mixfile do
...
       {:castore, "~> 0.1 or ~> 1.0", optional: true},
-      {:protobuf, "~> 0.11"},
+      {:protobuf, "~> 0.14"},
       {:protobuf_generate, "~> 0.1.1", only: [:dev, :test]},
-      {:googleapis,
-       github: "googleapis/googleapis",
-       branch: "master",
-       app: false,
-       compile: false,
-       only: [:dev, :test]},
```
polvalente, line 50: This is the main point of the update.
Should we also regenerate the example protos in the newer version?

sleipnir: Yes, I believe that would be ideal.

sleipnir (approved):

mix.exs
```diff
@@ -41,7 +41,6 @@ defmodule GRPC.Mixfile do
       {:cowlib, "~> 2.12"},
       {:castore, "~> 0.1 or ~> 1.0", optional: true},
       {:protobuf, "~> 0.14"},
-      {:protobuf_generate, "~> 0.1.1", only: [:dev, :test]},
```
sleipnir, line 44: @polvalente By doing this http transcoding will stop working (I remembered this now)

polvalente: I don't think it does break. It broke when I stopped using the protobuf generate, but if you look at the transcode_messages.pb.ex file, you'll see the transcoding is still there

sleipnir: > I don't think it does break. It broke when I stopped using the protobuf generate, but if you look at the transcode_messages.pb.ex file, you'll see the transcoding is still there

But if the user regenerates the files from these examples would they still be present?

polvalente: Yes, as I've regenerated them. But they need to be generated with the protobut_generate plugin instead of protoc directly

polvalente: Note that the protobuf version is 0.14 in the comitted files

sleipnir: > Yes, as I've regenerated them. But they need to be generated with the protobut_generate plugin instead of protoc directly

My concern is how we guide users on these issues. Since this is an example of transcoding, this caught my attention. I don't know if our documentation is clear enough for the public.

polvalente: https://github.com/elixir-grpc/grpc?tab=readme-ov-file#http-transcoding

polvalente: I think we need to bring the protobuf_generate dep back to the examples. But otherwise it seems pretty clear how to generate things

sleipnir: I agree. Thank you

Issue #237, opened 2022-07-19
@AleksandarFilipov I think we could use the interceptors in a similar way to how Tesla modifies the current response given the result of `next.(stream, req)`


AleksandarFilipov:
@AleksandarFilipov I think we could use the interceptors in a similar way to how Tesla modifies the current response given the result of `next.(stream, req)`

https://github.com/elixir-tesla/tesla/blob/master/lib/tesla/middleware/compression.ex

In this example, they decide if the resp body needs to be decompressed based on the result tuple tag. We can provide a custom interceptor that injects response headers based on a predicate MFA tuple (that receives the stream and returns true if the headers should be injected). Something like:

```elixir
defmodule MyServer do
  use GRPC.Endpoint

  intercept GRPC.Interceptors.ResponseHeaders, should_inject: {__MODULE__, :"add_cors?", 1}, headers: [{"Access-Control-Allow-Origin", "*"}, {"Access-Control-Allow-Headers", "Content-Type, x-grpc-web, x-user-agent"}]

  def add_cors?(...), do: ...
end
```

The one thing that might need to change (I'm not _that_ familiar with the code base yet) is that this only works if `intercept`s are called before `stream_reply`. 

I'm almost sure that `intercepts` are only called after the reply, though. If so, we'll need to introduce something like `reply_intercept` that will be called right before the reply, but the overall idea remains the same.
This idea would allow us to remove the Codec from the `server_headers` match as well, while still allowing us to modify the headers (the whole connection, really), as we see fit.

Shall we move this into a new issue?

_Originally posted by @polvalente in https://github.com/elixir-grpc/grpc/issues/206#issuecomment-1186731225_

polvalente:
Moving this to 0.7 because it's been too long since 0.5 and there are people depending on us releasing an update to Hex.

aseigo:
Is there any update on this? Trying to get grpcweb working, I found that the missing part was the headers. 

One of the main things missing is that by default the grpcweb code does a CORS preflight check using the OPTIONS HTTP verb, and then setting the appropriate headers.

[...]

PR #412, opened 2025-04-21
Access mode detection (transcode vs grpcweb vs grpc)

aseigo:
This introduces:

* preflight request detection, allowing sensible responses to CORS preflight checks using the OPTIONS http verb
* client type detection, which may be one of grpc, grpcweb, or web (the latter implying a need for http transcoding)
* an interceptor that applies CORS headers for non-grpc requests

This is a draft for discussion in #237

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,26 +11,26 @@ An Elixir implementation of [gRPC](http://www.grpc.io/).
 
 ## Table of contents
 
-- [Installation](#installation)
-- [Usage](#usage)
-  - [Simple RPC](#simple-rpc)
-  - [HTTP Transcoding](#http-transcoding)
-  - [Start Application](#start-application)
-- [Features](#features)
-- [Benchmark](#benchmark)
-- [Contributing](#contributing)
+-   [Installation](#installation)
+-   [Usage](#usage)
+    -   [Simple RPC](#simple-rpc)
+    -   [HTTP Transcoding](#http-transcoding)
+    -   [Start Application](#start-application)
+-   [Features](#features)
+-   [Benchmark](#benchmark)
+-   [Contributing](#contributing)
 
 ## Installation
 
 The package can be installed as:
 
-  ```elixir
-  def deps do
-    [
-      {:grpc, "~> 0.9"}
-    ]
-  end
-  ```
+```elixir
+def deps do
+  [
+    {:grpc, "~> 0.9"}
+  ]
+end
+```
 
 ## Usage
 
@@ -96,7 +96,7 @@ end
 
 We will use this module [in the gRPC server startup section](#start-application).
 
-**__Note:__** For other types of RPC call like streams see [here](interop/lib/interop/server.ex).
+****Note:**** For other types of RPC call like streams see [here](interop/lib/interop/server.ex).
 
 ### **HTTP Transcoding**
 
@@ -152,6 +152,7 @@ mix protobuf.generate \
 ```
 
 3. Enable http_transcode option in your Server module
+
 ```elixir
 defmodule Helloworld.Greeter.Server do
   use GRPC.Server,
@@ -169,7 +170,12 @@ See full application code in [helloworld_transcoding](examples/helloworld_transc
 
 ### **CORS**
 
-When accessing gRPC from a browser via HTTP transcoding or gRPC-Web, CORS headers may be required for the browser to allow access to the gRPC endpoint. Adding CORS headers can be done by using the included `Interceptor` in your `Endpoint` module:
+When accessing gRPC from a browser via HTTP transcoding or gRPC-Web, CORS headers may be required for the browser to allow access to the gRPC endpoint. Adding CORS headers can be done by using the included `Interceptor` in your `Endpoint` module, configuring it with an `allow_origin` and, optionally, `allow_headers`:
+
+-   `allow_origin` - Required. A string containing the allowed origin, or a remote function (e.g. `&MyApp.MyModule.function/2)`) which takes a `req` and a `stream` and returns a string.
+-   `allow_headers` - A string containing the allowed headers, or a remote function (e.g. `&MyApp.MyModule.function/2)`) which takes a `req` and a `stream` and returns a string. If not provided, the value of the `"access-control-request-headers"` request header from the client will be used in the response.
+
+Example:
 
 ```elixir
 # Define your endpoint
@@ -177,28 +183,10 @@ defmodule Helloworld.Endpoint do
   use GRPC.Endpoint
 
   intercept GRPC.Server.Interceptors.Logger
-  intercept GRPC.Server.Interceptors.CORS
+  intercept GRPC.Server.Interceptors.CORS, allow_origin: "mydomain.io"
   run Helloworld.Greeter.Server
 end
 ```
-
-By default, the CORS `Interceptor` responds to CORS requests with a permissive response that allows *all* points of origin to access the service. This may be undesirable, particularly in production environments. For this reason, both the allowed origin and headers are configurable:
-
-```elixir
-# Define your endpoint
-defmodule Helloworld.Endpoint do
-  use GRPC.Endpoint
-
-  intercept GRPC.Server.Interceptors.Logger
-  intercept GRPC.Server.Interceptors.CORS, allow_origin: "mydomain.com", allow_headers: "Authorization"
-  run Helloworld.Greeter.Server
-end
-```
-
-The set of configuration directives supported include:
-
-  * `allow_origin` - A string containing the allowed origin(s), or a remote function (e.g. `&MyApp.MyModule.function/2)`) which takes a `req` and a `stream` and returns a string. Defaults to `"*"`, which will allow all origins to access this endpoint.
-  * `allow_headers` - A string containing the allowed headers, or a remote function (e.g. `&MyApp.MyModule.function/2)`) which takes a `req` and a `stream` and returns a string. Defaults to the value of the `"access-control-request-headers"` request header from the client.
 
 ### **Start Application**
 
@@ -256,18 +244,18 @@ The accepted options for configuration are the ones listed on [Mint.HTTP.connect
 
 ## Features
 
-- Various kinds of RPC:
-  - [Unary](https://grpc.io/docs/what-is-grpc/core-concepts/#unary-rpc)
-  - [Server-streaming](https://grpc.io/docs/what-is-grpc/core-concepts/#server-streaming-rpc)
-  - [Client-streaming](https://grpc.io/docs/what-is-grpc/core-concepts/#client-streaming-rpc)
-  - [Bidirectional-streaming](https://grpc.io/docs/what-is-grpc/core-concepts/#bidirectional-streaming-rpc)
-- [HTTP Transcoding](https://cloud.google.com/endpoints/docs/grpc/transcoding)
-- [TLS Authentication](https://grpc.io/docs/guides/auth/#supported-auth-mechanisms)
-- [Error handling](https://grpc.io/docs/guides/error/)
-- Interceptors (See [`GRPC.Endpoint`](https://github.com/elixir-grpc/grpc/blob/master/lib/grpc/endpoint.ex))
-- [Connection Backoff](https://github.com/grpc/grpc/blob/master/doc/connection-backoff.md)
-- Data compression
-- [gRPC Reflection](https://github.com/elixir-grpc/grpc-reflection)
+-   Various kinds of RPC:
+    -   [Unary](https://grpc.io/docs/what-is-grpc/core-concepts/#unary-rpc)
+    -   [Server-streaming](https://grpc.io/docs/what-is-grpc/core-concepts/#server-streaming-rpc)
+    -   [Client-streaming](https://grpc.io/docs/what-is-grpc/core-concepts/#client-streaming-rpc)
+    -   [Bidirectional-streaming](https://grpc.io/docs/what-is-grpc/core-concepts/#bidirectional-streaming-rpc)
+-   [HTTP Transcoding](https://cloud.google.com/endpoints/docs/grpc/transcoding)
+-   [TLS Authentication](https://grpc.io/docs/guides/auth/#supported-auth-mechanisms)
+-   [Error handling](https://grpc.io/docs/guides/error/)
+-   Interceptors (See [`GRPC.Endpoint`](https://github.com/elixir-grpc/grpc/blob/master/lib/grpc/endpoint.ex))
+-   [Connection Backoff](https://github.com/grpc/grpc/blob/master/doc/connection-backoff.md)
+-   Data compression
+-   [gRPC Reflection](https://github.com/elixir-grpc/grpc-reflection)
 
 ## Benchmark
 
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ end
 
 We will use this module [in the gRPC server startup section](#start-application).
 
-****Note:**** For other types of RPC call like streams see [here](interop/lib/interop/server.ex).
+**Note:** For other types of RPC call like streams see [here](interop/lib/interop/server.ex).
 
 ### **HTTP Transcoding**
 
--- a/README.md
+++ b/README.md
@@ -11,14 +11,14 @@ An Elixir implementation of [gRPC](http://www.grpc.io/).
 
 ## Table of contents
 
--   [Installation](#installation)
--   [Usage](#usage)
-    -   [Simple RPC](#simple-rpc)
-    -   [HTTP Transcoding](#http-transcoding)
-    -   [Start Application](#start-application)
--   [Features](#features)
--   [Benchmark](#benchmark)
--   [Contributing](#contributing)
+- [Installation](#installation)
+- [Usage](#usage)
+  - [Simple RPC](#simple-rpc)
+  - [HTTP Transcoding](#http-transcoding)
+  - [Start Application](#start-application)
+- [Features](#features)
+- [Benchmark](#benchmark)
+- [Contributing](#contributing)
 
 ## Installation
 
@@ -172,8 +172,8 @@ See full application code in [helloworld_transcoding](examples/helloworld_transc
 
 When accessing gRPC from a browser via HTTP transcoding or gRPC-Web, CORS headers may be required for the browser to allow access to the gRPC endpoint. Adding CORS headers can be done by using the included `Interceptor` in your `Endpoint` module, configuring it with an `allow_origin` and, optionally, `allow_headers`:
 
--   `allow_origin` - Required. A string containing the allowed origin, or a remote function (e.g. `&MyApp.MyModule.function/2)`) which takes a `req` and a `stream` and returns a string.
--   `allow_headers` - A string containing the allowed headers, or a remote function (e.g. `&MyApp.MyModule.function/2)`) which takes a `req` and a `stream` and returns a string. If not provided, the value of the `"access-control-request-headers"` request header from the client will be used in the response.
+- `allow_origin` - Required. A string containing the allowed origin, or a remote function (e.g. `&MyApp.MyModule.function/2)`) which takes a `req` and a `stream` and returns a string.
+- `allow_headers` - A string containing the allowed headers, or a remote function (e.g. `&MyApp.MyModule.function/2)`) which takes a `req` and a `stream` and returns a string. If not provided, the value of the `"access-control-request-headers"` request header from the client will be used in the response.
 
 Example:
 
@@ -244,18 +244,18 @@ The accepted options for configuration are the ones listed on [Mint.HTTP.connect
 
 ## Features
 
--   Various kinds of RPC:
-    -   [Unary](https://grpc.io/docs/what-is-grpc/core-concepts/#unary-rpc)
-    -   [Server-streaming](https://grpc.io/docs/what-is-grpc/core-concepts/#server-streaming-rpc)
-    -   [Client-streaming](https://grpc.io/docs/what-is-grpc/core-concepts/#client-streaming-rpc)
-    -   [Bidirectional-streaming](https://grpc.io/docs/what-is-grpc/core-concepts/#bidirectional-streaming-rpc)
--   [HTTP Transcoding](https://cloud.google.com/endpoints/docs/grpc/transcoding)
--   [TLS Authentication](https://grpc.io/docs/guides/auth/#supported-auth-mechanisms)
--   [Error handling](https://grpc.io/docs/guides/error/)
--   Interceptors (See [`GRPC.Endpoint`](https://github.com/elixir-grpc/grpc/blob/master/lib/grpc/endpoint.ex))
--   [Connection Backoff](https://github.com/grpc/grpc/blob/master/doc/connection-backoff.md)
--   Data compression
--   [gRPC Reflection](https://github.com/elixir-grpc/grpc-reflection)
+- Various kinds of RPC:
+  - [Unary](https://grpc.io/docs/what-is-grpc/core-concepts/#unary-rpc)
+  - [Server-streaming](https://grpc.io/docs/what-is-grpc/core-concepts/#server-streaming-rpc)
+  - [Client-streaming](https://grpc.io/docs/what-is-grpc/core-concepts/#client-streaming-rpc)
+  - [Bidirectional-streaming](https://grpc.io/docs/what-is-grpc/core-concepts/#bidirectional-streaming-rpc)
+- [HTTP Transcoding](https://cloud.google.com/endpoints/docs/grpc/transcoding)
+- [TLS Authentication](https://grpc.io/docs/guides/auth/#supported-auth-mechanisms)
+- [Error handling](https://grpc.io/docs/guides/error/)
+- Interceptors (See [`GRPC.Endpoint`](https://github.com/elixir-grpc/grpc/blob/master/lib/grpc/endpoint.ex))
+- [Connection Backoff](https://github.com/grpc/grpc/blob/master/doc/connection-backoff.md)
+- Data compression
+- [gRPC Reflection](https://github.com/elixir-grpc/grpc-reflection)
 
 ## Benchmark
 
--- a/README.md
+++ b/README.md
@@ -169,7 +169,7 @@ See full application code in [helloworld_transcoding](examples/helloworld_transc
 
 ### **CORS**
 
-When accessing grpc from a browser via http transcoding or grpcweb, CORS headers may be required for the browser to allow access to the grpc endpoint. Adding CORS headers is can be done by using the included `Interceptor` in your `Endpoint` module:
+When accessing gRPC from a browser via HTTP transcoding or gRPC-Web, CORS headers may be required for the browser to allow access to the gRPC endpoint. Adding CORS headers can be done by using the included `Interceptor` in your `Endpoint` module:
 
 ```elixir
 # Define your endpoint
--- a/README.md
+++ b/README.md
@@ -167,6 +167,39 @@ end
 
 See full application code in [helloworld_transcoding](examples/helloworld_transcoding) example.
 
+### **CORS**
+
+When accessing grpc from a browser via http transcoding or grpcweb, CORS headers may be required for the browser to allow access to the grpc endpoint. Adding CORS headers is can be done by using the included `Interceptor` in your `Endpoint` module:
+
+```elixir
+# Define your endpoint
+defmodule Helloworld.Endpoint do
+  use GRPC.Endpoint
+
+  intercept GRPC.Server.Interceptors.Logger
+  intercept GRPC.Server.Interceptors.CORS
+  run Helloworld.Greeter.Server
+end
+```
+
+By default, the CORS `Interceptor` responds to CORS requests with a permissive response that allows *all* points of origin to access the service. This may be undesirable, particularly in production environments. For this reason, both the allowed origin and headers are configurable:
+
+```elixir
+# Define your endpoint
+defmodule Helloworld.Endpoint do
+  use GRPC.Endpoint
+
+  intercept GRPC.Server.Interceptors.Logger
+  intercept GRPC.Server.Interceptors.CORS, allow_origin: "mydomain.com", allow_headers: "Authorization"
+  run Helloworld.Greeter.Server
+end
+```
+
+The set of configuration directives supported include:
+
+  * `allow_origin` - A string containing the allowed origin(s), or a remote function (e.g. `&MyApp.MyModule.function/2)`) which takes a `req` and a `stream` and returns a string. Defaults to `"*"`, which will allow all origins to access this endpoint.
+  * `allow_headers` - A string containing the allowed headers, or a remote function (e.g. `&MyApp.MyModule.function/2)`) which takes a `req` and a `stream` and returns a string. Defaults to the value of the `"access-control-request-headers"` request header from the client.
+
 ### **Start Application**
 
 1. Start gRPC Server in your supervisor tree or Application module:
--- a/README.md
+++ b/README.md
@@ -250,7 +250,7 @@ The accepted options for configuration are the ones listed on [Mint.HTTP.connect
 - [HTTP Transcoding](https://cloud.google.com/endpoints/docs/grpc/transcoding)
 - [TLS Authentication](https://grpc.io/docs/guides/auth/#supported-auth-mechanisms)
 - [Error handling](https://grpc.io/docs/guides/error/)
-- Interceptors (See [`GRPC.Endpoint`](https://github.com/elixir-grpc/grpc/blob/master/lib/grpc/endpoint.ex))
+- [Interceptors](`GRPC.Endpoint`)
 - [Connection Backoff](https://github.com/grpc/grpc/blob/master/doc/connection-backoff.md)
 - Data compression
 - [gRPC Reflection](https://github.com/elixir-grpc/grpc-reflection)
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ An Elixir implementation of [gRPC](http://www.grpc.io/).
 - [Usage](#usage)
   - [Simple RPC](#simple-rpc)
   - [HTTP Transcoding](#http-transcoding)
+  - [CORS](#cors)
   - [Start Application](#start-application)
 - [Features](#features)
 - [Benchmark](#benchmark)
@@ -170,10 +171,7 @@ See full application code in [helloworld_transcoding](examples/helloworld_transc
 
 ### **CORS**
 
-When accessing gRPC from a browser via HTTP transcoding or gRPC-Web, CORS headers may be required for the browser to allow access to the gRPC endpoint. Adding CORS headers can be done by using the included `Interceptor` in your `Endpoint` module, configuring it with an `allow_origin` and, optionally, `allow_headers`:
-
-- `allow_origin` - Required. A string containing the allowed origin, or a remote function (e.g. `&MyApp.MyModule.function/2)`) which takes a `req` and a `stream` and returns a string.
-- `allow_headers` - A string containing the allowed headers, or a remote function (e.g. `&MyApp.MyModule.function/2)`) which takes a `req` and a `stream` and returns a string. If not provided, the value of the `"access-control-request-headers"` request header from the client will be used in the response.
+When accessing gRPC from a browser via HTTP transcoding or gRPC-Web, CORS headers may be required for the browser to allow access to the gRPC endpoint. Adding CORS headers can be done by using `GRPC.Server.Interceptors.CORS` as an interceptor in your `GRPC.Endpoint` module, configuring it as decribed in the module documentation:
 
 Example:
 
--- a/lib/grpc/server.ex
+++ b/lib/grpc/server.ex
@@ -179,8 +179,7 @@ defmodule GRPC.Server do
                 | service_name: unquote(service_name),
                   method_name: unquote(to_string(name)),
                   grpc_type: unquote(grpc_type),
-                  http_method: unquote(http_method),
-                  http_transcode: unquote(http_transcode)
+                  http_method: unquote(http_method)
               },
               unquote(Macro.escape(put_elem(rpc, 0, func_name))),
               unquote(func_name)
@@ -253,7 +252,7 @@ defmodule GRPC.Server do
            codec: codec,
            adapter: adapter,
            payload: payload,
-           http_transcode: true
+           client_type: :web
          } = stream,
          func_name
        ) do
--- a/lib/grpc/server.ex
+++ b/lib/grpc/server.ex
@@ -152,7 +152,8 @@ defmodule GRPC.Server do
         path = "/#{service_name}/#{name}"
         grpc_type = GRPC.Service.grpc_type(rpc)
 
-        def __call_rpc__(unquote(path), :post, stream) do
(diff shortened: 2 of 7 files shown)


lib/grpc/server.ex
```diff
@@ -252,7 +252,7 @@ defmodule GRPC.Server do
            codec: codec,
            adapter: adapter,
            payload: payload,
-           http_transcode: true
+           client_type: :web
```
sleipnir, line 255: Hi @aseigo I don't think this is a good way to go, since http transcode is an extra layer allowing two-way communication, and replacing it with a client_type doesn't seem right to me. To me the correct way would be to simply enable cors or not, without much extra magic, so it would be a new parameter instead of replacing what already exists. This would also avoid a breaking change.

sleipnir (changes requested):
I have only one objection to the approach and I explained it in the comment

lib/grpc/server.ex
polvalente: Agreed. This should be an extra option because, if anything, the removal is a breaking change in itself

aseigo: Mm.. if this is considered part of the public API then, yes, it would be a breaking change and I can put it back in.

It is still redundant with the client type, but that's fine.

> http transcode is an extra layer allowing two-way communication

This is in the server code, which (at least afaict?) is not shared with the client. It does represent one half of the communication stream.

> To me the correct way would be to simply enable cors or not,

Of course, it's not that simple (unfortunately). 

The options preflight requests which are not only a different HTTP verb but also typically come with a different content-type header (which I ignored in this patch, actually) and no content (e.g. a protobuf encoded binary or a json document), and adding these headers to straight grpc connections where CORS isn't really a "thing" seems incorrect. If nothing else, it's moving more bytes across the pipe in a protocol where one of the intended benefits is greater efficiency relative to typical HTTP communication.

In getting this to work in the first place, all the code to detect the above cases is needed, regardless of whether or not it is recorded as such in the `Stream` struct. I decided to record it there as it makes it both explicit and avoids some repetition elsewhere.

sleipnir: > This is in the server code, which (at least afaict?) is not shared with the client. It does represent one half of the communication stream.
>

By definition http transcode coexists with the conventional protocol, in other words both grpc and http clients can be co-existing clients. This is what I meant. Therefore http_transcode is one thing and grpc web is another thing, and in this case they should have their own parameters here.

> Of course, it's not that simple (unfortunately).
>

I know, but in terms of configuration which was the discussion I raised here this is simple.

polvalente: Don't worry about commit history too much, we'll squash and merge the PR

aseigo: > Therefore http_transcode is one thing and grpc web

Indeed; grpcweb is not http transcoding. This is reflected in them being two different types of client interactions in the `client_type` member: `:web` vs `:grpcweb`. They use entirely different encoding schemes :)

>  It's just that the parameter that will configure all this shouldn't be the one you override.

I am not sure if we are talking about the same thing here. If we are, apologies in advance :)

Both http transcoding (aka `web`) and grpcweb benefit from having CORS support, which is one reason (along with simpler code in e.g. the cowboy handler during detection) to harmonize handling of those two separate ways of accessing a grpc service in this case.

It just happens that "web" access also utilizes http transcoding.

So while http transcoding and grpcweb use different encoding approaches (and have different capabilities as a result), they do share some needs in common. My thinking was that instead of enumerating all the things each form requires ("this one might need CORS headers", "this one needs http transcoding", etc.) as separate booleans in the `Stream` struct, it should instead record what sort of access is occurring (e.g. grpc, grpcweb, web) based on the client's reqs, and then wherever such differences in behaviour are required, those can be made based on the form of client access.

This not can keep `Stream` from growing further, but also makes it easier to add access pattern-specific code paths in future as `client_type` can be checked rather than each intended behavioral difference being recorded.

That said .. I'm having second thoughts about the name `client_type`. It's not really the "type" of the client (whatever that might mean), but the encoding and communications method employed on that connection. *ponders*

Perhaps `access_method`? Do you have any ideas for that?

aseigo: Apologies for the commit churn here :/ .. I tried with `access_mode` instead of `clent_type` and changed `:web` to `:http_transcoding` to align with the actual terms used around grpc. It reads more clearly now (at last, imho) and should be more readily understood by people coming to this bit of code who are familiar with grpc.

(As an aside: I was not sure if this library even had support for grpcweb at all as it wasn't clearly advertised in the code base and it did not actually work with basic examples ... I do feel this change has the minor side-benefit of being a lot clearer on the capabilities of the library. It is a very minor thing, but still something I realized while writing this changeset.)

sleipnir: What I'm saying is that we can't (under any circumstances) mix protocols (grpc, http transcoding, grpc-web) and any security access rules (CORs, Headers or whatever). And so it should look something like this:

```
codec: codec,
adapter: adapter,
payload: payload,
http_transcode: true
cors: [] # any specific CORS configuration set here
```

sleipnir: The user may perfectly want to expose grpc, transcoding and web in the same application, so access_mode property would not solve the problem either.

aseigo: > What I'm saying is that we can't (under any circumstances) mix protocols (grpc, http transcoding, grpc-web) and any security access rules (CORs, Headers or whatever)

The wrinkle is that things such as CORS are dependent on the method being used to access the GRPc server, along with application- and deployment-specific concerns. The CORS headers are not a constant, either: what it should be set to, if anything at all, is application- and deployment-specific.

Instead of managing such policies in the connection handler, it should describe the current connection (e.g. that it is being http transcoded, or that it is regular grpc, or..) so that deployment- and/or application-specific code can manage these concerns, taking appropriate actions based on that. That includes the application that is used this library.

Deciding on security concerns in the connection handler is building a ball of mud.

aseigo: > And the implementation of the subsequent solution must abstract away any complexities this represents.

I agree with this. 

It is the reason a CORS interceptor is provided in this branch. The default covers the most common (and only 'knowable') case, while being easily configurable. Requiring that it is added explicitly, as CORS headers are not required in all cases, make their use explicit, a choice, and visible to the application developers.

Putting CORS directly into the adapter pushes optional functionality into the main code path, functionality which any application using regular grpc does not need, while making it harder to share between any future implementations of GRPC.Server.Adapter. It also keeps that concern in one place where it can be more easily reasoned about and tested separate from the adapter itself.

Keeping CORS out of the Adapter itself requires that the Adapter needs to document the nature of the connection it is satisfying, something it was already doing for http transcoding, so that code specific to e.g. a security concern can take the corresponding actions required.

The interceptor is the abstraction of CORS, and it relies only on knowing what the connection method in use for that connection is.

sleipnir: @aseigo  I think we are really talking about different things. What bothers me is just how you are configuring things and not the consequences or implications of your implementation. As I said, we should not mix protocols and security rules in the configuration. How you are going to achieve the goals of enabling CORS is not up for discussion, you understand? I am only bothered by the parameter replacement, nothing more.

sleipnir: If you tell me that you can't achieve your goals without eliminating the old parameter, we can discuss that. But if there's a way you can accomplish what you need by adding a new parameter instead of replacing the existing one, then do that.

aseigo: > If you tell me that you can't achieve your goals without eliminating the old parameter

You mean `http_transcode`? That has already been put back for compatibility reasons. 

However, it isn't sufficient on its own. HTTP transcoding and grpcweb are not the same things.

aseigo: > we should not mix protocols and security rules in the configuration

We agree on that point! :) .. and why the security rule for CORS is provided in an Interceptor, while the protocol only handles detecting the access method.

> I am only bothered by the parameter replacement, nothing more.

The `http_transcode` parameter is still there (for compatibility, if nothing else).

sleipnir: > 
> The `http_transcode` parameter is still there (for compatibility, if nothing else).

Captura de tela de 2025-04-22 12-31-31

It still appears like this to me...

sleipnir: > 
> However, it isn't sufficient on its own. HTTP transcoding and grpcweb are not the same things.

I know that's why I suggested a specific configuration for CORS

sleipnir: What I'm saying is that CORS is one thing and protocols are another thing. So regardless of whether it's transcode or web, when these protocols are accessed from the web directly (by nature the web and by user use case the transcode) we need to ensure that CORS is enabled and configured correctly on the server (in this library).

aseigo: > > The `http_transcode` parameter is still there (for compatibility, if nothing else).
> 
>
> It still appears like this to me...

That's the wrong file. Look at `lib/grpc/server/adapters/cowboy/handler.ex`  and `lib/grpc/server/stream.ex`

lib/grpc/server/adapters/cowboy/handler.ex
```diff
@@ -69,8 +69,10 @@ defmodule GRPC.Server.Adapters.Cowboy.Handler do
         local: opts[:local],
         codec: codec,
         http_method: http_method,
+        http_transcode: access_mode == :http_transcoding,
```
sleipnir, line 72: I would do the test somewhere else and here I would just assign the resulting value

```elixir
       # in some where: http_transcode = access_mode == :http_transcoding
       
        # and here just
        http_transcode: http_transcoding,
```

sleipnir (changes requested):
Minor suggestion

lib/grpc/server.ex
sleipnir: > tl;dr questions:
> 

Good questions, and I don't have any concrete answers. My high-level thinking about the expected outcome of this PR would be the following:

1. CORS and protocols should be exposed to the end user as separate things.
- The premise for me is: The user has an obligation to know whether he is going to use a web client or not; therefore he should configure CORS separately.
- The grpc+web protocol should be explicitly enabled by the user, as well as the transcode protocol. This implies that if it is enabled and no CORS configuration is defined, the grpc server should raise an exception.

2. Currently the features are enabled by the user in the GRPC.Server options:

```elixir
defmodule Helloworld.Greeter.Server do
  use GRPC.Server,
  service: Helloworld.Greeter.Service, 
  http_transcode: true

  @spec say_hello(Helloworld.HelloRequest.t, GRPC.Server.Stream.t) :: Helloworld.HelloReply.t
  def say_hello(request, _stream) do  
    %Helloworld.HelloReply{message: "Hello #{request.name}"}
  end
end
```
- So the web protocol and cors could follow this same path. And in this case perhaps the most sensible option would be:
  1. replace the http_transcode option with an array of `additional_protocol: [:transcode, :web]` and add a configuration option for CORS. 
  2. Add a new parameter to enable the web protocol.

3. Whether the implementation will be on the adapter or as an interceptor, I am not opposed to either approach. As long as the interface is transparent to the user in some way like the one I proposed above or in some other similar way, everything will fit well into the overall scheme of the library.

lib/grpc/server/adapters/cowboy/handler.ex
aseigo: Done in b2f4598

lib/grpc/server.ex
aseigo: > The user has an obligation to know whether he is going to use a web client or not; therefore he should configure CORS separately

Agreed. And even more so, they need to know if CORS is even relevant to their usage, as it could be same origin ....

> The grpc+web protocol should be explicitly enabled by the user, as well as the transcode protocol

I'm not entirely sure I agree with this, tbh. If a proto has a `google.api.http` annotation, having to take an extra step to enable it in your service is ceremony .. with what benefit?

For grpcweb, it's the same protocol in terms of protos and what not, with a few small adjustments to make it fit into the browser. The limitations around bidirectional streaming are intended to go away once browser tech catches up a bit more, so at some point it would just be another client. The difference between grpcweb and grpc on the server side should, hopefully, be an implementation detail that the application developer need not worry themselves over.

As a user of the library, I'd prefer for things to work as defined in my proto files without having to figure out the correct rituals to go through to enable them. Again as a *user* of this library, I see zero benefit of being able to turn on or off grpcweb. What even is the use case for it?

Regardless of that point, this also does not follow:

 > This implies that if it is enabled and no CORS configuration is defined, the grpc server should raise an exception.

CORS is not needed if it is same origin. So if I have a reverse proxy or LB configured such that access to the web content comes from the same public origin as the grpc server (a common configuration in production environments when combining API with SPAs), then CORS may not required.

Having a hard require on CORS configuration is adding more ceremony where none may be required, which is not a great dev experience. :/

> So the web protocol and CORS could follow this same path. And in this case perhaps the most sensible option would be: 

Would it not be even more sensible to not require configuration, where "sensible" is measured in terms of a developer's experience for ease of use and expected behaviour? Configuration only makes sense if there is a use case for turning something off. 

Even the current `http_transcode` configuration makes very little sense in that regard, as it asks me to make a decision I've probably already made via my proto files, and learn how to make it work with the right ritual in configuration. For what benefit? At a minimum, it would be nicer if `http_transcode` defaulted to `true` as it has zero impact unless an HTTP API annotation is provided on an rpc call .. meaning a decision was already made in the proto to allow for HTTP transcoding.

Again, this is just my perspective as a user of this library. 

With my library developer hat on: having more configuration and conditional paths in the code base needs to come with a clear use case and benefit, otherwise it's just maintenance burden and a new home for possible bugs to lurk.

> Whether the implementation will be on the adapter or as an interceptor, I am not opposed to either approach.

My concern with putting this in the adapter is that it means should someone write an adapter that uses e.g. Bandit (likely if / as Bandit continues to gain usage), anything in the Cowboy adapter would need to be replicated in such an alternate adapter.

Maybe in Server, though? I do like the interceptor approach, though, as it makes it not only clearly and explicit (and optional), but it also keeps all the code for handling CORS separated into its own module, which makes testing and maintenance rather easier.

The issue of accessing CORS-related headers remains, though. :/ I'll take a gander at what might be done for that and propose something if it seems possible in a non-ugly way.

sleipnir: > > The user has an obligation to know whether he is going to use a web client or not; therefore he should configure CORS separately
> 
> Agreed. And even more so, they need to know if CORS is even relevant to their usage, as it could be same origin ....
> 
> > The grpc+web protocol should be explicitly enabled by the user, as well as the transcode protocol
> 
> I'm not entirely sure I agree with this, tbh. If a proto has a `google.api.http` annotation, having to take an extra step to enable it in your service is ceremony .. with what benefit?
> 
> For grpcweb, it's the same protocol in terms of protos and what not, with a few small adjustments to make it fit into the browser. The limitations around bidirectional streaming are intended to go away once browser tech catches up a bit more, so at some point it would just be another client. The difference between grpcweb and grpc on the server side should, hopefully, be an implementation detail that the application developer need not worry themselves over.
> 
> As a user of the library, I'd prefer for things to work as defined in my proto files without having to figure out the correct rituals to go through to enable them. Again as a _user_ of this library, I see zero benefit of being able to turn on or off grpcweb. What even is the use case for it?
> 

We cannot assume the intentions of users, just because an annotation exists or a service looks like another does not allow us to determine that these are absolute truths. A user may have received the protos from a central share or a third party and the annotations are not for the service they are implementing. This is the benefit, not exposing information that the user did not explicitly define. And that is without counting the technical issues that would be added if we wanted to infer all of our users' behaviors.
Remember that the behavior that works best for you is not always the same as what works best for others.

sleipnir: > 
> Regardless of that point, this also does not follow:
> 
> > This implies that if it is enabled and no CORS configuration is defined, the grpc server should raise an exception.
> 
> CORS is not needed if it is same origin. So if I have a reverse proxy or LB configured such that access to the web content comes from the same public origin as the grpc server (a common configuration in production environments when combining API with SPAs), then CORS may not required.
> 
> Having a hard require on CORS configuration is adding more ceremony where none may be required, which is not a great dev experience. :/
> 

When I said that an exception should be thrown if no CORS is configured when grpc+web is enabled, I was excluding the issue of defaults because I thought you would naturally go that route. If a default path exists that will not cause an error, then that path should be followed, as much of the current codebase already does in other situations. If not, and all other paths have been exhausted, then an exception should be thrown.

sleipnir: > The issue of accessing CORS-related headers remains, though. :/ I'll take a gander at what might be done for that and propose something if it seems possible in a non-ugly way.

Complexity is relative to requirements, as I said, a little formality at the cost of not inferring decisions from our users is better than defining rigid behaviors that everyone must follow. I don't think it's wise to change a large part of the library because of a CORS requirement.
I understand your frustration, but we have use cases where inferring that http_transcode was true by default would bring more harm than good. This may not seem smart for your use case, but I guarantee it is smart for others. In the end, we only try to find a good balance and not perfection.
I hope that with these guidelines you can move forward with this PR (which is very welcome).

aseigo: > I was excluding the issue of defaults because I thought you would naturally go that route.

Ah, so there would always be some configuration, either user-provided or by default. The user could, of course, throw their own exceptions to halt things (e.g. in an Interceptor, or what-have-you), but there would always be *some* configuration: explicit or defaults.

That makes sense, imo :)

> I understand your frustration,

It's ok, I'm not frustrated :)

> we have use cases where inferring that http_transcode was true by default would bring more harm than good. 

Interesting. Could you share what those are, just so I could build a better understand myself?

FWIW, if such use cases do exist, I can imagine providing a means to turn *off* transcoding. I'm not sure what the argument for the default being false is? 

(And yes, we are drifting from the original topic a bit here...)

> A user may have received the protos from a central share or a third party and the annotations are not for the service they are implementing. This is the benefit, not exposing information that the user did not explicitly define.

What information is it exposing?

It goes through the same pipeline, on the same socket, calling the same rpc functions in the app's Service module(s). 

The (main) difference is in how the messages are encoded for transport, which is entirely encapsulated by this library, no? I'm missing how information is exposed.

sleipnir: @aseigo https://github.com/eigr/spawn creates the transcoding automatically if there are annotations in the protobuf (just like you want), but we need to check if the user wants to expose these http services to the world via Ingress (kubernetes) otherwise we could expose the private behavior of our actors to the entire world. That's why it's good to ask beforehand if this is really what the user wants to do.

sleipnir: Good luck with the PR, I'm looking forward to seeing CORS support work. If I can help with anything else let me know.

aseigo: > but we need to check if the user wants to expose these http services to the world via Ingress (kubernetes) otherwise we could expose the private behavior of our actors to the entire world. 

If they are already exposing a grpc endpoint how is that meaningfully different?  If one has a grpc endpoint on a network, it is either secured (e.g. with API authorization tokens,) or it is "public". All three access methods use the same pathways for that.  

So, how does transcoding affect exposure? This is the part I'm obviously missing here :)

> the behavior of grpc calls is one (private) and the behavior of external remote calls via http is another

Aren't they are equally public or equally private? As soon as the RPC, in any of its forms, is exposed to the network it is not private. An app can be doing http transcoded RPC on a non-public (or even a disconnected) network ... 

When you say "private", do you mean something other than its visibility on a / the network?



The one difference I *can* see in the codebase is http transcoding "exposes" the code in `GRPC.Server.Transcode` to queries it processes., which means a (slightly) larger attack surface. 

But it is only slightly larger,  and the hope would be that since this library hosts a (potentially, anyways) network-facing code path, that this library is secure in general . As an app writer, I'm trusting this library with ingress of untrusted data, transcoding or no.

(There are other things I've noticed regarding security in the code base unrelated to this issue that I may take up in future PRs ... but one thing at a time!)

sleipnir: > > but we need to check if the user wants to expose these http services to the world via Ingress (kubernetes) otherwise we could expose the private behavior of our actors to the entire world.
> 
> If they are already exposing a grpc endpoint how is that meaningfully different? If one has a grpc endpoint on a network, it is either secured (e.g. with API authorization tokens,) or it is "public". All three access methods use the same pathways for that.
> 

Anyway, I think I've explained the points. I'm sorry I can't help you more with this. Requirements are requirements, maybe you just don't understand them at this point, I think this discussion won't be productive to continue. We can just assume that it's a premise, right?

sleipnir: @aseigo I really appreciate the clarity and the thoughtful approach to security and exposure. These kinds of discussions are essential for the quality and evolution of the project. 👏

Regarding the HTTP transcoding point: your technical argument is absolutely valid — from a network standpoint, once something is exposed, both gRPC and HTTP can be equally public or private depending on the environment’s security setup.

However, within the logical and semantic context of this project, HTTP transcoding carries a specific meaning: it represents an explicit decision by the user to make that behavior externally accessible, even when annotations already exist in the .proto files.

That’s why our current convention requires a clear signal of that intention. In our design (spawn related as example), not every service with transcoding annotations is automatically exposed — this logical separation allows us to maintain a more predictable and secure architecture, especially since HTTP transcoding is only enabled for specific actors, and some behaviors are intentionally internal by design.

In other words, it’s not just about network exposure — it’s about intended usage, and for that reason, it’s important to keep this decision point explicit in the configuration workflow. This distinction may seem subtle, but it's proven to be quite effective in helping us maintain control and predictability across the system.

If you’re up for it, we can think together about ways to make this flow even clearer — maybe through documentation for example. Also, I’m really looking forward to your future security-related contributions — your attention to detail will be a huge asset to the project!

sleipnir: I believe that with all the examples and context we've discussed, we can consider this matter settled for now and move forward by focusing on the CORS implementation.

Thanks again for the thoughtful discussion and your contributions — let’s keep the momentum going! 🚀

aseigo: Fair enough. I will start another discussion thread below to see where we are now with the patch.

aseigo:
I just pushed a couple more commits to this branch which:L

* adds the http req headers to the `Stream` struct
* uses those in the included CORS `Interceptor` to only add headers when it's a cors request.

I debated adding the http request headers to the `Stream` struct until I looked at some of the use cases I already have in projects that do similar things, as well as the documentation for CORS extenstions for other languages and frameworks ... and the headers they need to look at to make an informed decision varies greatly: authorization, host, etc.

To make the included Interceptor generically usable (or make it plausible for others to write their own), it appears "necessary" to include all the req headers. Without changing the `Interceptor` API pattern, this means adding it to the `Stream`.

Thoughts on that are welcome.

FWIW, I have done functional testing with a SPA written with React, next.js, and the `grpc-web` npm package. 

Other summary points that I will reiterate here for clarity and comment:

* http_transcoding members stay where they are, and with defaults retained as they currently are as well
* for grpcweb, the default is no CORS headers, which restricts access by default
* CORS headers are sent via an `Interceptor`, making inclusion and configuration of CORS (and therefore grpcweb access) explicit and opt-in

Anything else? Other thoughts?

I may need to write some more unit tests somewhere around `Server` .. I have cleaned up formatting (still have to rebuild my lsp for the version of Elixir/Erlang this repo uses.. )

sleipnir:
> I just pushed a couple more commits to this branch which:
> 
> * adds the http req headers to the `Stream` struct
> * uses those in the included CORS `Interceptor` to only add headers when it's a cors request.
> 
> I debated adding the http request headers to the `Stream` struct until I looked at some of the use cases I already have in projects that do similar things, as well as the documentation for CORS extenstions for other languages and frameworks ... and the headers they need to look at to make an informed decision varies greatly: authorization, host, etc.
> 
> To make the included Interceptor generically usable (or make it plausible for others to write their own), it appears "necessary" to include all the req headers. Without changing the `Interceptor` API pattern, this means adding it to the `Stream`.
> 
> Thoughts on that are welcome.
> 
> FWIW, I have done functional testing with a SPA written with React, next.js, and the `grpc-web` npm package.
> 
> Other summary points that I will reiterate here for clarity and comment:
> 
> * http_transcoding members stay where they are, and with defaults retained as they currently are as well
> * for grpcweb, the default is no CORS headers, which restricts access by default
> * CORS headers are sent via an `Interceptor`, making inclusion and configuration of CORS (and therefore grpcweb access) explicit and opt-in
> 
> Anything else? Other thoughts?
> 
> I may need to write some more unit tests somewhere around `Server` .. I have cleaned up formatting (still have to rebuild my lsp for the version of Elixir/Erlang this repo uses.. )

Great job!
It would be great to add a usage example with CORS settings and a client example to the documentation.

lib/grpc/server/interceptors/cors.ex
```diff
@@ -0,0 +1,62 @@
...
+    end
+  end
+
+  @impl true
+  def call(req, stream, next, allowed) do
+    if stream.access_mode != :grpc and
+         Map.get(stream.http_request_headers, "sec-fetch-mode") == "cors" do
+      stream.adapter.set_headers(stream.payload, %{
+        "access-control-allow-origin" => resolve_allowed(req, stream, allowed),
+        "access-control-allow-headers" => "content-type, x-grpc-web, x-user-agent, x-api-key"
```
sleipnir, line 50: Following the protocol guide you may need to change/add these headers: https://github.com/grpc/grpc-web/blob/master/doc/browser-features.md#cors-support

sleipnir (changes requested):
I have filed a request for changes but just so you can check a question I raised regarding the specification. Can you check it for us?

sleipnir (changes requested):

lib/grpc/server/interceptors/cors.ex
aseigo: Indeed. They need to follow the client, or be set to a value by the application that is more suitable. See bcd4ddba9e531caa2f92adedad76a1ce26cf7cb1

aseigo:
> It would be great to add a usage example with CORS settings and a client example to the documentation.

I have added one to the `README.md`. The module documentation already contained an example.

sleipnir (approved):

sleipnir:
I approved this PR to approve the pipeline, but in this specific case I would like a view from @polvalente before moving on to the merge, maybe I missed something along the way, a second opinion is always good

sleipnir:
After some adjustments to the pipeline the tests ran ok.

lib/grpc/server/interceptors/cors.ex
```diff
@@ -0,0 +1,98 @@
...
+  @impl true
+  def init(opts \\ []) do
+    # the funky first clause matches a 2-arity remote is_function
+    # note that this function is run in the context of a macro, which brings some limitations with it
+    allow_origin =
+      case Keyword.get(opts, :allow_origin) do
+        {:&, [], [{:/, [], [_signature, 2]}]} = fun -> fun
+        static when is_binary(static) -> static
+        _ -> "*"
+      end
```
polvalente, line 43: As this is running in a macro context (I can't recall why, but let's roll with this), the current code does not allow for a variable to be referenced here, I think. It will end up defaulting to the catch-all silently and this is a security risk.

Perhaps this validation should be done at run time?

lib/grpc/server/interceptors/cors.ex
```diff
@@ -0,0 +1,98 @@
...
+      stream.adapter.set_headers(stream.payload, headers)
+    end
+
+    next.(req, stream)
+  end
+
+  defp add_allowed_origins(headers, req, stream, allow) do
+    value =
+      case allow do
+        allow when is_function(allow, 2) -> allow.(req, stream)
```
polvalente, line 73: You can use apply(mod, name, [req, stream]) here if we move to the mod/name suggestion

lib/grpc/server/interceptors/cors.ex
```diff
@@ -0,0 +1,98 @@
+defmodule GRPC.Server.Interceptors.CORS do
+  @moduledoc """
+  Sends CORS headers when the client is calling the rpc via web transcoding or grpcweb.
```
polvalente, line 3:
```suggestion
  Sends CORS headers when the client is using RPC via Web transcoding or gRPC-web.
```

polvalente (commented):
Some security concerns, but overall looks good to me

README.md
```diff
@@ -167,6 +167,39 @@ end
...
+defmodule Helloworld.Endpoint do
+  use GRPC.Endpoint
+
+  intercept GRPC.Server.Interceptors.Logger
+  intercept GRPC.Server.Interceptors.CORS
+  run Helloworld.Greeter.Server
+end
+```
+
+By default, the CORS `Interceptor` responds to CORS requests with a permissive response that allows *all* points of origin to access the service. This may be undesirable, particularly in production environments. For this reason, both the allowed origin and headers are configurable:
```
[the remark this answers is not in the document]
polvalente: Yeah, make the option required then, without any defaults. We just can't default to allowing everything

README.md
```diff
@@ -167,6 +167,39 @@ end
...
+  intercept GRPC.Server.Interceptors.Logger
+  intercept GRPC.Server.Interceptors.CORS, allow_origin: "mydomain.com", allow_headers: "Authorization"
+  run Helloworld.Greeter.Server
+end
+```
+
+The set of configuration directives supported include:
+
+  * `allow_origin` - A string containing the allowed origin(s), or a remote function (e.g. `&MyApp.MyModule.function/2)`) which takes a `req` and a `stream` and returns a string. Defaults to `"*"`, which will allow all origins to access this endpoint.
+  * `allow_headers` - A string containing the allowed headers, or a remote function (e.g. `&MyApp.MyModule.function/2)`) which takes a `req` and a `stream` and returns a string. Defaults to the value of the `"access-control-request-headers"` request header from the client.
```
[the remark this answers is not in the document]
polvalente: Re headers:

We should add that link to the docs

Re function:

Given that we can enforce that it's a capture, we can keep it as is.

lib/grpc/server/interceptors/cors.ex
polvalente: Pass an atom/integer/tuple/list/variable referencing a binary instead of a capture or literal binary and this will fall through to "*".

polvalente: We should raise with ArgumentError

test/grpc/server/interceptors/cors_test.exs
```diff
@@ -0,0 +1,285 @@
...
+            @requested_allowed_headers
+          )
+    }
+
+    {:ok, :ok} =
+      CORSInterceptor.call(
+        request,
+        %{stream | access_mode: :grpcweb},
+        fn _request, _stream -> {:ok, :ok} end,
+        CORSInterceptor.init()
```
polvalente, line 175: This style of calling init explicitly won't allow for passing the capture without also allowing a function.

If we keep the capture as the chosen syntax, we need to guard against allowing the function AST.

IMO moving to the tuple will be simpler to support in both situations

lib/grpc/server/interceptors/cors.ex
aseigo: > Pass an atom/integer/tuple/list/variable referencing a binary instead of a capture or literal binary and this will fall through to "*".

This is actually covered in the documentation: a string or a remote function, otherwise is defaults to '*'. As was the default values used.

It's a moot point now, in any case, as I've removed the default for `allow_origin` altogether in  18ad11f69d38a4b75e312ba617b1d59e93ba5216 which should address this and another of your concerns.

test/grpc/server/interceptors/cors_test.exs
aseigo: > This style of calling init explicitly won't allow for passing the capture without also allowing a function.

This is a convenience for the unit test only. (... and is what the unit tests for the logging `Interceptor` does.)  The unit tests for utilizing a remote function use an `Endpoint`, as that way `init` is called at compile-time in a macro context. I would do this for all the tests, but it's far uglier and less clear so I followed the precedent set in `logger_test.exs` where possible.

Of course, an Interceptor's `init` is always, and only, called within a macro as this library treats an `Interceptor`'s `init` as a source of compile-time configuration data.

> If we keep the capture as the chosen syntax, we need to guard against allowing the function AST.

If I understand what you are suggesting here, that already isn't possible as it is being run within a macro and so fails at compile time.

polvalente: Let's use the other method, though, as your code heavily relies on macro comparison and that can't be tested if you're calling things as you are here

lib/grpc/server/interceptors/cors.ex
```diff
@@ -0,0 +1,97 @@
...
+
+  @behaviour GRPC.Server.Interceptor
+  @impl true
+  def init(opts \\ []) do
+    # the funky first clause matches a 2-arity remote is_function
+    # note that this function is run in the context of a macro, which brings some limitations with it
+    allow_origin =
+      case Keyword.get(opts, :allow_origin) do
+        {:&, [], [{:/, [], [_signature, 2]}]} = fun -> fun
+        static when is_binary(static) -> static
```
polvalente, line 41: You can add a explicit catch all here raising and ArgumentError for better UX.

This matching mode is also already enough to guard against runtime functions as I'd mentioned below.

lib/grpc/server/interceptors/cors.ex
```diff
@@ -0,0 +1,97 @@
...
+      end
+
+    allow_headers =
+      case Keyword.get(opts, :allow_headers) do
+        {:&, [], [{:/, [], [_signature, 2]}]} = fun -> fun
+        static when is_binary(static) -> static
+        _ -> nil
```
polvalente, line 48: I think we should also require "*" to be explicitly passed here

aseigo: "*" is not equivalent to what the current default is. To get the behaviour of "*", you already need to pass it explicitly. Rather, what this does by default is echo the requested headers, which is different to what "*" indicates.

(As an aside, what should *probably* be added for completeness is support for `Access-Control-Allow-Methods`)

polvalente: Right. My main point is: I heavily dislike the fallback to a permissive value. 

I'm open to whatever suggestions we can do to mitigate this, even if we end up having a `nil -> nil` explicit match.

lib/grpc/server/interceptors/cors.ex
aseigo: > You can add a explicit catch all here raising and ArgumentError for better UX.

see b51feda

> This matching mode is also already enough to guard against runtime functions as I'd mentioned below.

When called outside of a macro, yes. When called in a macro, it fails at compilation before it even gets there.

lib/grpc/server/interceptors/cors.ex
aseigo: It's not a permissive value; it's what one would expect as a default if the origin is allowed.

polvalente: The problem is that if, say, the user constructs a binary via concatenation, they'll end up falling back to this default unknowingly.

This is why I think we shouldn't default to it via catch-all

polvalente: To be clear, the AST of a concatenation isn't a raw binary

polvalente: We can default to nil and hard-match on nil to get this base case

aseigo: > The problem is that if, say, the user constructs a binary via concatenation, they'll end up falling back to this default unknowingly.

Do you mean if they do something like: `allow_headers: "header1" <> "header2"` or `allow_headers: Enum.join(["headers1", "headers2"], ",")`? Because if so .. that all works.

polvalente: You need to test those 2 alternatives in the context of compile time tests. Your tests use runtime values and then those will indeed return a raw binary.

That's not the case if you're capturing the AST.

I didn't check if you added the compile time tests

aseigo:
My available time for this PR has run out. Please feel free to do with it as you wish from here.

lib/grpc/server/interceptors/cors.ex
polvalente: I am not trying things out. 

The thing that's bugging me is that the way your case is written, I don't see how it both supports the capture AST match and the AST for binary concatenation.

And as I requested, we really need the tests to mirror the use case.

If you want to show something works, you need to add a test for that, not just say it works.

I understand this process is frustrating, and this really is the last piece to converge on before merging this.

If you're not up for it, say the word and I'll take over the PR (you'll of course be credited in the merge).

polvalente: As an aside, you should be mindful of language in public forums. No matter how frustrating things can be, remember maintainers for libraries, as a rule of thumb, do it in their free time and it's not nice to read things that have a harsh tone.

aseigo: I wasn't going to say anything further, but as you are offering personal feedback, I can only assume you are also open for the same

>  remember maintainers for libraries, as a rule of thumb, do it in their free time

I've been contributing to, and maintaining, open source software for nearly 30 years, including much bigger things than this library. I did so full time for over a decade, at one point. So I am quite aware of how the work gets done. To that point, this PR was done in my free time as well.

Now, when someone makes observations on how code works without checking it themselves, essentially guessing at things as if they were fact, it ends up being a request for others to do that work instead of you. For them to spend time and energy delving and perhaps making changes in response to those guesses. Guesses are often incorrect, and it communicates a lack of regard for other people's time and energy.

Numerous spurious comments have been made in the process of this PR, along with some valid points raised. I've generally dealt with the former with a kind response and/or  a change in the code that, while perhaps not strictly necessary, didn't hurt anything either. 

Up until your comment above, you didn't acknowledge any of this, either. When you were wrong you just moved on without acknowledgment, or made out that you were perhaps talking about some other aspect of things.

And that, taken together, is a recipe for losing contributors. I don't know if that matters to you (not every project cares, and that's perfectly within their rights).

Among the things I've learned over the years of both doing FOSS and generally working in teams is to try things out when I'm not sure, to ask questions rather than make assertions if I can, and not worry about responding quickly if I don't have time to do things correctly right now. Things will wait until tomorrow, or the next day when I do have time to do it better, especially with open source. When I've tried to rush responses rather than allow myself the time needed to look things through, the results have often been worse than desired.

> As an aside, you should be mindful of language in public forums. 
...
>  to read things that have a harsh tone.

Hmmmm.

And about this observation:

> If you want to show something works, you need to add a test for that, not just say it works.

a) it was you making assertions saying what does or does not work, without doing any testing; I've been testing literally everything I've done here.

b) I would have happily committed unit tests for that if I had an inkling it would matter. Your pattern of communication has shown me that there will probably be something else somewhere, you won't actually be testing or researching anything but you will be asking for me to make changes to this or that, whatever it might be.

Honestly, there is nothing hard about saying something like "Oh, that surprises me! Cool that it works, though! Would you mind adding tests for those two cases?" There's also nothing wrong with actually trying things out before you make assertions that waste both of our time.

I do wish you all the best in your endeavours, however, including this project.

polvalente: Thanks for the feedback, and perhaps most importantly, for the time spent here.

polvalente: @aseigo to be clear as this is related to one of the points in you message: I will think about your message carefully, specially in that it highlights some important points about how others might receive communication that's not necessarily aggressive in nature.

Again, thanks for the time invested and I'll get this PR through soon.

polvalente (approved):
Thank you for the contribution! This was a tricky one to get through.

PR #416, opened 2025-04-27
refactor: use Keyword.validate! and bump to v0.10 and elixir 1.15

polvalente:
This adds Keyword.validate! calls to places where we have a closed set of keys in keyword lists

### DIFF
--- a/lib/grpc/server/interceptors/cors.ex
+++ b/lib/grpc/server/interceptors/cors.ex
@@ -41,7 +41,6 @@ defmodule GRPC.Server.Interceptors.CORS do
     # This is not a full-on Macro context, so binary concatenations and
     # variables are handled before this step.
 
-    # TODO: use Keyword.validate! once we drop support for Elixir < 1.13
 
     opts = Keyword.validate!(opts, [:allow_origin, allow_headers: nil])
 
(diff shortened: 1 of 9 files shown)


mix.exs
```diff
@@ -1,13 +1,13 @@
 defmodule GRPC.Mixfile do
   use Mix.Project
 
-  @version "0.9.0"
+  @version "0.10.0"
```
polvalente, line 4: Also took the opportunity for bumping to 0.10

lib/grpc/server/interceptors/cors.ex
```diff
@@ -43,16 +43,10 @@ defmodule GRPC.Server.Interceptors.CORS do
 
     # TODO: use Keyword.validate! once we drop support for Elixir < 1.13
```
sleipnir, line 44: Remove the comment

sleipnir (approved):
Just remove the comment and let's go

lib/grpc/server/interceptors/cors.ex
sleipnir, line 44:
```suggestion
```

sleipnir (approved):

Issue #270, opened 2022-09-02
New API Proposal


sleipnir:
**Is your feature request related to a problem? Please describe.**

One thing that has always bothered me about the elixir-grpc API is that it differs a lot in how we handle Stream gRPC in other languages and why we need a separate module and functions to handle responses from a gRPC stream. I would like to handle a grpc stream as fluidly as we do in other languages in which basically the types generated for input and output of an rpc stream are a stream type in the language, whereas in the case of elixir-grpc an output type of a stream is not an output type of the elixir function, instead we use a helper module (Server.send_reply) to output a response but this does not reflect in the function's Typespec as a stream.
I know it's hard to visualize what I'm saying so I'll put some examples from other languages here to demonstrate:

Kotlin Raw:
```kotlin
override fun routeChat(requests: Flow<RouteNote>): Flow<RouteNote> =
  flow {
    // could use transform, but it's currently experimental
    requests.collect { note ->
      val notes: MutableList<RouteNote> = routeNotes.computeIfAbsent(note.location) {
        Collections.synchronizedList(mutableListOf<RouteNote>())
      }
      for (prevNote in notes.toTypedArray()) { // thread-safe snapshot
        emit(prevNote)
      }
      notes += note
    }
  }
```

kotlin with Akka gRPC:

```kotlin
override fun handle(streamIn: Source<EventSourcedProto.EventSourcedStreamIn, NotUsed>?): Source<EventSourcedProto.EventSourcedStreamOut, NotUsed> =
        streamIn!!.log("CloudState-User")
                .map {
                    if (it.hasInit()) {
                        return@map it.init
                    }

                    if (it.hasCommand()) {
                        return@map it.command
                    }

                    if (!it.hasEvent()) {
                    } else {
                        return@map it.event
                    }

                }
}
```

Java with Akka gRPC:
```java
@Override
  public Source<HelloReply, NotUsed> sayHelloToAll(Source<HelloRequest, NotUsed> in) {
    return in.runWith(inboundHub, system);
  }
```

Java Raw:

```java
@Override
public StreamObserver<Stock> bidirectionalStreamingGetListsStockQuotes(StreamObserver<StockQuote> responseObserver) {
    return new StreamObserver<Stock>() {
        @Override
        public void onNext(Stock request) {
            for (int i = 1; i <= 5; i++) {
                StockQuote stockQuote = StockQuote.newBuilder()
                  .setPrice(fetchStockPriceBid(request))
[...]

PR #418, opened 2025-05-09
[Feat]: Add support for streams pipelines

sleipnir:
This closes https://github.com/elixir-grpc/grpc/issues/270


sleipnir:
@polvalente, I think this is ready for review now. I would like to rename the modules:

`GRPC.Server.Stream` -> `GRPC.Server.Materializer`
and maybe their client-side equivalent.

Justification: I don't think they are a real Stream structure and in the new proposal they are only used to materialize the response to the client. It also looks a bit strange to have two modules named Stream in the library.
But I haven't gone ahead with this yet, waiting for your opinion.

The PR was a bit big but sorry there wasn't much that could be done since it's a likely breaking change.

lib/grpc/stream.ex
```diff
@@ -1,17 +1,433 @@
...
+
+  def to_flow!(%__MODULE__{flow: flow}), do: flow
+
+  @doc """
+  Executes the flow and emits responses into the provided gRPC server stream.
+
+  ## Parameters
+
+    - `flow`: A `GRPC.Stream` struct containing the flow to be executed.
+    - `stream`: A `GRPC.Server.Stream` to which responses are sent.
```
polvalente, line 222: I think we should do a full-on breaking change and merge GRPC.Server.Stream into this module. Then, these would be a single `stream` parameter. The fact that we have to Stream structs in the same library and that they both work in the same domain is a code smell to me, and I think you agree from something you commented elsewhere.

lib/grpc/stream.ex
```diff
@@ -1,17 +1,433 @@
...
+    - `opts`: Optional keyword list. Currently supported:
+    - `:dry_run` — If `true`, responses are not sent (used for testing or inspection).
```
polvalente, line 224: Missing indentation here.

In general, we can have a `## Options` section where we describe the options accepted in `opts`

lib/grpc/stream.ex
```diff
@@ -1,17 +1,433 @@
...
+
+  ## Returns
+
+    - `:ok` if the stream was processed successfully.
```
polvalente, line 228: Would it make sense to return `{:ok, stream}`?

lib/grpc/stream.ex
```diff
@@ -1,17 +1,433 @@
...
+    - For unary mode, returns the first item in the stream instead of sending responses.
+
```
polvalente, line 230: I'm not sure I like the polymorphic result type. Could we merge these 2 somehow?

lib/grpc/stream.ex
```diff
@@ -1,17 +1,433 @@
...
+        opts \\ []
+      ) do
+    dry_run? = Keyword.get(opts, :dry_run, false)
+    unary? = Keyword.get(flow_opts, :unary, false)
+
+    if unary? do
+      flow
+      |> Enum.to_list()
+      |> List.flatten()
+      |> List.first()
```
polvalente, line 248:
```suggestion
      # We have to call `Enum.to_list` because we want to actually run and materialize the full stream. List.flatten and List.first are used so that we can obtain the first result of the materialized list.
      flow
      |> Enum.to_list()
      |> List.flatten()
      |> List.first()
```

Let me know if the comment is conveying the intention. Otherwise, this could be `Enum.at(flow, 0)`

lib/grpc/stream.ex
```diff
@@ -1,17 +1,433 @@
...
+    else
+      flow
+      |> Flow.map(fn msg ->
+        unless dry_run?, do: send_response(from, msg)
+        flow
+      end)
+      |> Flow.run()
```
polvalente, line 255: I think dry_run would be cleaner if extracted to the outer if-else:

```elixir
cond do
  dry_run? ->
    ...
  unary? ->
    ...
   true -> 
     ...
end
```

This also makes dry-run work for unary flows as well, especially if we make the result value of an uniform type.

lib/grpc/stream.ex
```diff
@@ -1,17 +1,433 @@
...
+    end
+  end
+
+  @doc """
+  Sends a request to an external process and awaits a response.
+
+  If `target` is a PID, a message in the format `{:request, item, from}` is sent, and a reply
+  in the format `{:response, msg}` is expected.
+
+  If `target` is a GenServer module (atom), the message is `{:request, item}`, and a reply
```
polvalente, line 265: We should just send the same message to both. We shouldn't care what type of process is handling the message.

lib/grpc/stream.ex
```diff
@@ -1,17 +1,433 @@
...
+  in the format `{:response, msg}` is expected.
+
+  ## Parameters
+
+    - `stream`: The `GRPC.Stream` pipeline.
```
polvalente, line 270: This also goes to the point I made above. Here you call GRPC.Stream the stream. In the other function, GRPC.Stream is flow and GRPC.Server.Stream is the stream

lib/grpc/stream.ex
```diff
@@ -1,17 +1,433 @@
...
+    - `target`: Target process PID or GenServer module name.
```
polvalente, line 271:
```suggestion
    - `target`: Target process (pid or registered name).
```

lib/grpc/stream.ex
```diff
@@ -1,17 +1,433 @@
...
+  @spec ask(t(), pid | atom, non_neg_integer) :: t() | {:error, item(), reason()}
+  defdelegate ask(stream, target, timeout \\ 5000), to: Operators
+
+  @doc """
+  Same as `ask/3`, but raises an exception on failure.
+
+  ## Caution
+
+  This version propagates errors via raised exceptions, which can crash the Flow worker and halt the pipeline.
+  Prefer `ask/3` for production use unless failure should abort the stream.
```
polvalente, line 288:
```suggestion
  Prefer `ask/3` for production usage unless failure should abort the stream.
```

lib/grpc/stream.ex
```diff
@@ -1,17 +1,433 @@
...
+  """
+  @spec ask!(t(), pid | atom, non_neg_integer) :: t()
+  defdelegate ask!(stream, target, timeout \\ 5000), to: Operators
+
+  @doc """
+  Replaces the stream with a new one generated by the given factory function.
+  The factory function receives the metadata and the current item as arguments.
+  The factory function should return a new `GRPC.Stream` struct.
```
polvalente, line 296:
```suggestion
  The factory function returns a new `GRPC.Stream` struct.
```

lib/grpc/stream.ex
```diff
@@ -1,17 +1,433 @@
...
+
+  ## Caution
+  This function materializes the current stream and replaces it with a new one.
+  This can lead to performance issues if used excessively. In fact, it is more commonly used
+  in following the `single/2` of `single_as_ctx/3` functions.
```
polvalente, line 301:
```suggestion
  in following the `single/2` or `single_as_ctx/3` functions. 
```

lib/grpc/stream.ex
```diff
@@ -1,17 +1,433 @@
...
+    GRPC.Stream.single(request)
+    # Create new output stream from the request and drop the input
+    |> GRPC.Stream.replace(fn _metadata, msg ->
+      create_output_stream(msg) # returns some Enumerable
+      |> GRPC.Stream.from()
+    end)
+    |> GRPC.Stream.run_with(materializer)
+  """
+  @spec replace(t(), (map(), term -> term)) :: t()
+  defdelegate replace(stream, factory), to: Operators
```
polvalente, line 322: I'm not sure I understand the usefulness of this function

lib/grpc/stream.ex
```diff
@@ -1,17 +1,433 @@
...
+
+  @doc """
+  Filters the stream using the given predicate function in parallel.
```
polvalente, line 325:
```suggestion
  Filters the stream using the given predicate function.
  
  The filter function is applied concurrently to the stream entries, so it shouldn't rely on execution order.
```

lib/grpc/stream.ex
```diff
@@ -1,17 +1,433 @@
...
+  """
+  @spec filter(t(), (term -> term)) :: t
+  defdelegate filter(stream, filter), to: Operators
+
+  @doc """
+  Applies a mapping function and flattens the results (one level deep).
```
polvalente, line 331:
```suggestion
  Applies a function to each entry and concatenates the resulting lists.
```

lib/grpc/stream.ex
```diff
@@ -1,17 +1,433 @@
...
+
+  Useful for emitting multiple messages for each input.
+  """
+  @spec flat_map(t, (term -> Enumerable.t())) :: t()
+  defdelegate flat_map(stream, flat_mapper), to: Operators
+
+  @doc """
+  Applies a transformation function to each stream item.
```
polvalente, line 339:
```suggestion
  Applies a function to each stream item.
```

lib/grpc/stream.ex
```diff
@@ -1,17 +1,433 @@
...
+  """
+  @spec map(t(), (term -> term)) :: t()
+  defdelegate map(stream, mapper), to: Operators
+
+  @doc """
+  Applies a transformation function to each stream item, passing the context as an additional argument.
+  This is useful for operations that require access to the stream's headers.
+  """
+  @spec map_with_ctx(t(), (map(), term -> term)) :: t()
+  defdelegate map_with_ctx(stream, mapper), to: Operators
```
polvalente, line 349: Not sure why this one is necessary

lib/grpc/stream.ex
```diff
@@ -1,17 +1,433 @@
...
+
+  @doc """
+  Partitions the stream to allow grouping of items by key or condition.
+
+  Use this before stateful operations such as `reduce/3`.
+
+  ## Note
+
+  Excessive use of partitioning can impact performance and memory usage.
+  Only partition when required for correctness or performance.
```
polvalente, line 359: We should point to the Flow docs here, as this is really specific to the fact that we're relying on Flow for the orchestration.

lib/grpc/stream.ex
```diff
@@ -1,17 +1,433 @@
...
+  """
+  @spec partition(t(), keyword()) :: t()
+  defdelegate partition(stream, options \\ []), to: Operators
+
+  @doc """
+  Reduces items in the stream using an accumulator.
+
+  ## Parameters
+
```
polvalente, line 368: We should also point to Flow for users to read about the restrictions on this.

lib/grpc/stream.ex
```diff
@@ -1,17 +1,433 @@
...
+    - `acc_fun` initializes the accumulator,
```
polvalente, line 369: why do we need a 0-arity function for acc? Is this just Flow leaking?

lib/grpc/stream.ex
```diff
@@ -1,17 +1,433 @@
...
+    - `reducer_fun` updates it for each input.
+  """
+  @spec reduce(t, (-> acc), (term, acc -> acc)) :: t when acc: term()
+  defdelegate reduce(stream, acc_fun, reducer_fun), to: Operators
+
+  @doc """
+  Applies the given function rejecting each input in parallel.
+  """
+  @spec reject(t, (term -> term)) :: t
+  defdelegate reject(stream, filter), to: Operators
```
polvalente, line 379: Let's not expose this for now to keep the API more concise

lib/grpc/stream.ex
```diff
@@ -1,17 +1,433 @@
...
+
+  @doc """
+  Emits only distinct items from the stream.
+  """
+  @spec uniq(t) :: t
+  defdelegate uniq(stream), to: Operators
```
polvalente, line 385: What happens if we partition first? IIRC there's some interplay there

lib/grpc/stream.ex
```diff
@@ -1,17 +1,433 @@
...
     headers = adapter.get_headers(stream.payload)
     GRPC.Transport.HTTP2.decode_headers(headers)
   end
+
+  defp build_grpc_stream(input, opts) do
+    dispatcher = Keyword.get(opts, :default_dispatcher, GenStage.DemandDispatcher)
+
+    flow =
+      case Keyword.get(opts, :join_with) do
+        pid when is_pid(pid) ->
```
polvalente, line 411: maybe also allow an atom and check that Process.whereis finds a process from it

lib/grpc/stream/operators.ex
```diff
@@ -0,0 +1,172 @@
...
+    end
+
+    %GRPCStream{stream | flow: Flow.map(flow, mapper)}
+  end
+
+  def ask(%GRPCStream{flow: flow} = stream, target, timeout) when is_atom(target) do
+    mapper = fn item ->
+      if function_exported?(target, :handle_call, 3) do
+        try do
+          case GenServer.call(target, {:request, item}, timeout) do
```
polvalente, line 40: As I suggested above, if we just `send` the message, the receive-after combo above will effectively do what GenServer.call does, and the GenServer can just implement `handle_info` to deal with this.

polvalente (commented):
Sorry about the lengthy review. This overall is looking good, but seeing that we're aiming for this to mark v1.0, I think we can commit to the breaking change of merging the GRPC.Stream and GRPC.Server.Stream structs, and remove other kinds of streaming, even if we have to do this through multiple PRs.

lib/grpc/stream.ex
```diff
@@ -1,17 +1,433 @@
...
+
+  ## Returns
+    - A `GRPCStream` that emits the single gRPC message under demand.
+
+  ## Example
+
+      flow = GRPCStream.single(request, max_demand: 5)
+  """
+  @spec single(any(), Keyword.t()) :: t()
+  def single(input, opts \\ []) when is_struct(input),
```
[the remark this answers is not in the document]
sleipnir: For two reasons:
1. It is possible to create a stream directly without relying on a unary request (see the `replace` operator)
2. It simply follows semantics already seen in other stream frameworks out there now.

lib/grpc/stream.ex
```diff
@@ -1,17 +1,433 @@
...
-  Some useful operations for streams.
+  Provides a `Flow`-based abstraction layer for building gRPC streaming pipelines in Elixir.
+
+  This module allows you to consume gRPC request streams as `Flow` pipelines with support for
+  backpressure via GenStage. You can also produce gRPC responses by materializing a `Flow`
+  back into the gRPC stream.
+
+  ## Capabilities
+
+  - Transforms an incoming gRPC request stream into a `Flow` with backpressure.
```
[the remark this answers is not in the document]
sleipnir: I think it will become clearer later, but basically it was a better abstraction to build on top of and it already provides the closest semantics I imagined. In addition to providing the binds for GenStage, Stream, Enumerables, without me having to code many of these binds manually.

lib/grpc/stream.ex
```diff
@@ -1,17 +1,433 @@
...
+      flow = GRPCStream.single(request, max_demand: 5)
+  """
+  @spec single(any(), Keyword.t()) :: t()
+  def single(input, opts \\ []) when is_struct(input),
+    do: build_grpc_stream([input], Keyword.merge(opts, unary: true))
+
+  @doc """
+  Converts a single gRPC request into a `Flow` pipeline while extracting and propagating context metadata from the `GRPC.Server.Stream`.
+
+  This is useful for unary gRPC calls that require metadata propagation within a Flow-based pipeline.
```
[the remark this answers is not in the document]
sleipnir: It makes sense and I followed the suggestion here and also for the function

lib/grpc/stream.ex
```diff
@@ -1,17 +1,433 @@
...
+
+    - `flow`: A valid `Flow` pipeline.
+    - `opts`: Optional keyword options.
+
+  ## Returns
+
+    A `%GRPC.Stream{}` struct wrapping the flow.
+  """
+  @spec from_flow!(Flow.t(), Keyword.t()) :: t()
+  def from_flow!(%Flow{} = flow, opts \\ []) do
```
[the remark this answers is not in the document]
sleipnir: I don't implement 100% of Flow's functions, only the ones I felt necessary at the time. The motivation for this is that the user may want to do more advanced things. If you think it's better, we can remove them.

lib/grpc/stream.ex
```diff
@@ -1,17 +1,433 @@
...
+  Raises an `ArgumentError` if the `flow` has not been initialized.
+
+  ## Returns
+
+    The internal `Flow` struct.
+  """
+  @spec to_flow!(t()) :: Flow.t()
+  def to_flow!(%__MODULE__{flow: nil}) do
+    raise ArgumentError, "GRPC.Stream has no flow, initialize it with from/2"
+  end
```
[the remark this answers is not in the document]
sleipnir: Ok. Just remember that the user does not access this directly.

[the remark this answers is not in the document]
sleipnir: Done!

lib/grpc/stream.ex
sleipnir: I agree, but can we do this interactively? I thought about just renaming it to Materializer because it would involve changing fewer parts for now. Also because semantically Materializer is a common name for this kind of thing, see for example https://doc.akka.io/libraries/akka-core/current/stream/operators/Source-or-Flow/fromMaterializer.html (and I know it might seem OO at first, but it still makes more sense to me at this point than the alternative of changing everything).
Wdyt?

lib/grpc/stream.ex
sleipnir: I don't think so because this stream shouldn't be used for anything else after it has been materialized.

lib/grpc/stream.ex
sleipnir: Solved in my last commit

lib/grpc/stream.ex
sleipnir: I removed the polymorphic return

lib/grpc/stream.ex
sleipnir: A server-side request means that after a client sends a single message a finite or infinite set of messages will be returned by the server to the client. This function allows this to be achieved in a more idiomatic way and with fewer steps for the library user. See the example in examples/helloworld_streams for more clarity.

lib/grpc/stream.ex
sleipnir: so that the user can have access to the connection metadata/headers and make decisions with this if necessary

lib/grpc/stream.ex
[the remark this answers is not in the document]
polvalente: I still think we should call it unary because  that's exactly what you're doing - creating a stream with unary: true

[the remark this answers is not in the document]
polvalente: The semantics discussed in other places also references unary instead of single

lib/grpc/stream.ex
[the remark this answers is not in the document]
polvalente: I think we should remove this. My point is exactly that it's not _all_ Flows that work here

lib/grpc/stream.ex
polvalente: Yeah, it's fine to do this in separate PRs!

lib/grpc/stream.ex
polvalente: The execution can't change anything in the struct that would be meaningful? If not, let's just return :ok

lib/grpc/stream.ex
polvalente: After reading the example, I think we should just remove this function. It is doing too much at the same time. The same results can be obtained by:

- Getting the result from the stream which you want to replace (your examples are both unary, which indicate you only want to deal with one result).
- Likewise, get the metadata and options from the same result
- Create a new Flow and GRPC.Stream as you would create one from scratch.

lib/grpc/stream.ex
polvalente: Ok. Let's just use map_with_context instead

lib/grpc/stream.ex
sleipnir: It was mapped 1:1 from `Flow` api but this is perfectly normal for any flow engine. This exists in Flow, in Akka, in Python, in Rust frameworks. I see no harm in supporting it.

polvalente: I think for all functions that have this kind of Flow abstraction "leaking" we should have something like: see Flow.<function> for more information on <argument>

polvalente: Just lean onto Flow directly

lib/grpc/stream.ex
sleipnir: Yes, it needs to be used inside a window otherwise it can cause problems.

sleipnir: I added a link to the documentation.

lib/grpc/stream.ex
sleipnir: I prefer to work only with processes, we can use handle_info if necessary

lib/grpc/stream.ex
sleipnir: I renamed the function to `via` and improved the doc

lib/grpc/stream.ex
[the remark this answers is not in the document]
sleipnir: I'm going to do it but for some reason I can't agree 😄

lib/grpc/stream.ex
sleipnir: Docs updates!

lib/grpc/stream.ex
sleipnir: Done!

lib/grpc/stream.ex
sleipnir: Let's deal with this in a second PR ok?!

lib/grpc/stream.ex
sleipnir: I'm going to kindly ask you to keep this one. I really want to avoid a lot of unnecessary steps and make life easier for the user. Can you keep it?

sleipnir: > After reading the example, I think we should just remove this function. It is doing too much at the same time. The same results can be obtained by:
> 
> * Getting the result from the stream which you want to replace (your examples are both unary, which indicate you only want to deal with one result).
> * Likewise, get the metadata and options from the same result
> * Create a new Flow and GRPC.Stream as you would create one from scratch.

Ok. I read it again and removed the function. I also updated the example.

sleipnir:
@polvalente shall we continue?

lib/grpc/stream.ex
```diff
@@ -64,12 +64,15 @@ defmodule GRPC.Stream do
   ## Parameters
 
     - `input`: A gRPC request stream (struct, enumerable, or Elixir `Stream`).
-    - `opts`: Optional keyword list to configure the Flow.
 
   ## Options
 
     - `:join_with` — An optional external `GenStage` producer to merge with the gRPC input.
     - `:dispatcher` — Specifies the `Flow` dispatcher (defaults to `GenStage.DemandDispatcher`).
+    - `:propagate_context` - If `true`, the context from the `materializer` is propagated to the Flow.
```
polvalente, line 72:
```suggestion
    - `:propagate_context` - If `true`, the context from the `materializer` is propagated to the `Flow`.
```

lib/grpc/stream.ex
```diff
@@ -88,49 +91,22 @@ defmodule GRPC.Stream do
 
   def from(input, opts) when not is_nil(input), do: from([input], opts)
```
polvalente, line 92: Thinking if we should use `new` instead of `from` here...

lib/grpc/stream.ex
```diff
@@ -88,49 +91,22 @@ defmodule GRPC.Stream do
...
 
     - `input`: The single gRPC message to convert into a Flow.
-    - `opts`: Optional keyword list for configuring the Flow or GenStage producer.
 
   ## Options
-    - `:join_with` - (optional) An additional producer stage PID to include in the Flow.
-    - `:dispatcher` - (optional) The dispatcher to use for the Flow. Default is `GenStage.DemandDispatcher`.
+
+    - `:join_with` - An optional additional producer stage PID to include in the Flow.
+    - `:dispatcher` - An optional GenStage dispatcher to use for the Flow. Default is `GenStage.DemandDispatcher`.
```
polvalente, line 105:
```suggestion
    - `:dispatcher` - An optional `GenStage` dispatcher to use in the underlying `Flow`. Defaults to `GenStage.DemandDispatcher`.
```

lib/grpc/stream.ex
```diff
@@ -88,49 +91,22 @@ defmodule GRPC.Stream do
...
+    - `:propagate_context` - If `true`, the context from the `materializer` is propagated to the Flow.
```
polvalente, line 106:
```suggestion
    - `:propagate_context` - If `true`, the context from the `materializer` is propagated to the `Flow`.
```

lib/grpc/stream.ex
```diff
@@ -139,121 +115,89 @@ defmodule GRPC.Stream do
...
-  This is useful for unary gRPC calls that require metadata propagation within a Flow-based pipeline.
-
-  ## Parameters
-
-    - `input`: The enumerable stream of incoming gRPC messages.
-    - `materializer`: A `%GRPC.Server.Stream{}` struct representing the current gRPC stream context.
-    - `opts`: Optional keyword list for configuring the Flow or GenStage producer.
-
-  ## Behavior
+  Extracts the internal `Flow` pipeline from a `GRPC.Stream`.
```
polvalente, line 123:
```suggestion
  Extracts the underlying `Flow` pipeline from a `GRPC.Stream`.
```

lib/grpc/stream.ex
```diff
@@ -139,121 +115,89 @@ defmodule GRPC.Stream do
...
 
-    - Automatically extracts gRPC metadata from the stream headers and injects it into the options as `:metadata`.
+  Raises an `ArgumentError` if the `flow` has not been initialized.
```
polvalente, line 125:
```suggestion
  Raises an `ArgumentError` if the `Flow` has not been initialized.
```

lib/grpc/stream.ex
```diff
@@ -139,121 +115,89 @@ defmodule GRPC.Stream do
...
 
   ## Returns
 
-    - A `GRPC.Stream` that emits the single gRPC message under demand.
-
-  ## Example
-
-      flow = GRPC.Stream.single_as_ctx(request, stream, max_demand: 10)
-
+    The internal `Flow` struct.
```
polvalente, line 129:
```suggestion
    A `Flow` pipeline.
```

lib/grpc/stream.ex
```diff
@@ -139,121 +115,89 @@ defmodule GRPC.Stream do
...
+  @spec to_flow(t()) :: Flow.t()
+  def to_flow(%__MODULE__{flow: flow}) when is_nil(flow), do: Flow.from_enumerable([])
 
-  @doc """
-  Wraps an existing `Flow` into a `GRPC.Stream` struct.
+  def to_flow(%__MODULE__{flow: flow}), do: flow
 
-  > Note: `:join_with` is not supported in this mode and will raise an error if passed.
+  @doc """
+  Executes the flow for unary/single streams and emits responses into the provided gRPC server stream.
```
polvalente, line 137:
```suggestion
  Executes the underlying `Flow` for unary/single streams and emits responses into the provided gRPC server stream.
```

lib/grpc/stream.ex
```diff
@@ -139,121 +115,89 @@ defmodule GRPC.Stream do
...
 
   ## Parameters
 
-    - `flow`: A valid `Flow` pipeline.
-    - `opts`: Optional keyword options.
+    - `flow`: A `GRPC.Stream` struct containing the flow to be executed.
+    - `stream`: A `GRPC.Server.Stream` to which responses are sent.
+    - `:dry_run` — If `true`, responses are not sent (used for testing or inspection).
```
polvalente, line 143: Not sure we should introduce options just for testing. Is this really useful for users?

lib/grpc/stream.ex
```diff
@@ -139,121 +115,89 @@ defmodule GRPC.Stream do
...
       GRPC.Stream.run_with(request, mat)
   """
-  @spec run_with(t(), Stream.t(), Keyword.t()) :: :ok | any()
+  @spec run_with(t(), Stream.t(), Keyword.t()) :: :ok
   def run_with(
         %__MODULE__{flow: flow, options: flow_opts} = _stream,
         %Materializer{} = from,
         opts \\ []
       ) do
+    unless Keyword.get(flow_opts, :unary, true) do
```
polvalente, line 189:
```suggestion
    if not Keyword.get(flow_opts, :unary, true) do
```

lib/grpc/stream.ex
```diff
@@ -139,121 +115,89 @@ defmodule GRPC.Stream do
...
 
-    if unary? do
-      flow
-      |> Enum.to_list()
-      |> List.flatten()
-      |> List.first()
-    else
+    flow
+    |> Flow.map(fn msg ->
+      unless dry_run?, do: send_response(from, msg)
```
polvalente, line 197:
```suggestion
      if not dry_run? do 
        send_response(from, msg)
      end
```

lib/grpc/stream.ex
polvalente: unless is deprecated in newer elixir

polvalente (approved):
Just some minor comments and questions, but this looks mostly good to go (given the pending stuff we're leaving for follow-up PRs)

lib/grpc/stream.ex
sleipnir: I preferred to use the source sink semantics very common in Stream frameworks, I prefer not to touch this unless you really want to touch it.

lib/grpc/stream.ex
sleipnir: I think so because it allows users to test their own workflows.

sleipnir:
> Just some minor comments and questions, but this looks mostly good to go (given the pending stuff we're leaving for follow-up PRs)

Great! I accepted all suggestions except two so we could discuss them further.

lib/grpc/stream.ex
sleipnir: Totally get your point — appreciate the thoughtful feedback!

You're right that from isn’t a super common standalone function in Elixir, and it doesn’t follow the typical from_<type> or into patterns. That said, I’m gonna stick with from for now. In gRPC and streaming systems, from is often used to express the origin of a stream — like from(client) or from(server) — and that’s the kind of mental model I’m trying to bring in here.

And hey, I’m also not exactly known for strictly following language conventions 😄 — I like borrowing semantics from other ecosystems when they help convey intent more clearly.

But really appreciate you pointing it out — it’s a good perspective and definitely something to keep in mind moving forward.

polvalente: Yeah, this is why I'm not blocking this outright. Just given the patterns we have in the Elixir ecosystem, it might be a bit strange for Elixir users (even if this might be more straightforward for grpc users)

sleipnir: Ok, as we said we can change this in the future. I truly believe that good documentation makes anyone understand concepts and intentions, so we will have to document this properly. But if people complain we will change it.

PR #419, opened 2025-05-09
Pass nodelay as a TCP option instead of a TLS option

v0idpwn:
Gun passes tcp options in `gen_tcp:connect`, and tls options in `ssh:connect/2`. `ssh:connect/2` ignores the `nodelay` option.

This patch ensures that when using `ssl`, the client still pass `nodelay` in `tcp_options`, which makes sure it's used.


polvalente (approved):

sleipnir (approved):

v0idpwn:
Folks, I'd **love** if this could get released. I see there's no more changes after v0.10, so perhaps could make a patch release :)

sleipnir:
> Folks, I'd **love** if this could get released. I see there's no more changes after v0.10, so perhaps it could make a patch release :)

I think we can do one of two things:

* Release `v0.10.1` now
* Wait for [this PR](https://github.com/elixir-grpc/grpc/pull/418) and release v1.0.x

Wdyt guys?

v0idpwn:
I'm biased for the 0.10.1 release, but I think this can benefit people now, and I assume 1.0 will probably take some time between RCs and ensuring the new APIs work, so (if I have a right to vote) I'd vote for a quick 0.10.1 :)

polvalente:
+1 for 0.10.1!

sleipnir:
Ok. I will prepare the version!

sleipnir:
Release published https://hexdocs.pm/grpc/0.10.1/readme.html
Thanks again @v0idpwn

PR #431, opened 2025-08-04
fix: passing the adapter_opts properly

XinGuanTubi:
The `adapter_opts` keyword list encapsulates the :ip and :net options.

### DIFF
--- a/lib/grpc/server/adapters/cowboy.ex
+++ b/lib/grpc/server/adapters/cowboy.ex
@@ -31,7 +31,7 @@ defmodule GRPC.Server.Adapters.Cowboy do
   @impl true
   def start(endpoint, servers, port, opts) do
     start_args = cowboy_start_args(endpoint, servers, port, opts)
-    start_func = if opts[:cred], do: :start_tls, else: :start_clear
+    start_func = if cred_opts(opts), do: :start_tls, else: :start_clear
 
     case apply(:cowboy, start_func, start_args) do
       {:ok, pid} ->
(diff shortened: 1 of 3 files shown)


lib/grpc/server/adapters/cowboy.ex
```diff
@@ -270,8 +272,10 @@ defmodule GRPC.Server.Adapters.Cowboy do
         _, acc -> acc
       end)
 
-    if opts[:cred] do
-      opts[:cred].ssl ++
+    cred_opts = cred_opts(opts)
+
+    if cred_opts do
+      cred_opts.ssl ++
```
sleipnir, line 280: This may fail if misconfigured by the user, I suggest performing an extra validation and throwing an error if this key does not exist.

lib/grpc/server/adapters/cowboy.ex
```diff
@@ -285,6 +289,10 @@ defmodule GRPC.Server.Adapters.Cowboy do
     end
   end
 
+  defp cred_opts(opts) do
+    Kernel.get_in(opts, [:cred]) || Kernel.get_in(opts, [:adapter_opts, :cred])
```
sleipnir: Since we don't allow passing other keys for options, wouldn't it be better to just search in adapter_opts ?

sleipnir (changes requested):
Just a few minor comments and questions

lib/grpc/server/adapters/cowboy.ex
xinguanz: I added a `validate_cred` in server/supervisor.ex to make sure the `cred` and `ssl` are in correct format.

lib/grpc/server/adapters/cowboy.ex
xinguanz: Updated. I added additional retrieval conditions because the `cred_opts` was called from both `start` and `child_spec`. Since we already receive `merged_adapter_opts` as the third element in the result from `cowboy_start_args`, I believe we can safely change `cred_opts`.

sleipnir (approved):
LGTM

XinGuanTubi:
@sleipnir I've formatted the file

PR #434, opened 2025-08-24
Always set CORS headers

aseigo:
We have discovered that not all gRPC-web sessions advertise CORS in their headers, but they are still limited by CORS by the browser itself. This causes those non-advertising gRPC requests to fail unexpectedly. It is possible to trigger this with the official upstream JS library, even, depending on the configuration.

So this PR cuases the interceptor to always attempt to set CORS headers. In practice, this is still setting the headers just as often, as GRPC sessions that do advertise CORS in the request headers do so consistently, so the performance impact is essentially nil.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,2 +1,0 @@
-elixir 1.15.7-otp-26
-erlang 26.1.2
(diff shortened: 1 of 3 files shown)


sleipnir (approved):

sleipnir:
Tks!

Issue #435, opened 2025-09-08
Transcoding HTTP request fails if accept header is "*/*


the-destro:
**Describe the bug**
When exposing an endpoint via transcoding the accept header seems to require "application/json". This makes it a challenge when you are, say, adding a health check route and you don't have control over what the service sends as a request header.

**To Reproduce**

a minimal service definition
```
message HealthCheckRequest {
}

message HealthCheckResponse {
}

service Health {
  rpc Check(HealthCheckRequest) returns (HealthCheckResponse){
    option (google.api.http) = {
      get: "/_health"
    };
  }
}
```

```
2025-09-08T15:25:42.607 [error] ** (GRPC.Server.Adapters.ReportException) Exception raised while handling /_health:
** (GRPC.RPCError) Message is malformed.
```

sleipnir:
> **Describe the bug** When exposing an endpoint via transcoding the accept header seems to require "application/json". This makes it a challenge when you are, say, adding a health check route and you don't have control over what the service sends as a request header.
> 
> **To Reproduce**
> 
> a minimal service definition
> 
> ```
> message HealthCheckRequest {
> }
> 
> message HealthCheckResponse {
> }
> 
> service Health {
>   rpc Check(HealthCheckRequest) returns (HealthCheckResponse){
>     option (google.api.http) = {
>       get: "/_health"
>     };
>   }
> }
> ```
> 
> ```
> 2025-09-08T15:25:42.607 [error] ** (GRPC.Server.Adapters.ReportException) Exception raised while handling /_health:
> ** (GRPC.RPCError) Message is malformed.
> ```

Hello, how are you?
The error doesn't seem to be related to what you said. How did you arrive at that conclusion about the Accept header?

PR #436, opened 2025-09-09
[Chore]: Fallback to json if accept header not defined

sleipnir:


sleipnir:
This closes https://github.com/elixir-grpc/grpc/issues/435

Makefile
```diff
@@ -1,7 +1,7 @@
 LOCAL_CLEAN=$(shell git diff-index --quiet HEAD -- && echo 1)
 
 test:
-	mix test
+	mix deps.get && mix test
```
polvalente:
```suggestion
	@mix deps.get
      mix test
```

lib/grpc/server/adapters/cowboy/handler.ex
```diff
@@ -136,11 +132,32 @@ defmodule GRPC.Server.Adapters.Cowboy.Handler do
...
+
     access_mode = resolve_access_mode(req, access_mode, subtype)
     {:ok, access_mode, subtype, content_type}
   end
 
+  defp extract_http_method(req) do
+    req
+    |> :cowboy_req.method()
+    |> String.downcase()
+    |> String.to_existing_atom()
```
polvalente: I think we could get away with just the downcase string comparison to avoid unnecessary raises

polvalente (approved):

lib/grpc/server/adapters/cowboy/handler.ex
sleipnir: Done!

PR #438, opened 2025-09-11
[Feat] Client LB using gRPC specs 

sleipnir:


lib/grpc/client/resolver.ex
```diff
@@ -1,14 +1,82 @@
...
+
+    * `dns`, `ipv4`, `ipv6` → 50051
+    * `xds` → 443
+
+  ## Resolver Output
+
+  Returns:
+
+    * `{:ok, %{addresses: list(map()), service_config: GRPC.Client.ServiceConfig.t() | nil}}`
+      - `addresses` – list of endpoint maps
```
polvalente:
```suggestion
      - `addresses` – list of endpoint maps containing the keys:
```

lib/grpc/client/resolver.ex
```diff
@@ -0,0 +1,139 @@
...
+
+  @doc """
+  Initialization for resolvers.
+
+  Calls `resolver_module.start(args)` and returns its result.
+  """
+  @spec start_resolver(module(), args :: term()) :: :ok | {:ok, pid()} | {:error, term()}
+  def start_resolver(resolver, args \\ nil) when is_atom(resolver) do
+    resolver.start(args)
+  end
```
polvalente: not sure this function is needed

sleipnir: This PR is still a draft, but the idea is that certain resolvers may need specific initialization, for example when their backend is a GenServer, as is the case with the DNS resolver, which is why this function exists.

polvalente: It's just that one could easily just call `resolver.start(args)` directly

sleipnir: > It's just that one could easily just call `resolver.start(args)` directly

I will need to interact a little more to know if this is really a fact, I will bring you news soon about this

benchmark/lib/grpc/core/stats.pb.ex
```diff
@@ -1,7 +1,7 @@
 defmodule Grpc.Core.Bucket do
```
polvalente, line 1: Are these changes intentional?

sleipnir: I believe this was already like this, I don't know when or why these files were originally created.

lib/grpc/client/load_balacing/round_robin.ex
```diff
@@ -0,0 +1,24 @@
...
+  @behaviour GRPC.Client.LoadBalancing
+
+  @impl true
+  def init(opts) do
+    addresses = Keyword.get(opts, :addresses, [])
+
+    if addresses == [] do
+      {:error, :no_addresses}
+    else
+      {:ok, %{addresses: addresses, index: 0}}
```
polvalente:
```suggestion
      {:ok, %{addresses: addresses, index: 0, n: length(addresses)}}
```

lib/grpc/client/load_balacing/round_robin.ex
```diff
@@ -0,0 +1,24 @@
...
+
+  @impl true
+  def pick(%{addresses: addresses, index: idx} = state) do
+    count = length(addresses)
+    next_idx = rem(idx, count)
+    %{address: host, port: port} = Enum.at(addresses, next_idx)
+
+    new_state = %{state | index: idx + 1}
+    {:ok, {host, port}, new_state}
+  end
```
polvalente:
```suggestion
  def pick(%{addresses: addresses, index: idx, n: n} = state) do
    %{address: host, port: port} = Enum.fetch!(addresses, idx)

    new_state = %{state | index: rem(idx + 1, n)}
    {:ok, {host, port}, new_state}
  end
```

lib/grpc/client/resolver/dns/adapter.ex
```diff
@@ -0,0 +1,12 @@
...
+  @moduledoc """
+  Adapter to resolve DNS (A and TXT).
+  """
+
+  @callback lookup(charlist(), :a | :txt) ::
+              {:ok, [tuple() | String.t()]} | {:error, term()}
+
+  def lookup(name, type) do
+    :inet_res.lookup(name, :in, type)
+  end
```
polvalente: Not sure I understand why there's an adapter definining a callback

lib/grpc/client/resolver/dns.ex
```diff
@@ -0,0 +1,55 @@
...
+  """
+  @behaviour GRPC.Client.Resolver
+
+  alias GRPC.Client.ServiceConfig
+
+  @impl GRPC.Client.Resolver
+  def resolve(target) do
+    uri = URI.parse(target)
+    host = uri.host || target
+    port = uri.port || 50051
```
polvalente, line 19: Why 50051? Should this be configurable?

lib/grpc/client/resolver/dns.ex
```diff
@@ -0,0 +1,55 @@
...
+    host_charlist = String.to_charlist(host)
```
polvalente: Shouldn't the lookup accept strings and specific adapters convert to charlists as needed instead? Charlists are not very elixir-y IMO

lib/grpc/client/resolver/dns.ex
```diff
@@ -0,0 +1,55 @@
...
+
+        case adapter().lookup(~c"_grpc_config." ++ host_charlist, :txt) do
+          {:ok, txt_records} ->
+            service_config_json =
+              txt_records
+              |> Enum.map(&List.to_string/1)
+              |> Enum.find_value(fn str ->
+                case String.split(str, "grpc_config=") do
+                  [_, json] -> json
+                  _ -> nil
```
polvalente: I think this would be more readable as a with and each lookup as a separate defp.

lib/grpc/client/resolver/dns.ex
```diff
@@ -0,0 +1,55 @@
...
+          Enum.map(addresses, fn ip ->
+            %{address: ip |> :inet.ntoa() |> to_string(), port: port}
+          end)
+
+        case adapter().lookup(~c"_grpc_config." ++ host_charlist, :txt) do
+          {:ok, txt_records} ->
+            service_config_json =
+              txt_records
+              |> Enum.map(&List.to_string/1)
+              |> Enum.find_value(fn str ->
```
polvalente: No need to map if you're doing find_value

lib/grpc/client/resolver/dns/adapter.ex
sleipnir: No problem, I explain. 
The idea was that we could have other implementations for name resolution, but it provides a simple default implementation. This is using no dns.ex

lib/grpc/client/resolver/ipv4.ex
```diff
@@ -0,0 +1,55 @@
...
+
+  @impl GRPC.Client.Resolver
+  def resolve(target) do
+    uri = URI.parse(target)
+    addrs_str = uri.path
+
+    addresses =
+      String.split(addrs_str, ",")
+      |> Enum.map(fn addr ->
+        [ip, port] = String.split(addr, ":")
```
polvalente:
```suggestion
      String.split(addrs_str, ",", trim: true)
      |> Enum.map(fn addr ->
        [ip, port] = String.split(addr, ":", trim: true, parts: 2)
```

lib/grpc/client/resolver/dns.ex
sleipnir: This is the commonly used port for gRpcs but of course, this should be configurable in the final version.

lib/grpc/client/resolver/dns.ex
sleipnir: Done!

lib/grpc/client/resolver/ipv6.ex
```diff
@@ -0,0 +1,65 @@
...
+        case Regex.run(~r/\[(?<ip>.*?)\](?::(?<port>\d+))?/, entry) do
+          [_, ip, port] ->
+            %{address: ip, port: String.to_integer(port)}
+
+          [_, ip] ->
+            %{address: ip, port: 443}
+
+          _ ->
+            {:error, :invalid_ipv6}
+        end
```
polvalente:
```suggestion
        case Regex.run(~r/\[([^\])+)\](?:(\d+))?/, entry, capture: :all_but_first) do
          [ip, port] ->
            %{address: ip, port: String.to_integer(port)}

          [ip] ->
            %{address: ip, port: 443}

          _ ->
            {:error, :invalid_ipv6}
        end
```

I've changed the regex so it can fail faster with the `[^\[]` matcher. The other one was a bit too greedy.

However, we might want to consider a different approach here, where we use :inet to validate the address part. In the code below, if port is not provided, it'll be empty string. Otherwise, do `Integer.parse` after removing the `:` prefix.

```elixir
"[" <> str = entry 
[addr, port] = String.split(str, "]", parts: 2)
case :inet.parse_address(String.to_charlist(addr)) do
  {:ok, _} -> ...
  ...
end
```

lib/grpc/client/conn.ex
```diff
@@ -0,0 +1,507 @@
+defmodule GRPC.Client.Conn do
```
polvalente:
```suggestion
defmodule GRPC.Client.Connection do
```

I think Conn is particularly overloaded with Plug, and using a different name will be less confusing, especially with this being a Process and not a struct.

lib/grpc/client/connection.ex
```diff
@@ -0,0 +1,507 @@
...
+    ch = initial_state.virtual_channel
+
+    case GenServer.whereis(via(ref)) do
+      nil ->
+        # start the orchestration server, register by name
+        case GenServer.start_link(__MODULE__, initial_state, name: via(ref)) do
+          {:ok, _pid} ->
+            # only now persist the chosen channel (which should already have adapter_payload
+            # because build_initial_state connected real channels and set virtual_channel)
+            :persistent_term.put({__MODULE__, :lb_state, ref}, ch)
```
polvalente: I don't understand how this works. If we're using via(ref), why is the second call different?

polvalente: Misread here. Ignore.

polvalente: Shouldn't this be put in start_link itself?

lib/grpc/client/resolver/dns.ex
sleipnir: done

lib/grpc/client/connection.ex
```diff
@@ -0,0 +1,507 @@
...
+            {:ok, ch}
+
+          {:error, {:already_started, pid}} ->
+            # race: someone else started it first, ask the running process for its current channel
+            if is_pid(pid) do
+              {:ok, ch}
+            end
```
polvalente: `pid` should always be a pid

lib/grpc/client/resolver/dns.ex
sleipnir: Done!

lib/grpc/client/conn.ex
```diff
@@ -0,0 +1,507 @@
...
+
+      pid when is_pid(pid) ->
+        case pick(opts) do
+          {:ok, %Channel{} = channel} ->
+            {:ok, channel}
+
+          _ ->
+            {:error, :no_connection}
+        end
+    end
```
polvalente: I think we should just rely on start_link to either return the pid or the already_started case.

lib/grpc/client/conn.ex
```diff
@@ -0,0 +1,507 @@
...
+    * `{:error, :no_connection}` – if the orchestrator is not available
+
+  ## Example
+
+      iex> {:ok, ch} = GRPC.Client.Conn.connect("dns://my-service.local:50051")
+      iex> GRPC.Client.Conn.pick(ch)
+      {:ok, %GRPC.Channel{host: "192.168.1.1", port: 50051}}
+  """
+  @spec pick(Channel.t(), keyword()) :: {:ok, Channel.t()} | {:error, term()}
+  def pick(%Channel{ref: ref} = _channel, _opts \\ []) do
```
polvalente:
```suggestion
  def pick_channel(%Channel{ref: ref} = _channel, _opts \\ []) do
```

I think pick is not clear on whether you're picking the channel or from the channel

lib/grpc/client/connection.ex
```diff
@@ -0,0 +1,507 @@
...
+        {:ok, channel}
+    end
+  end
+
+  @impl true
+  def init(%__MODULE__{} = state) do
+    Process.flag(:trap_exit, true)
+    Process.send_after(self(), :refresh, @refresh_interval)
+    {:ok, state}
+  end
```
polvalente: IMO init should be close to start_link, both at the top of the module, followed by public apis and then handle_* callbacks

lib/grpc/client/connection.ex
```diff
@@ -0,0 +1,507 @@
...
+      Enum.map(state.real_channels, fn {_key, ch} ->
+        adapter.disconnect(ch)
+      end)
+
+      keys_to_delete = [:real_channels, :virtual_channel]
+
+      new_state =
+        Enum.reduce(keys_to_delete, state, fn key, acc ->
+          if Map.has_key?(acc, key), do: Map.delete(acc, key), else: acc
+        end)
```
polvalente:
```suggestion
      keys_to_delete = [:real_channels, :virtual_channel]
      new_state = Map.drop(state, keys_to_delete)
```

lib/grpc/client/connection.ex
```diff
@@ -0,0 +1,507 @@
...
+      virtual_channel: virtual_channel,
+      resolver: resolver,
+      adapter: adapter
+    }
+
+    case resolver.resolve(norm_target) do
+      {:ok, %{addresses: addresses, service_config: config}} ->
+        lb_policy =
+          cond do
+            is_map(config) && Map.has_key?(config, :load_balancing_policy) ->
```
polvalente:
```suggestion
            is_map(config) and Map.has_key?(config, :load_balancing_policy) ->
```

lib/grpc/client/connection.ex
```diff
@@ -0,0 +1,507 @@
...
+    |> adapter.connect(opts[:adapter_opts])
+  end
+
+  defp connect_real_channel(vc, host, port, opts, adapter) do
+    %Channel{vc | host: host, port: port}
+    |> adapter.connect(opts[:adapter_opts])
+  end
+
+  defp split_host_port(target) do
+    case String.split(target, ":") do
```
polvalente:
```suggestion
    case String.split(target, ":", trim: true) do
```

lib/grpc/client/service_config.ex
```diff
@@ -0,0 +1,102 @@
...
+
+  def parse(json) when is_binary(json) do
+    case Jason.decode(json) do
+      {:ok, map} -> from_map(map)
+      error -> error
+    end
+  end
+
+  defp from_map(map) do
+    # Extrai a primeira política de load balancing do array loadBalancingConfig
```
polvalente: translate to english please

lib/grpc/client/service_config.ex
```diff
@@ -0,0 +1,102 @@
...
+    lb_policy =
+      map
+      |> Map.get("loadBalancingConfig", [%{"pick_first" => %{}}])
+      |> List.first()
+      |> Map.keys()
+      |> List.first()
```
polvalente: What is this doing?

polvalente: Map.keys does not ensure ordering

lib/grpc/client/resolver/dns/adapter.ex
polvalente: I see. I think this should be a bit different. This module would have:

```elixir
def lookup(name, type), do: adapter().lookup(name, type)

defp adapter, do: Application.get_env(:grpc, __MODULE__, InetRes)
```

and then a separate module, such as DNS.Adapter.InetRes would have this default implementation.

polvalente: in short, we can have functions here for common access, but they should all delegate to the adapter as a pattern

benchmark/lib/grpc/core/stats.pb.ex
polvalente: I mean the downgrade in the version. It seems that when you re-generated you used 14.0 instead of 14.1

lib/grpc/client/connection.ex
sleipnir: In that case we do not have start_link....

sleipnir: reordered

lib/grpc/client/service_config.ex
sleipnir: Resolved

lib/grpc/client/connection.ex
sleipnir: I don't know if I understand, we don't have the start_link function, we use genserver but we don't expose it as a genserver as usual, we only have the connect function. Can you explain better?

lib/grpc/client/conn.ex
sleipnir: I don't know if I understand, but I don't know if it's possible to do this without breaking the API

lib/grpc/client/conn.ex
polvalente: We should rename the file too

lib/grpc/client/conn.ex
polvalente: I mean that you're doing Process.whereis on the genserver and acting upon it. However, you're already dealing with this case implicitly by handling the `error already_started` case above.

lib/grpc/client/conn.ex
sleipnir: I had already done this
Captura de tela de 2025-09-26 14-02-26

lib/grpc/client/connection.ex
polvalente: Yeah, I re-read the module now and understand it better. I think I misread this originally and thought it was happening on the "already started" case

lib/grpc/client/conn.ex
sleipnir: Ok. I got it!

lib/grpc/client/resolver/dns.ex
```diff
@@ -17,38 +17,58 @@ defmodule GRPC.Client.Resolver.DNS do
...
+      {:error, reason} -> {:error, {:dns_error, reason}}
+    end
+  end
+
+  defp lookup_addresses(host) do
+    adapter().lookup(host, :a)
+  end
+
+  defp lookup_service_config(host) do
+    case adapter().lookup(~c"_grpc_config." ++ host, :txt) do
```
polvalente: I think this needs fixing (and tests). The default impl will break on String.to_charlist

sleipnir: Done!

lib/grpc/stub.ex
```diff
@@ -105,19 +105,27 @@ defmodule GRPC.Stub do
   end
 
   @doc """
-  Establish a connection with gRPC server and return `GRPC.Channel` needed for
-  sending requests.
+  Establishes a connection with a gRPC server and returns a `GRPC.Channel` required
+  for sending requests. Supports advanced connection resolution via the gRPC `Resolver`
+  and various target schemes (`dns`, `unix`, `xds`, `host:port`, etc).
```
polvalente, line 112:
```suggestion
  Establishes a connection with a gRPC server and returns a `GRPC.Channel` required
  for sending requests. 
  
  Supports advanced connection resolution via the gRPC `Resolver`
  and various target schemes (`dns`, `unix`, `xds`, `host:port`, etc).
```

First line of the doc should always be short.

README.md
```diff
@@ -133,7 +144,30 @@ def say_bid_stream_hello(request, materializer) do
   |> GRPC.Stream.run_with(materializer)
 end
 ```
-__💡__ The Stream API supports composable stream transformations via `ask`, `map`, `run` and others functions, enabling clean and declarative stream pipelines. For a complete list of available operators see [here](lib/grpc/stream.ex).
+The Stream API supports composable stream transformations via `ask`, `map`, `run` and others functions, enabling clean and declarative stream pipelines. See the table below:
```
polvalente: This table should be placed in the moduledoc of GRPC.Stream and we should point the user there instead.
The function references should be in the ex_doc style. I'm gonna apply these changes.

PR #442, opened 2025-10-08
feat: add error details support Signed-off-by: Yordis Prieto <yordis.prieto@gmail.com>

yordis:
Signed-off-by: Yordis Prieto <yordis.prieto@gmail.com>


lib/google/rpc/status.pb.ex
```diff
@@ -0,0 +1,9 @@
+defmodule Google.Rpc.Status do
```
yordis: Note, I need to figure out who will be actually compiling this module since the application layer may have a duplicate of this based on some codegen, pending to test.

In the worst case, I check an application environment to opt-out/in from compiling this into it.

It is weird that the gRPC spec "depends on" this Google RPC message that is external, unless I fundamentally misunderstood the assignment 😄

test/support/google/protobuf/descriptor.proto
```diff
@@ -1,1223 +0,0 @@
-// Protocol Buffers - Google's data interchange format
```
yordis, line 1: I do not think we need this file, do we?

lib/grpc/rpc_error.ex
```diff
@@ -96,4 +109,25 @@ defmodule GRPC.RPCError do
   def exception(status, message) when is_integer(status) do
     %GRPC.RPCError{status: status, message: message}
   end
+
+  @doc false
+  @spec from_grpc_details(GRPC.Status.t(), String.t() | nil, String.t() | nil) :: t()
```
yordis: I created this function because both adapters will call it, so I rather keep the same code path between them

lib/grpc/google/rpc.ex
```diff
@@ -0,0 +1,20 @@
+defmodule GRPC.Google.Rpc do
+  @moduledoc false
+
+  @spec encode_status(Google.Rpc.Status.t()) :: String.t()
+  def encode_status(%Google.Rpc.Status{} = status) do
```
yordis, line 5: I rather isolate these two function around encoding and decoding the rpc status, just internal module for now

lib/grpc/transport/http2.ex
```diff
@@ -141,7 +156,7 @@ defmodule GRPC.Transport.HTTP2 do
   end
 
   defp encode_metadata_pair({key, val}) do
-    val = if String.ends_with?(key, "-bin"), do: Base.encode64(val), else: val
+    val = if String.ends_with?(key, "-bin"), do: Base.encode64(val, padding: true), else: val
```
yordis, line 159: I made this explicit, I can roll it back

lib/grpc/rpc_error.ex
```diff
@@ -51,11 +51,15 @@ defmodule GRPC.RPCError do
...
   """
 
-  defexception [:status, :message]
+  defexception [:status, :message, :details]
 
   defguard is_rpc_error(e, status) when is_struct(e, __MODULE__) and e.status == status
 
-  @type t :: %__MODULE__{status: GRPC.Status.t(), message: String.t()}
+  @type t :: %__MODULE__{
+          status: GRPC.Status.t(),
```
yordis, line 59: Exactly, what GitHub Actions is saying, the `GRPC.Status.t()` is not there, that is because it requires the proper `protoc` compilation, this is related to the comment in the status module.

Or I remove this, so we do not use the type

lib/grpc/client/adapters/gun.ex
```diff
@@ -260,16 +277,7 @@ defmodule GRPC.Client.Adapters.Gun do
       {:response, :nofin, status, headers} ->
         if status == 200 do
           headers = GRPC.Transport.HTTP2.decode_headers(headers)
-
-          if headers["grpc-status"] && headers["grpc-status"] != "0" do
```
yordis, line 264: handle one layer above

lib/grpc/client/adapters/gun.ex
```diff
@@ -230,25 +265,7 @@ defmodule GRPC.Client.Adapters.Gun do
       {:response, :fin, status, headers} ->
         if status == 200 do
           headers = GRPC.Transport.HTTP2.decode_headers(headers)
-
-          case headers["grpc-status"] do
-            nil ->
```
yordis, line 235: handle one layer above

yordis:
@sleipnir I am still working on it, but if you want to leave early comments in case I missed the goal too much

test/grpc/integration/error_details_test.exs
```diff
@@ -0,0 +1,135 @@
+defmodule GRPC.Integration.ErrorDetailsTest do
```
yordis, line 1: I am following existing pattern, where I do not see the Mint adapter being test; I am not sure if I should introduce that, or maybe that should be a separate task

lib/google/rpc/status.pb.ex
sleipnir: I think the worst side effect here is redefining the module at runtime, which has no effect. Unless, of course, the user has redefined the module itself, which seems very unlikely.

yordis: @sleipnir the protobuff already comes with predefined compiled messages; I can create a package, get it to work, and then transfer under the organization so we can depend on it.

What do you think?

yordis: it could be under,

- beam-community - org
- elixir-protobuf - org 
- elixir-grpc - org
- yordis - personal

polvalente: Given these are protobuf specific, ideally they'd be inside the protobuf org

yordis: @polvalente I setup the repo already, either way it is fine by me, as long as we get the support of getting things in, and properly release.

polvalente: Ideally we validate this using git deps, and merge after the other repo is on hex, so we can release grpc right after with this in

yordis: Aye!

mix.exs
```diff
@@ -35,6 +35,7 @@ defmodule GRPC.Mixfile do
 
   defp deps do
     [
+      {:googleapis, github: "yordis/googleapis", branch: "main"},
```
yordis: Migrate and Release before merging

polvalente: if we're to keep this outside of the protobuf org, we should move it inside the grpc org

yordis: I can transfer to wherever

sleipnir:
> @sleipnir I am still working on it, but if you want to leave early comments in case I missed the goal too much

I haven't had time to look carefully at this material yet, but I'll check it out soon.

mix.exs
sleipnir: I agree with @polvalente we can absorb this within the org

lib/grpc/google/rpc.ex
sleipnir: I have my doubts about whether this should be here and not in the googleapis library itself. I think we could just import it here since the behavior seems to be universal.

lib/grpc/client/adapters/gun.ex
```diff
@@ -155,24 +155,74 @@ defmodule GRPC.Client.Adapters.Gun do
...
+      {:ok, response, %{headers: headers}}
+    else
+      {:ok, response}
+    end
+  end
+
+  defp handle_fin_response(headers, opts) do
+    # Trailers-only response: headers contain trailers, check status
+    case parse_trailers(headers) do
+      :ok ->
```
polvalente: we should either change the with on L187 or this case for uniformity

lib/grpc/google/rpc.ex
polvalente: Agreed. However, if we keep this module, I'd like to call it RPC instead of Rpc

lib/grpc/transport/http2.ex
```diff
@@ -21,12 +21,27 @@ defmodule GRPC.Transport.HTTP2 do
...
+
+  defp put_details_bin_grpc_status(trailers, status, message, details) when is_list(details) do
+    encoded_details =
+      GRPC.Google.Rpc.encode_status(%Google.Rpc.Status{
+        code: status,
+        message: message,
+        details: details
+      })
+
+    Map.put(trailers, "grpc-status-details-bin", encoded_details)
```
polvalente, line 44: are these name standardized? if not, I'd rather get rid of the -bin/_bin suffixes throughout

test/grpc/integration/error_details_test.exs
polvalente: If you think it's easy to test the mint adapter here, I'd rather have tests for it too

lib/grpc/transport/http2.ex
yordis: As far as I can tell, this is part of the spec https://github.com/grpc/grpc-go/blob/20c9f65ac25f1f14330e8403e4c101b5a20abbda/internal/transport/http_util.go#L88

Otherwise interoperability wouldn't work

test/grpc/integration/error_details_test.exs
yordis: @polvalente I do not disagree with you in terms of testing Mint; for this PR.

Could I merge it as is and then follow up on how to make Mint to run for all integration tests, no just this one?

lib/grpc/google/rpc.ex
yordis: @polvalente I can rename the repo that I wrote, I can not rename the `Google.Rpc` since that is codegen; I used the name intentionally like that to follow the codegen pkg name.

What do you y'all want to do? No strong opinions

polvalente: Yeah, I meant the GRPC.Google.Rpc -> GRPC.Google.RPC  rename only

test/grpc/integration/error_details_test.exs
polvalente: We can have them in a separate PR

mix.exs
```diff
@@ -35,6 +35,7 @@ defmodule GRPC.Mixfile do
 
   defp deps do
     [
+      {:googleapis, github: "elixir-grpc/googleapis", branch: "main"},
```
sleipnir:
```suggestion
      {:googleapis, "~> 0.1.0"},
```

lib/grpc/google/rpc.ex
```diff
@@ -0,0 +1,28 @@
...
+    e -> {:error, e}
+  end
+
+  defp decode64(str) when rem(byte_size(str), 4) == 0 do
+    Base.decode64!(str, padding: true)
+  end
+
+  defp decode64(str) do
+    Base.decode64!(str, padding: false)
+  end
```
yordis, line 27: this is following Go implementation:

https://github.com/grpc/grpc-go/blob/8110884be128afca96c21ca7ff15a490993f9a7b/internal/transport/http_util.go#L132-L138

PR #443, opened 2025-10-08
Refactor ETS table creation in tests to use unique names for parallel execution

yordis:

PR #444, opened 2025-10-09
fix: catch exits from crashing genserver in tests

polvalente:
closes #443

### DIFF
--- a/test/grpc/server/adapters/report_exception_test.exs
+++ b/test/grpc/server/adapters/report_exception_test.exs
@@ -60,13 +60,14 @@ defmodule GRPC.Server.Adapters.ReportExceptionTest do
       GenServer.cast(pid, :case_boom)
 
       assert_receive {:boom, {_reason, stack} = err}
+
       assert %GRPC.Server.Adapters.ReportException{
-              __exception__: true,
-              adapter_extra: [req: :ok],
-              kind: :error,
-              reason: %CaseClauseError{term: :ok},
-              stack: stack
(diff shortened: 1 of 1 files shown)


test/grpc/server/adapters/report_exception_test.exs
```diff
@@ -51,41 +51,65 @@ defmodule GRPC.Server.Adapters.ReportExceptionTest do
...
     end
 
     test "with badarg error" do
-      {:ok, pid} = GenServer.start_link(ExceptionServer, self())
-
+      test_pid = self()
+
+      %Task{ref: task_ref} = Task.async(fn ->
+        Process.flag(:trap_exit, true)
+        {:ok, pid} = GenServer.start_link(ExceptionServer, test_pid)
```
polvalente: @yordis @sleipnir the main problem is that the raises in handle_cast were propagating exits instead of normal crashes, which were then crashing the test PID due to the start_link

sleipnir: Thank you @polvalente

yordis: oooooooh thank you so much @polvalente sorry I wasted your time, honestly was trying to figure out what it was; TIL!

polvalente: Not a waste of time at all! This was a tricky one to figure out!

polvalente: Btw the final solution to the problem was much simpler than this Task setup

PR #446, opened 2025-10-11
feat: add support for tuple results

yordis:
Signed-off-by: Yordis Prieto <yordis.prieto@gmail.com>

### DIFF
--- a/lib/grpc/server.ex
+++ b/lib/grpc/server.ex
@@ -15,7 +15,7 @@ defmodule GRPC.Server do
         use GRPC.Server, service: Greeter.Service
 
         def say_hello(request, _stream) do
-          Reply.new(message: "Hello")
+          %Reply{message: "Hello"}
         end
 
         def say_goodbye(request_enum, stream) do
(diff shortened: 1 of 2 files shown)


lib/grpc/server.ex
```diff
@@ -30,6 +30,97 @@ defmodule GRPC.Server do
...
+  ## Unary Response Return Formats
+
+  For unary (non-streaming) responses, handlers can return values in multiple formats:
+
+      defmodule Greeter.Server do
+        use GRPC.Server, service: Greeter.Service
+
+        # Option 1: Return bare struct (existing behavior)
+        def say_hello(request, _stream) do
+          Reply.new(message: "Hello")
```
polvalente: I wonder if we should for the tuple result in a future version and give warnings for non-tuple results from this PR onwards

polvalente: Ideally in v+2 we hard deprecate the bare struct results

lib/grpc/server.ex
```diff
@@ -30,6 +30,97 @@ defmodule GRPC.Server do
...
+  ## Streaming Response Error Handling
+
+  For streaming responses, handlers can return `{:error, %GRPC.RPCError{}}` to abort the stream.
+  The handler can send zero or more messages via `send_reply/2` before returning the error:
+
+      defmodule RouteGuide.Server do
+        use GRPC.Server, service: RouteGuide.Service
+
+        # Server streaming: validate before sending any messages
+        def list_features(rect, stream) do
```
polvalente: What's rect?

lib/grpc/server.ex
```diff
@@ -30,6 +30,97 @@ defmodule GRPC.Server do
...
+          case validate_rect(rect) do
+            :ok ->
+              features = fetch_features()
+
+              Enum.each(features, fn feature ->
+                GRPC.Server.send_reply(stream, feature)
+              end)
+              # Implicit success - can return nil, :ok, etc.
```
polvalente: Can we enforce implicit :ok?

lib/grpc/server.ex
```diff
@@ -351,13 +442,23 @@ defmodule GRPC.Server do
...
+
+          {true, _} ->
+            # Streaming handlers may return nil, :ok, or any value for success
+            {:ok, stream}
+
+          {false, {:ok, response}} ->
+            {:ok, stream, response}
+
+          {false, response} ->
+            # Backward compatibility: support bare struct returns
```
polvalente, line 451: We can warn here

PR #450, opened 2025-10-14
[FIX] Virtual channel creation with non resolvable address

sleipnir:


sleipnir:
@polvalente I think now is ok 4 review

lib/grpc/client/connection.ex
```diff
@@ -367,63 +370,114 @@ defmodule GRPC.Client.Connection do
...
-              nil
-          end
+  defp validate_adapter_opts!(opts) when is_list(opts), do: :ok
 
-        lb_mod = choose_lb(lb_policy)
-        {:ok, lb_state} = lb_mod.init(addresses: addresses)
+  defp validate_adapter_opts!(_),
+    do: raise(ArgumentError, ":adapter_opts must be a keyword list if present")
 
+  defp build_compressor_list(compressor, accepted) do
```
polvalente:
```suggestion
  defp build_compressor_list(compressor, accepted) when is_list(accepted) do
```

lib/grpc/client/connection.ex
```diff
@@ -367,63 +370,114 @@ defmodule GRPC.Client.Connection do
...
+    lb_policy =
+      cond do
+        is_map(config) and Map.has_key?(config, :load_balancing_policy) ->
+          config.load_balancing_policy
+
+        lb_policy_opt ->
+          lb_policy_opt
+
+        true ->
+          nil
```
polvalente, line 406: This seems a bit redundant. Can `lb_policy_opt` be `false`? If not, we can turn this cond into an if

lib/grpc/client/connection.ex
```diff
@@ -367,63 +370,114 @@ defmodule GRPC.Client.Connection do
...
-          | virtual_channel: ch,
-            real_channels: %{"#{host}:#{port}" => ch}
-        }
+      {:error, reason} ->
+        {:error, reason}
+    end
+  end
+
+  defp build_real_channels(addresses, virtual_channel, norm_opts, adapter) do
+    Enum.into(addresses, %{}, fn %{port: port, address: host} ->
```
polvalente:
```suggestion
    Map.new(addresses, fn %{port: port, address: host} ->
```

purely stylistic suggestion

polvalente (commented):
Just these few points and I think we're good to go!
You can bump the version to 0.11.2 in this PR too

PR #451, opened 2025-10-21
[Feat] Error handler in streams

sleipnir:
## 🎯 Context

This PR introduces a **new error handling and side-effect model** for the `GRPC.Stream` module, covering both **unary** and **streaming** flows.
The implementation improves safety, consistency, and expressiveness across Flow-based pipelines used in gRPC message processing, ensuring that exceptions, errors, and inconsistent return values are handled uniformly and predictably.

---

### 🧩 1. effect/2

A new utility function to safely perform **side effects** within a stream pipeline **without modifying** the values being processed.
```elixir
GRPC.Stream.from(....)
|> GRPC.Stream.effect(fn x -> Logger.debug("Processing #{x}") end)
|> GRPC.Stream.run()
```

* Ensures exceptions raised inside the effect function **do not break the stream**.
* Uses the internal safe_invoke/2 mechanism to capture exceptions and thrown values.
* Behaves like a safe tap/2, preserving all original stream data.

---

### ⚙️ 2. map_error/2

Adds a declarative mechanism for **mapping or transforming errors** (`{:error, reason}`) within a stream into structured `GRPC.RPCError` values.
```elixir
GRPC.Stream.from([1, 2])
|> GRPC.Stream.map(fn
  2 -> raise "boom"
  x -> x
end)
|> GRPC.Stream.map_error(fn
  {:error, reason} ->
    GRPC.RPCError.exception(message: "Validation or runtime error: #{inspect(reason)}")
  other ->
    other
end)
```

Behavior:

* {:error, reason} → transformed into %GRPC.RPCError{}
* {:ok, value} → unwrapped to value
* exceptions from previous operators (e.g. map/2) → can be transformed into domain-specific GRPC.RPCError
* normal values are propagated unchanged

This enables **localized error recovery and translation**, useful for both input validation and unexpected runtime exceptions.

---

### 🔄 3. Unified Error Matching and Propagation

All stream operators now use the internal safe_invoke/2 wrapper, which standardizes how functions inside the pipeline behave.

Function result   | Propagated value | Description
---------------------- | ----------------------------- | ----------------
{:ok, value}          | value                           | Normalized success
{:error, reason}    | {:error, reason}            | Propagated as-is
raise (exception) | {:error, exception}        | Caught and wrapped
throw value         | {:error, {:throw, value}} | Uniformly captured
raw value            | same                            | Passed through unchanged

This makes all pipelines exception-safe — any operator can raise or throw, but the stream will continue gracefully with well-defined {:error, reason} items.

---

### 💡 4. Integration with run/1 and run_with/3

Both stream finalizers (run/1, run_with/3) have been updated to:

* Handle {:error, reason} tuples and convert them to %GRPC.RPCError{}.
* Support both unary and streaming flows.
* Ensure that errors in one element do not interrupt subsequent elements in the stream.
* Allow dry_run mode to test flow execution without emitting gRPC responses.

---

### ✅ 5. Test Coverage

New and extended tests added in GRPC.StreamTest include:

* `effect/2`: ensures side effects are applied safely and do not affect data flow, even when exceptions occur inside the callback.
* `map_error/2`: verifies correct error mapping and recovery.
* Validation of error propagation across all operators (`map`, `flat_map`, `filter`, `ask`, etc.).
* Resilience against process failures (`:process_not_alive`, `:timeout`) in `ask/3`.
* Full pipeline integration tests combining multiple operators and error recovery logic.

---

## 🧠 Impact

This refactor:

* Unifies and simplifies error handling across the entire streaming subsystem.
* Makes pipelines more robust — no unexpected crashes due to raised errors.
* Improves observability: all errors now flow through %GRPC.RPCError{}.
* Adds new extensibility points for metrics, tracing, and recovery.

## 🔮 Next Steps

* Integrate tracing hooks to improve runtime observability.
* Update the docs.

### DIFF
--- a/lib/grpc/client/connection.ex
+++ b/lib/grpc/client/connection.ex
@@ -273,7 +273,8 @@ defmodule GRPC.Client.Connection do
         :refresh,
         %{lb_mod: lb_mod, lb_state: lb_state, real_channels: channels, virtual_channel: vc} =
           state
-      ) do
+      )
+      when not is_nil(lb_mod) do
     {:ok, {prefer_host, prefer_port}, new_lb_state} = lb_mod.pick(lb_state)
 
     channel_key = "#{prefer_host}:#{prefer_port}"
@@ -293,6 +294,8 @@ defmodule GRPC.Client.Connection do
         {:noreply, %{state | lb_state: new_lb_state, virtual_channel: picked_channel}}
     end
   end
+
+  def handle_info(:refresh, state), do: {:noreply, state}
 
   def handle_info({:DOWN, _ref, :process, pid, reason}, state) do
     Logger.warning(


yordis:
> A new utility function to safely perform side effects within a stream pipeline without modifying the values being processed.

Other ecosystem call it https://doc.rust-lang.org/std/result/enum.Result.html#method.inspect_err

Eitherway, fine by me

lib/grpc/stream.ex
```diff
@@ -155,6 +155,19 @@ defmodule GRPC.Stream do
...
+
+      {:error, reason} ->
+        GRPC.RPCError.exception(message: "[Error] #{inspect(reason)}")
+
+      msg ->
+        msg
+    end)
     |> Enum.to_list()
     |> List.flatten()
     |> List.first()
```
polvalente: I don't recall _why_ we want to return the first result only and not the full materialized stream.

lib/grpc/stream.ex
```diff
@@ -190,19 +203,125 @@ defmodule GRPC.Stream do
...
-
     flow
-    |> Flow.map(fn msg ->
-      if not dry_run? do
-        send_response(from, msg)
-      end
+    |> Flow.map(fn
+      {:ok, msg} ->
+        send_response(from, msg, opts)
+
```
polvalente:
```suggestion
```

lib/grpc/stream.ex
```diff
@@ -190,19 +203,125 @@ defmodule GRPC.Stream do
...
+      msg ->
+        send_response(from, msg, opts)
+        flow
     end)
     |> Flow.run()
   end
 
+  @doc """
+  Intercepts and transforms **error tuples** or **unexpected exceptions** that occur
```
polvalente:
```suggestion
  Intercepts and transforms error tuples or unexpected exceptions that occur
```

lib/grpc/stream.ex
```diff
@@ -190,19 +203,125 @@ defmodule GRPC.Stream do
...
+  within a gRPC stream pipeline.
+
+  `map_error/3` allows graceful handling or recovery from errors produced by previous
+  operators (e.g. `map/2`, `flat_map/2`) or from validation logic applied to incoming data.
+
+  The provided `handler/1` function receives the error reason (or the exception struct)
```
polvalente: Should probably indicate that it receives a tuple {:error, reason} where reason is either a caught exception struct or a term. And in this regard, ideally we should tag these differently.

lib/grpc/stream.ex
```diff
@@ -190,19 +203,125 @@ defmodule GRPC.Stream do
...
+  and can either:
+
+    * Return a new error tuple — e.g. `{:error, new_reason}` — to re-emit a modified error.
+    * Return any other value to recover from the failure and continue the pipeline.
+
+  This makes it suitable for both **input validation** and **capturing unexpected runtime errors**
```
polvalente: Let's avoid these `**` bold tags as they make it harder to read the docs with IEx helpers and IMO don't add a lot of readability in Hex

lib/grpc/stream.ex
```diff
@@ -190,19 +203,125 @@ defmodule GRPC.Stream do
...
+  ## Examples
+
+      iex> GRPC.Stream.from([1, 2])
+      ...> |> GRPC.Stream.map(fn
+      ...>   2 -> raise "boom"
+      ...>   x -> x
+      ...> end)
+      ...> |> GRPC.Stream.map_error(fn
+      ...>   {:error, _reason} ->
+      ...>     GRPC.RPCError.exception(message: "Validation or runtime error")
```
polvalente: shouldn't this return {:error, GRPC.RPCError...}? That's what I expected from the docs above

lib/grpc/stream.ex
```diff
@@ -190,19 +203,125 @@ defmodule GRPC.Stream do
...
+      ...>   msg -> msg
```
polvalente: The docs seem to indicate that the function would only receive errors. Is that the case? If so, there shouldn't be a second clause here.

lib/grpc/stream/operators.ex
```diff
@@ -100,4 +157,26 @@ defmodule GRPC.Stream.Operators do
...
+        other -> other
+      end
+    rescue
+      e ->
+        {:error, e}
+    catch
+      kind, reason ->
+        {:error, {kind, reason}}
+    end
+  end
```
polvalente:
```suggestion
    res = fun.(arg)

    case res do
      {:ok, v} -> v
      {:error, reason} -> {:error, reason}
      other -> other
    end
    
  rescue
    e ->
      {:error, e}
  catch
    kind, reason ->
      {:error, {kind, reason}}
  end
```

polvalente: I think the exception should be caught with {:error, {:exception, exception}} instead to keep in line with {kind, reason} and the function IMO shouldn't unwrap {:ok, v} into v

lib/grpc/stream.ex
```diff
@@ -190,19 +203,125 @@ defmodule GRPC.Stream do
...
+
+  - Use this operator to implement **robust error recovery**, **input validation**, or
+    to normalize exceptions from downstream Flow stages into well-defined gRPC errors.
+  """
+  defdelegate map_error(stream, func), to: Operators
+
+  @doc """
+  Applies a **side-effecting function** to each element of the stream **without altering** its values.
+
+  The `effect/2` function is useful for performing **imperative or external actions**
```
polvalente: since we have map, maybe `each` would be a nice counterpart since we have Enum.each and Enum.map

lib/grpc/stream.ex
sleipnir: Because in this case we return the user protobuf struct response `%HelloReply{}` not a list `[%HelloReply{}]`. The list would break the adapter which would not know how to serialize the response to the client.

polvalente: How are the other responses returned to the client? Perhaps we should just return something that the adapter can treat as a no-op instead of returning a valid response here. Is there a chance that this is sending the first reply twice?

lib/grpc/stream.ex
sleipnir: Yes. You are right!

lib/grpc/stream.ex
sleipnir: Right again!

lib/grpc/stream/operators.ex
sleipnir: I changed it to send the error in the format compatible with {:error, {:exception, exception}}, but I'm still unwrapping the success return to avoid breaking the API and needing a new MAJOR

lib/grpc/stream.ex
sleipnir: Interesting, but we already have `run` and `run_with` that make this materialization occur. If we were to create an `each` here, it would have to fulfill the same role as the `run` and `run_with` we already have.

lib/grpc/stream/operators.ex
polvalente: Breaking APIs in 0.x is allowed in minor versions :)

lib/grpc/stream.ex
polvalente: I don't understand this comment

lib/grpc/stream.ex
sleipnir: This return type is expected by the adapter when the **RPC type is unary**. Otherwise, it expects it to be sent via `GRPC.Server.send_reply`. This is essentially the difference between `run` and `run_with`materialized functions. Since `run_with` operates on stream RPC types, it materializes the stream and sends a `GRPC.Server.send_reply` for **each item**, while `run` simply sends the stream's response (the first and only item in the stream).

Regarding duplication, I only see this being possible if the user does something incorrectly in their stream operators. The consequence would probably be an error, as the server wouldn't send two responses anyway. But this is a good point you raised.

polvalente: Does it make sense for this module to be used with unary (i.e. non streaming calls?)

polvalente: I think this module shouldn't support unary

polvalente: What happens of the user uses send_reply in the first stream item and returns the reply as the firs entry in the enumerable?

lib/grpc/stream.ex
sleipnir: You said it would be nice to have an `each` function, right? That's what I understood from your comment.
So I said that `each` would definitely have to materialize the stream (you can't have a stream operator after `each`). Then I said that we already have the equivalent of `each` in the `run` and `run_with` functions.

polvalente: I just meant that effect could be named each instead

sleipnir: because of the confusion I made I don't think it's a good name hahaha

polvalente: I don't think effect is a clear name, and map doesn't materialize the flow too, so each as its non-mapping counterpart does make sense

lib/grpc/stream.ex
sleipnir: > Does it make sense for this module to be used with unary (i.e. non streaming calls?)

I think we've had this discussion, or at least discussions very similar to this, a few times and we've gone down this path.

polvalente: If you can find the reasoning, please point me towards it. I don't think it makes sense for a module related to streams support non-streaming features

sleipnir: > What happens of the user uses send_reply in the first stream item and returns the reply as the firs entry in the enumerable?

the first response would send the trailers and the client-server interaction for that response would be terminated. But I think the doc is quite clear to indicate that the response is being sent automatically by `run`. But if not, I think it's just a matter of improving the doc

polvalente: This is very confusing and would be solved by not returning any of the stream items

sleipnir: > If you can find the reasoning, please point me towards it. I don't think it makes sense for a module related to streams support non-streaming features

From a gRPC perspective, at the connection level, everything is a stream; unary is just a type of RPC. This might be confusing for you, as you might be thinking in terms of HTTP, etc.
When we proposed the issue that gave rise to this API, we touched on the issue of uniformity, as we did in many others. It was a long discussion, but we realized that treating everything as a stream would be more elegant than providing two different interaction mechanisms. The idea was to create a more modern API inspired by other gRPC libraries. Technically, a stream also has its own advantages and disadvantages, and it's up to the end user to decide what to use or not (since you can opt out of using Stream for unary, we don't prevent the user from doing so).
Regarding your reservations, I don't see how to duplicate messages to the end client.
However, you can continue with the old approach; the API is still there. I don't see this as a serious problem, or even a problem at all. They're just different worldviews that can coexist.

polvalente: Right, so of we're treating everything as a stream, I don't get why we have to return the first item from it. Either return all or none at all.

This is a code smell to me any way we put this.

Maybe it's a question of changing the contract elsewhere in the library, or introducing a special no-op return here to be matched elsewhere.

polvalente: Regarding duplication, I don't think there can be any given that the trailers are sent with the reply

lib/grpc/stream.ex
sleipnir: For me, a side effect, which is actually what I envisioned for this, has nothing to do with an 'each', whose name is intended to iterate over a set. However, that's not what effect is doing. It works very similarly to a 'Kernel.tap', so it's an effect because it doesn't affect the flow, which is very common terminology for functional languages. I've never seen an each used as a map, even though it's internally implemented as a map. It's not the semantics it emits in its name; it's not intended to map anything, only to iterate. Similarly, an effect isn't intended to iterate or map anything, only to generate value from an item in the flow as a side effect of the flow itself.

Anyway, it's just a name. I don't agree with each, but if you want to go that route, let's not fight over it, although I make it clear that I don't agree with the name.

lib/grpc/stream.ex
sleipnir: > 
> This is a code smell to me any way we put this.
> 

Ok so this is a code smell of the language itself that doesn't allow materializing a single item from a stream. This is not my fault 😄

sleipnir: > Regarding duplication, I don't think there can be any given that the trailers are sent with the reply

```
stream
|> GRPC.Server.send_reply(response)
|> GRPC.Server.send_trailers(@default_trailers)
```

this is from the adapter

sleipnir: After a long night working together, @polvalente and I reached a consensus. I'll consider this resolved through PR https://github.com/elixir-grpc/grpc/pull/453 and update this branch with the changes.

README.md
```diff
@@ -101,8 +105,8 @@ defmodule HelloworldStreams.Server do
   alias Helloworld.HelloReply
 
   @spec say_unary_hello(HelloRequest.t(), GRPC.Server.Stream.t()) :: any()
-  def say_unary_hello(request, _materializer) do
-    GRPC.Stream.unary(request)
+  def say_unary_hello(request, materializer) do
+    GRPC.Stream.unary(request, materializer: materializer)
     |> GRPC.Stream.map(fn %HelloReply{} = reply ->
```
polvalente:
```suggestion
    request
    |> GRPC.Stream.unary(materializer: materializer)
    |> GRPC.Stream.map(fn %HelloReply{} = reply ->
```

README.md
```diff
@@ -144,28 +148,104 @@ def say_bid_stream_hello(request, materializer) do
...
+...> |> GRPC.Stream.map_error(fn
+...>   {:error, {:exception, _reason}} ->
+...>     {:error, GRPC.RPCError.exception(message: "Booomm")}
+...> end)
+``` 
+
+In this example:
+
+* The function inside `map/2` raises an exception for the value `2`.
+* map_error/2 captures and transforms that error into a structured `GRPC.RPCError` response.
```
polvalente:
```suggestion
* `map_error/2` captures and transforms that error into a structured `GRPC.RPCError` response.
```

lib/grpc/stream.ex
```diff
@@ -193,19 +193,71 @@ defmodule GRPC.Stream do
...
+
+      msg ->
+        send_response(from, msg, opts)
+        flow
     end)
     |> Flow.run()
   end
 
+  @doc """
+  Applies a **side-effecting function** to each element of the stream **without altering** its values.
```
polvalente:
```suggestion
  Applies a side-effect function to each element of the stream without altering its values.
```

Let's remove all of these bold highlights

lib/grpc/stream.ex
```diff
@@ -223,9 +275,9 @@ defmodule GRPC.Stream do
   ## Returns
 
     - Updated stream if successful.
-    - `{:error, item, reason}` if the request fails or times out.
+    - `{:error, reason}` if the request fails or times out.
```
polvalente: This is a bit redundant with the typespec

lib/grpc/stream/operators.ex
```diff
@@ -9,9 +9,10 @@ defmodule GRPC.Stream.Operators do
   @type reason :: any()
 
   @spec ask(GRPCStream.t(), pid | atom, non_neg_integer) ::
-          GRPCStream.t() | {:error, any(), :timeout | :not_alive}
+          GRPCStream.t() | {:error, :timeout | :process_not_alive}
   def ask(%GRPCStream{flow: flow} = stream, target, timeout \\ 5000) do
-    mapper = fn item -> do_ask(item, target, timeout, raise_on_error: false) end
+    # mapper = fn item -> do_ask(item, target, timeout, raise_on_error: false) end
```
polvalente: stray comment

lib/grpc/stream/operators.ex
```diff
@@ -45,25 +46,46 @@ defmodule GRPC.Stream.Operators do
...
+  @spec effect(GRPCStream.t(), (term -> term())) :: GRPCStream.t()
+  def effect(%GRPCStream{flow: flow} = stream, effect_fun) when is_function(effect_fun, 1) do
+    %GRPCStream{
+      stream
+      | flow:
+          Flow.map(flow, fn flow_item ->
+            tap(flow_item, fn item -> safe_invoke(effect_fun, item) end)
+          end)
+    }
+  end
```
polvalente: let's break this up a bit for readability

lib/grpc/stream/operators.ex
```diff
@@ -73,7 +95,41 @@ defmodule GRPC.Stream.Operators do
...
+            other ->
+              other
+          end
+
+        {:ok, other} ->
+          other
+
+        other ->
+          other
+      end)
```
polvalente: can we extract this to a defp?

test/grpc/integration/server_test.exs
```diff
@@ -32,19 +36,25 @@ defmodule GRPC.Integration.ServerTest do
...
+      |> GRPC.Stream.run()
     end
 
-    def stream_messages(msg_request, stream) do
-      Enum.each(1..5, fn i ->
-        msg = %Transcode.Message{
+    def stream_messages(msg_request, materializer) do
+      Stream.cycle(1..5)
+      |> Stream.take(5)
+      |> GRPC.Stream.from()
```
polvalente:
```suggestion
      1..5
      |> GRPC.Stream.from()
```

Doesn't this work? If not, we should make it work in a separate PR. Suggestion for if enumerables aren't allowed directly:

```suggestion
      1..5
      |> Stream.take(5)
      |> GRPC.Stream.from()
```

test/grpc/stream_test.exs
```diff
@@ -165,6 +165,141 @@ defmodule GRPC.StreamTest do
     end
   end
 
+  describe "ask/3 error handling" do
+    test "returns timeout error if response not received in time" do
+      pid =
+        spawn(fn ->
```
polvalente:
```suggestion
        spawn_link(fn ->
```

test/grpc/stream_test.exs
```diff
@@ -165,6 +165,141 @@ defmodule GRPC.StreamTest do
...
+          # do not send any response
+          receive do
+            _ -> :ok
+          end
```
polvalente:
```suggestion
          Process.sleep(:infinity)
```

polvalente: Also instead of spawning we could have self() be the pid in ask and maybe even have an assert_receive

test/grpc/integration/server_test.exs
sleipnir: I confess that the cycle was more intentional for me. But in the end, I got the version of the list + take(5)

polvalente: I don't get why cycle would matter since take limits the stream to 5 items anyway

sleipnir: it has already been resolved

PR #456, opened 2025-10-28
fix: refresh error spam on direct_state (no lb)

JoeriDijkstra:
Some context - even with the direct state where no lb_mod is picked the refresh is sent. This causes the lb_mod.pick on line 277 to crash, even though we don't use any loadbalancing


sleipnir (approved):

sleipnir:
@JoeriDijkstra please format and send again

sleipnir:
Thank you for your contribution @JoeriDijkstra

PR #458, opened 2025-11-06
docs: overhaul library documentation

sleipnir:
This PR improves the documentation as whole:

- Groups modules in the ex_doc rendering
- Moves the example repositories into Livebooks
- Adds function groups to the GRPC.Stream module docs


.github/workflows/ci.yml
```diff
@@ -52,7 +52,7 @@ jobs:
       - name: Install Dependencies
         run: mix deps.get 1>/dev/null
       - name: Run Tests
-        run: mix test --warnings-as-errors
+        run: mix test
```
polvalente: We should keep this. Why is it being removed?

guides/advanced/load_balancing.md
```diff
@@ -0,0 +1,47 @@
...
+
+| Scheme    | Example                     | Description                                  |
+|:----------|:----------------------------|:---------------------------------------------|
+| `dns://`  | `"dns://example.com:50051"` | Resolves via DNS `A/AAAA` records            |
+| `ipv4:`   | `"ipv4:10.0.0.5:50051"`     | Connects directly to an IPv4 address         |
+| `unix:`   | `"unix:/tmp/service.sock"`  | Connects via a Unix domain socket            |
+| `xds:///` | `"xds:///my-service"`       | Resolves via xDS control plane (Envoy/Istio) |
+| none      | `"127.0.0.1:50051"`         | Implicit DNS (default port `50051`)          |
+
+>__Note__: The implementation of xDS is still underway.
```
polvalente: if this is not supported yet it shouldn't be listed as supported

guides/cheatsheets/streams.cheatmd
```diff
@@ -0,0 +1,127 @@
...
+```elixir
+iex> parent = self()
+iex> stream =
+...>   GRPC.Stream.from(request)
+...>   |> GRPC.Stream.effect(fn x -> send(parent, {:seen, x*2}) end)
+...>   |> GRPC.Stream.to_flow()
+...>   |> Enum.to_list()
+iex> assert_receive {:seen, 2}
+iex> assert_receive {:seen, 4}
+iex> assert_receive {:seen, 6}
```
polvalente: assert_receive requires importing ExUnit.Assertions.

Examples should ideally be self-contained. I suggest replacing this with `flush()` which is auto-imported in IEx.

guides/getting_started/client.md
```diff
@@ -0,0 +1,86 @@
...
+
+---
+
+## Basic Connection and RPC
+
+Typically, you start this client supervisor as part of your application's supervision tree:
+
+```elixir
+children = [
+  {GRPC.Client.Supervisor, []}
```
polvalente:
```suggestion
  GRPC.Client.Supervisor
```

polvalente: Let's also change this in the moduledoc

guides/getting_started/installation.md
```diff
@@ -0,0 +1,14 @@
+# Installation
+
+The package can be installed as:
+
+```elixir
+def deps do
+  [
+    {:grpc, "~> 0.11"},
+    {:protobuf, "~> 0.14"}, # optional for import wellknown google types
```
polvalente:
```suggestion
    {:protobuf, "~> 0.14"}, # optional for importing well-known Google gRPC types
```

guides/getting_started/installation.md
```diff
@@ -0,0 +1,14 @@
...
+    {:grpc_reflection, "~> 0.2"}, # optional enable grpc reflection
```
polvalente:
```suggestion
    {:grpc_reflection, "~> 0.2"}, # optional for enabling gRPC reflection
```

guides/getting_started/installation.md
```diff
@@ -0,0 +1,14 @@
...
+    {:protobuf_generate, "~> 0.1", only: :dev} # optional for protobuf codegen
```
polvalente:
```suggestion
    {:protobuf_generate, "~> 0.1", only: :dev} # optional for Protobuf code generation with plugins
```

guides/getting_started/introduction.md
```diff
@@ -0,0 +1,22 @@
+# Introduction
+
+GRPC is a fully featured Elixir implementation of the gRPC protocol (grpc.io),
```
polvalente:
```suggestion
`GRPC` is a fully featured Elixir implementation of the gRPC protocol (grpc.io),
```

guides/getting_started/quickstart.livemd
```diff
@@ -0,0 +1,124 @@
+# Quickstart
```
polvalente, line 1: I think Installation, Introduction, CodeGen and quickstart could all be part of the same file

guides/getting_started/stream.livemd
```diff
@@ -0,0 +1,216 @@
+# Streaming
+
```
polvalente, line 2: This livebook is missing a description of what it teaches/showcases

lib/grpc/client/resolver/xds.ex
```diff
@@ -1,4 +1,5 @@
 defmodule GRPC.Client.Resolver.XDS do
+  @moduledoc false
```
polvalente, line 2: This is yet another sign to remove XDS mentions from the docs and API

lib/grpc/server.ex
```diff
@@ -454,6 +454,7 @@ defmodule GRPC.Server do
 
       iex> GRPC.Server.send_reply(stream, reply)
   """
+  @deprecated "Use GRPC.Stream API instead"
```
polvalente:
```suggestion
  @deprecated "Use GRPC.Stream API instead", since: "0.12.0"
```

lib/grpc/stream.ex
```diff
@@ -128,6 +130,7 @@ defmodule GRPC.Stream do
 
     A `Flow` pipeline.
   """
+  @doc type: :transformer
```
polvalente:
```suggestion
  @doc type: :transforms
```

let's use transforms as the tag instead of transformer

lib/grpc/stream.ex
```diff
@@ -142,7 +145,22 @@ defmodule GRPC.Stream do
...
+          %HelloReply{} = reply ->
+            %HelloReply{message: "[Reply] message"}
+
+          {:error, _reason} ->
+            GRPC.RPCError.exception(message: "[Error] Something bad happened")
+        end)
+        |> GRPC.Stream.run()
+      end
   """
+  @doc type: :materializer
```
polvalente:
```suggestion
  @doc type: :materialization
```

lib/grpc/stream.ex
```diff
@@ -255,6 +289,7 @@ defmodule GRPC.Stream do
   - The use of `effect/2` ensures that the original item is returned unchanged,
   enabling seamless continuation of the pipeline.
   """
+  @doc type: :utility
```
polvalente: util/ultility is non-informative. We should have another name.
doc types should be less granular than what you have here. For instance, filter is a :transform, :mapper is also a :transform

lib/grpc/stub.ex
```diff
@@ -164,6 +164,7 @@ defmodule GRPC.Stub do
 
   * When using DNS or xDS targets, the connection layer periodically refreshes endpoints.
   """
+  @doc type: :client_connection
```
polvalente: Stub is client by definition.
Also, there are so few functions here I don't think there's much to gain in using types

mix.exs
```diff
@@ -43,12 +40,14 @@ defmodule GRPC.Mixfile do
...
       {:protobuf, "~> 0.14"},
-      {:protobuf_generate, "~> 0.1.1", only: [:dev, :test]},
       {:mint, "~> 1.5"},
-      {:ex_doc, "~> 0.29", only: :dev},
+      {:telemetry, "~> 1.0"},
+      {:protobuf_generate, "~> 0.1.1", only: [:dev, :test]},
       {:ex_parameterized, "~> 1.3.7", only: :test},
       {:mox, "~> 1.2", only: :test},
-      {:telemetry, "~> 1.0"}
+      {:ex_doc, "~> 0.29", only: [:dev, :docs]},
```
polvalente: We should pin a more recent ex_doc version here

mix.exs
```diff
@@ -61,6 +60,139 @@ defmodule GRPC.Mixfile do
...
+        "Functions: Stream Creation": &(&1[:type] == :creation),
+        "Functions: Stream Materializers": &(&1[:type] == :materializer),
+        "Functions: Stream Transformers": &(&1[:type] == :transformer),
+        "Functions: Stream Mappers": &(&1[:type] == :mapper),
+        "Functions: Stream Reducers": &(&1[:type] == :reducer),
+        "Functions: Stream Filters": &(&1[:type] == :filter),
+        "Functions: Stream Windowning": &(&1[:type] == :windowing),
+        "Functions: Stream Utilities": &(&1[:type] == :utility),
+        "Functions: Client Connection": &(&1[:type] == :client_connection),
+        "Functions: Client Request": &(&1[:type] == :client_request)
```
polvalente: `Functions: Stream *` -> `Functions: *`, as they're only used in Stream. Likewise for Client if we keep groups there, but I don't think we should.

mix.exs
```diff
@@ -61,6 +60,139 @@ defmodule GRPC.Mixfile do
...
+        renderMathInElement(document.body, {
+          delimiters: [
+            { left: "$$", right: "$$", display: true },
+            { left: "$", right: "$", display: false },
+          ]
+        });
+      });
+    </script>
+    """
+  end
```
polvalente: I don't think we have Latex markdown in our docs, so this shouldn't be added.

lib/grpc/server.ex
polvalente: Actually, isn't send_reply still used for non-unary streams?

lib/grpc/client/resolver/xds.ex
sleipnir: Since the scope here is documentation, I will remove or implement this later in another PR. I have already removed it from the documentation for now.

lib/grpc/server.ex
sleipnir: In fact, the streams API covers all types of RPCs in the gRPC protocol in a unified way. Therefore, we don't need this for making unary requests. However, I've reverted this deprecation for now.

guides/getting_started/quickstart.livemd
sleipnir: I still kept Codegen separate and condensed the others into the quickstart. Codegen follows a different approach, and I would like to be more detailed with it in its current form.

lib/grpc/server.ex
polvalente: I think a good stepping stone would be to replace all usage of the send_reply function in our repo with the equivalent GRPC.Stream calls

Issue #389, opened 2024-08-08
Clarification on handling of raised RPCErrors as crash events


ksanderer:
Hello Team!

Firstly, thank you for the recent update and the introduction of the crash report logger metadata. This feature is indeed beneficial for debugging and monitoring purposes (related PR and Issue: #381, #377)

However, I have a concern, or perhaps I have misunderstood something regarding the current implementation where all raised errors are considered crash events. Based on the logs, it appears that even standard gRPC errors, such as bad_request or not_found, are treated as crash events. Here is an example:

```
13:12:02.786 [error] ** (GRPC.Server.Adapters.ReportException) Exception raised while handling /api.messages.v1.MessageService/UpdateMessage:
** (GRPC.RPCError) Message with provided id doesn’t exist.
...

%GRPC.Server.Adapters.ReportException{
    kind: :error, 
    reason: %GRPC.RPCError{
        status: 5,  # GRPC.Status.not_found()
        message: \"Message with provided id doesn’t exist.\"}, 
    ...
}
```

These are "normal" gRPC errors supported by the protocol, indicating statuses like bad_request or not_found. Treating these as crash events might lead to unnecessary noise in the logs and potentially obscure actual issues that need attention.

By "normal," I refer to the Google APIs guidelines:
> Because most Google APIs use resource-oriented API design, the error handling follows the same design principle by using a small set of standard errors with a large number of resources. For example, instead of defining different kinds of "not found" errors, the server uses one standard google.rpc.Code.NOT_FOUND error code and tells the client which specific resource was not found. 
>
> [Google Cloud API Error Model](https://cloud.google.com/apis/design/errors#error_model)

According to the [documentation](https://hexdocs.pm/grpc/GRPC.RPCError.html), the only way to propagate `google.rpc.Status` is by raising a `GRPC.RPCError`.

At my workplace, we adhere to the approach outlined in the documentation, as demonstrated below:

```
  def delete_message(%DeleteMessageRequest{organization_id: ""}, _stream) do
    raise GRPC.RPCError,
      status: GRPC.Status.invalid_argument(),
      message: "`organization_id` is required"
  end
```

#### Questions:
1. **Rationale**: Could you please clarify the rationale behind treating all errors as crash events? 
2. **Strategy**: What would be the proper strategy to handle these errors or suppress these crash reports for standard gRPC errors?

Looking forward to your insights on this matter. Thank you for your time and assistance.

sleipnir:
[...]

PR #462, opened 2025-11-07
Add `exception_log_filter` option to server

ollien:
Fixes https://github.com/elixir-grpc/grpc/issues/389

This PR adds the ability to filter which exceptions flow through the `ReportException` flow. Users can pass a module/function pair to the supervisor, which decides whether or not each exception should be logged. Of course, this allows them to filter on gRPC status codes, too, by simply pattern matching on `%GRPC.RPCError{status: status}`.

I'm certainly open to adjustments here!

### DIFF
--- a/test/grpc/integration/server_test.exs
+++ b/test/grpc/integration/server_test.exs
@@ -344,7 +344,7 @@ defmodule GRPC.Integration.ServerTest do
 
   defmodule ExceptionFilterMustBeRaisedError do
     def filter(exception) do
-      data = get_in(exception.adapter_extra[:req][:headers]["test-data"])
+      data = exception.adapter_extra[:req][:headers]["test-data"]
 
       {pid, ref} = :erlang.binary_to_term(data)
       send(pid, {:exception_log_filter, ref})


lib/grpc/server/adapters/cowboy/handler.ex
```diff
@@ -110,6 +115,19 @@ defmodule GRPC.Server.Adapters.Cowboy.Handler do
...
+  defp extract_exception_log_filter_opt(opts) do
+    case opts[:exception_log_filter] do
+      {module, func_name} when is_atom(module) and is_atom(func_name) ->
+        {module, func_name}
+
+      nil ->
+        nil
+
+      invalid ->
+        raise ArgumentError, "invalid exception log filter: #{inspect(invalid)}"
```
sleipnir, line 127: I don't like this raise. Can't we return this function without breaking it?

sleipnir (commented):
I only have one question: is it possible to avoid that raise? Other than that, it's okay for me.

lib/grpc/server/adapters/cowboy/handler.ex
polvalente: I don't think we should. This is intentional because if this is reached, an invalid option that does not conform to the spec was passed.

polvalente: Not having this raise would mean the function would fail silently.

Issue #468, opened 2025-11-22
[Feature] HTTP transcoding: configurable `emit_unpopulated` for `GRPC.Codec.JSON`


bolunusi-drw:
**Is your feature request related to a problem? Please describe.**

When using `http_transcode: true` empty/default fields in JSON responses are omitted because `GRPC.Codec.JSON` doesn't set `emit_unpopulated: true`. This makes HTTP/JSON output inconsistent with gRPC output when default values are present (e.g., empty lists).

**Describe the solution you'd like**

Add a config option to enable `emit_unpopulated: true` for the [JSON codec](https://github.com/elixir-grpc/grpc/blob/d26d48a206d34daf8e019be00eb59d20e4943e8a/lib/grpc/codec/json.ex#L41) used in HTTP transcoding or allow passing in a custom codec when `http_transcode` is `true`.

**Describe alternatives you've considered**

I've tried passing in a custom codec but it gets [overridden](https://github.com/elixir-grpc/grpc/blob/d26d48a206d34daf8e019be00eb59d20e4943e8a/lib/grpc/server.ex#L139) in `GRPC.Server`.

**Additional context**

Assume we have

```proto
service UserService {
  rpc ListUsers(ListUsersRequest) returns (ListUsersResponse) {
    option (google.api.http) = {
      get: "/api/v1/users"
    };
  }
}

message User {
  string username = 2;
  string email = 3;
}

message ListUsersRequest {}

message ListUsersResponse {
  repeated User users = 1;
}
```

```elixir
iex> Protobuf.JSON.encode!(%ListUsersResponse{users: []})
"{}"

iex> Protobuf.JSON.encode!(%ListUsersResponse{users: []}, emit_unpopulated: true)
"{"users":[]}"
```

In the service, this translates to:

```bash
# Currently
curl localhost:50051/api/v1/users
{}

# With emit_unpopulated
curl localhost:50051/api/v1/users
{"users": []}

# Default gRPC
grpcurl ...
{"users": []}
```

sleipnir:
Hi @bolunusi-drw thanks for the suggestion. I've kept this as the main behavior for now; in my understanding, it was a mistake not to respect the protobuf defaults.

polvalente:
One thing I don't understand here is where do the default values come from. I might be missing something, but I don't see where in the proto spec a default value is set besides just knowing that the users field is a list. 

And more importantly, the encoding shouldn't matter if the decoder side decodes things consistently.

[...]

PR #469, opened 2025-11-22
[Feat] protobuf json defaults

sleipnir:
This closes https://github.com/elixir-grpc/grpc/issues/468

### DIFF
--- a/test/grpc/codec/json_test.exs
+++ b/test/grpc/codec/json_test.exs
@@ -1,0 +1,99 @@
+defmodule GRPC.Codec.JSONTest do
+  use ExUnit.Case, async: true
+
+  alias GRPC.Codec.JSON
+
+  describe "name/0" do
+    test "returns the codec name" do
+      assert JSON.name() == "json"
+    end
+  end
+
+  describe "encode/1" do
+    test "encodes a protobuf struct to JSON" do
+      request = %Helloworld.HelloRequest{name: "Alice"}
+      encoded = JSON.encode(request)
+
+      assert is_binary(encoded)
+      decoded_map = Jason.decode!(encoded)
+      assert decoded_map["name"] == "Alice"
+    end
+
+    test "includes unpopulated fields with default values" do
+      request = %Helloworld.HelloRequest{name: "Bob", duration: nil}
+      encoded = JSON.encode(request)
+
+      decoded_map = Jason.decode!(encoded)
+      assert Map.has_key?(decoded_map, "duration")
+    end
+
+    test "encodes empty struct with all default fields" do
+      request = %Helloworld.HelloRequest{name: "", duration: nil}
+      encoded = JSON.encode(request)
+
+      decoded_map = Jason.decode!(encoded)
+      assert Map.has_key?(decoded_map, "name")
+      assert Map.has_key?(decoded_map, "duration")
+      assert decoded_map["name"] == ""
+    end
+
+    test "encodes nested messages with defaults" do
+      reply = %Helloworld.HelloReply{message: ""}
+      encoded = JSON.encode(reply)
+
+      decoded_map = Jason.decode!(encoded)
+      assert Map.has_key?(decoded_map, "message")
+      assert decoded_map["message"] == ""
+    end
+  end
+
+  describe "decode/2" do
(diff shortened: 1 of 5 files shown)


test/grpc/codec/json_test.exs
```diff
@@ -0,0 +1,99 @@
...
+      assert is_binary(encoded)
+      decoded_map = Jason.decode!(encoded)
+      assert decoded_map["name"] == "Alice"
+    end
+
+    test "includes unpopulated fields with default values" do
+      request = %Helloworld.HelloRequest{name: "Bob", duration: nil}
+      encoded = JSON.encode(request)
+
+      decoded_map = Jason.decode!(encoded)
```
polvalente, line 26: Why encode with one encoder and decode with another?

test/grpc/codec/json_test.exs
```diff
@@ -0,0 +1,99 @@
...
+      assert Map.has_key?(decoded_map, "duration")
+    end
+
+    test "encodes empty struct with all default fields" do
```
polvalente, line 30: Why encode all fields?

test/grpc/codec/json_test.exs
```diff
@@ -0,0 +1,99 @@
...
+    test "decodes JSON with missing fields" do
+      json = ~s({"name": "Dave"})
+      result = JSON.decode(json, Helloworld.HelloRequest)
+
+      assert result["name"] == "Dave"
+      refute Map.has_key?(result, "duration")
+    end
+
+    test "decodes complex JSON structures" do
+      json = ~s({"message": "Hello, World!"})
```
polvalente, line 73: Not sure this us a "complex" json structure

test/grpc/codec/json_test.exs
```diff
@@ -0,0 +1,99 @@
...
+      encoded = JSON.encode(original)
+      decoded = JSON.decode(encoded, Helloworld.HelloRequest)
+
+      assert decoded["name"] == "Eve"
+      assert decoded["duration"] == 100
+    end
+
+    test "handles empty values correctly" do
+      original = %Helloworld.HelloRequest{name: "", duration: nil}
+      encoded = JSON.encode(original)
```
polvalente, line 92: This is not consistent with the pattern in other tests

test/grpc/codec/json_test.exs
sleipnir: Because that was exactly what I wanted to test. But I'll change it.

polvalente: You can just add a comment. By reading this it just seems arbitrary.

I'd expect the converse test as well then, encoding with the latter and decoding with the former

polvalente: You can just add a comment. By reading this it just seems arbitrary.

I'd expect the converse test as well then, encoding with the latter and decoding with the former

test/grpc/codec/json_test.exs
sleipnir: I didn't understand which test you're referring to.

test/grpc/codec/json_test.exs
sleipnir: I swear I didn't catch the cue.

test/grpc/codec/json_test.exs
polvalente: This was a general question regarding the motivation of the PR. But I read the issue afterwards and I understand it a bit better.

As I mentioned, I think we should add the config and change the default as you did here

polvalente: This was a general question regarding the motivation of the PR. But I read the issue afterwards and I understand it a bit better.

As I mentioned, I think we should add the config and change the default as you did here

test/grpc/codec/json_test.exs
polvalente: This test is encoding and decoding with the same decoder, other tests encode with JSON and decode with Jason

polvalente: This test is encoding and decoding with the same decoder, other tests encode with JSON and decode with Jason

sleipnir: Ah no more, now all have the same

PR #473, opened 2025-11-23
[Docs]: Explain why custom codec does not works with http transcode

sleipnir:
This better documents the codecs and explains why custom codecs are not supported for the HTTP transcode protocol.


lib/grpc/server.ex
```diff
@@ -107,6 +107,7 @@ defmodule GRPC.Server do
         "message": "Hello gRPC"
       }
 
+  Custom codecs does not work with HTTP/JSON transcoding.
```
polvalente:
```suggestion
  Custom codecs do not work with HTTP/JSON transcoding.
```

PR #477, opened 2025-11-26
[Feat]: separating client and server into two distinct packages

sleipnir:


.github/workflows/release.yml
```diff
@@ -0,0 +1,90 @@
+name: Release to Hex
```
polvalente: As we discussed, we won't have a workflow for releasing to hex

.formatter.exs
```diff
@@ -1,12 +1 @@
...
-    "{config,lib,test}/**/*.{ex,exs}",
-    "examples/*/{config,lib,priv}/*.ex"
-  ],
-  import_deps: [:protobuf],
-  locals_without_parens: [rpc: 3, intercept: 1, intercept: 2, run: 1, run: 2],
-  export: [
-    locals_without_parens: [rpc: 3, intercept: 1, intercept: 2, run: 1, run: 2]
-  ]
-]
+[subdirectories: ["grpc_core", "grpc_server", "grpc_client"]]
```
polvalente, line 1: This file shouldn't exist. top level of the repo should just contain the readme and the folders for each sub-library

.gitignore
```diff
@@ -24,4 +24,6 @@ erl_crash.dump
 
 .elixir_ls
 
+.elixir_tools
```
polvalente, line 27: I think we should remove this and the .elixir_ls line from .gitignore. These should be set per user in their own global gitignores

.formatter.exs
polvalente: Actually, I think I understand why this is here. Tell me if I'm correct, but this and the "skeleton" mix project serve the purpose of running mix test/compile/format in each subdirectory, is this right?

Makefile
```diff
@@ -13,10 +13,10 @@ endif
```
polvalente: How about we remove this makefile?

benchmark/lib/benchmark/client_worker.ex
```diff
@@ -3,6 +3,7 @@ defmodule Benchmark.ClientWorker do
   use GenServer
 
   def init(args) do
+    {:ok, _pid} = GRPC.Client.Supervisor.start_link()
```
polvalente: this should be in the benchmark application.ex file along with the Server.Supervisor

grpc_core/README.md
```diff
@@ -0,0 +1,46 @@
...
+end
+```
+
+## Usage
+
+Most users won't need to depend on `grpc_core` directly. Instead, use:
+
+- `{:grpc_client, "~> 0.12"}` for client-only applications
+- `{:grpc_server, "~> 0.12"}` for server-only applications  
+- `{:grpc, "~> 0.12"}` for applications using both client and server
```
polvalente: I believe we should not publish grpc anymore and just publish grpc_server and grpc_client separately.

benchmark/lib/benchmark/client_worker.ex
sleipnir: Did you mean server_manager.ex?

grpc_client/lib/grpc_client.ex
```diff
@@ -0,0 +1,3 @@
+defmodule GRPC.Client do
+  @moduledoc false
```
polvalente: Why have this module at all?

benchmark/lib/benchmark/client_worker.ex
sleipnir: I just followed the existing pattern.

.formatter.exs
sleipnir: correct

grpc_client/guides/advanced/custom_codecs.md
```diff
@@ -301,4 +301,4 @@ If you need custom encoding for HTTP endpoints, consider:
 
 For more examples, see the test suite:
 - `test/grpc/integration/custom_codec_test.exs`
-- `test/grpc/codec/proto_test.exs`
+- `test/grpc/codec/proto_test.exs`
```
polvalente, line 304: here's a general thought: should we keep grpc as a pseudo library for documentation?

Or only split into grpc and grpc_server (and grpc_core) and have client be the main lib?

Makefile
sleipnir: For me ok

grpc_client/mix.exs
```diff
@@ -0,0 +1,99 @@
...
+      {:grpc_core, path: "../grpc_core"},
+      # {:grpc_core, "~> 1.0"}, # Uncomment for hex release
+      {:gun, "~> 2.0"},
+      {:mint, "~> 1.5"},
+      {:castore, "~> 0.1 or ~> 1.0", optional: true},
+      {:ex_doc, "~> 0.39", only: [:dev, :docs], runtime: false},
+      {:ex_parameterized, "~> 1.3.7", only: :test},
+      {:mox, "~> 1.2", only: :test},
+      {:grpc_server, path: "../grpc_server", only: :test}
+      # {:grpc_server, "~> 1.0", only: :test} # Uncomment for hex release
```
polvalente: if it's only: :test we don't need to uncomment for hex release

grpc_client/mix.exs
```diff
@@ -0,0 +1,99 @@
...
+        "CHANGELOG.md",
+        "guides/getting_started/client.md",
+        "guides/advanced/custom_codecs.md",
+        "guides/advanced/load_balancing.md"
+      ],
+      groups_for_modules: [
+        "Client Core": [
+          GRPC.Stub,
+          GRPC.Channel,
+          GRPC.Client.Stream,
```
polvalente, line 67: I think this is a sign that we should make grpc be the client library (and keep docs here)

grpc_core/README.md
```diff
@@ -0,0 +1,45 @@
...
+def deps do
+  [
+    {:grpc_core, "~> 0.12"}
+  ]
+end
+```
+
+## Usage
+
+Most users won't need to depend on `grpc_core` directly. Instead, use:
```
polvalente:
```suggestion
Most users shouldn't depend on `grpc_core` directly. Instead, use:
```

grpc_core/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Documentation
+
+For full documentation, see the main [gRPC Elixir](https://hexdocs.pm/grpc) documentation.
+def deps do
+  [
+    {:grpc_core, "~> 0.1.0"}
+  ]
+end
+```
```
polvalente: stray text?

grpc_client/guides/advanced/custom_codecs.md
sleipnir: In my opinion, if we have a grpc package, it should contain both dependencies and the complete documentation, but I personally disagree with this. For me, we only have grpc_server and grpc_client. I only kept the module namespaces to avoid major refactoring by our users.

grpc_client/mix.exs
sleipnir: I don't see this sign; to me, this is just documentation.

grpc_core/lib/grpc/transport/http2.ex
```diff
@@ -64,7 +61,7 @@ defmodule GRPC.Transport.HTTP2 do
     [
       # It seems only gRPC implemenations only support "application/grpc", so we support :content_type now.
       {"content-type", content_type(opts[:content_type], codec)},
-      {"user-agent", "grpc-elixir/#{opts[:grpc_version] || GRPC.version()}"},
+      {"user-agent", "grpc-elixir/#{opts[:grpc_version] || GRPC.Core.version()}"},
```
polvalente: use the GRPC.Core.user_agent version here instead

grpc_core/mix.exs
```diff
@@ -0,0 +1,61 @@
+defmodule GRPC.Core.MixProject do
+  use Mix.Project
+
+  @source_url "https://github.com/elixir-grpc/grpc"
+  @version "1.0.0"
```
polvalente:
```suggestion
  @version "0.12.0"
```

grpc_core/mix.exs
```diff
@@ -0,0 +1,61 @@
+defmodule GRPC.Core.MixProject do
+  use Mix.Project
+
+  @source_url "https://github.com/elixir-grpc/grpc"
```
polvalente: This link is likely wrong

grpc_server/README.md
```diff
@@ -0,0 +1,516 @@
...
+- [Client Usage](#client-usage)
+  - [Basic Connection and RPC](#basic-connection-and-rpc)
+  - [Using Interceptors](#using-interceptors)
+  - [Target Schemes and Resolvers](#target-schemes-and-resolvers)
+    - [Supported formats](#supported-formats)
+    - [Example (DNS)](#example-dns)
+    - [Example (Unix socket)](#example-unix-socket)
+  - [Compression and Metadata](#compression-and-metadata)
+  - [Client Adapters](#client-adapters)
+    - [Using Mint Adapter](#using-mint-adapter)
```
polvalente: There are client docs here

grpc_core/mix.exs
sleipnir: Do you intend to publish this version of the libraries separately as version 0.12?

polvalente: yeah, I think we should keep 0.12, or at least 1.0.0-rc.1 if we're really sure we're ready to bump. We won't be able to make breaking changes after the bump

grpc_server/mix.exs
```diff
@@ -0,0 +1,152 @@
+defmodule GRPC.Server.MixProject do
+  use Mix.Project
+
+  @source_url "https://github.com/elixir-grpc/grpc"
+  @version "1.0.0"
```
polvalente: let's not leave 1.0.0 in these modules. If any publishing ends up happening we'll lose this version on Hex

grpc_server/test/grpc/server/supervisor_test.exs
```diff
@@ -0,0 +1,83 @@
...
+
+      # assert {:ok,
+      #         {%{strategy: :one_for_one},
+      #          [
+      #            %{
+      #              id: {:ranch_listener_sup, ^endpoint_str},
+      #              start: _,
+      #              type: :supervisor
+      #            }
+      #          ]}} = Supervisor.init(endpoint: MockEndpoint, port: 1234, start_server: true)
```
polvalente: stray comment

interop/.formatter.exs
```diff
@@ -0,0 +1,7 @@
+[
```
polvalente, line 1: how about we merge interop into benchmarking or vice-versa, so minimize the amount of auxiliary directories?

grpc_client/mix.exs
polvalente: I mean the fact that Channel and Stub have stayed with the client lib

benchmark/lib/benchmark/client_worker.ex
polvalente: I meant application.ex itself. This type of start_link is a bit of an antipattern and users will likely look at this as an example lib user project

interop/.formatter.exs
sleipnir: I prefer to deal with this in another PR.

PR #480, opened 2025-12-05
Fix Connection state created by build_direct_state

ypconstante:
`build_direct_state` was storing the channel in `real_channels` map instead of tuples like it is done in `build_real_channels`, causing connection refresh and disconnect to fail.


sleipnir (approved):
LGTM

PR #481, opened 2025-12-08
Support grpcweb trailers encoded in the message

aseigo:
As the Fetch API does not expose HTTP trailers to the Javascript runtime, grpcweb mandates that tailers are included in the message payload with the most-significant bit of the leading byte (flags) set to 1 followed by a run-length-encoded block of text that follows the same formatting as normal headers.

Most extant grpcweb libraries written in JS/TS are lenient about this and will happily forego receiving trailers. However, some are more picky about this and REQUIRE trailers (the buf.read Connect libraries are an example of this).

With this changset:

`GRPC.Server` follows the spec when sending protos over grpcweb, allowing errors and other custom trailers to be sent in a way that is visible to the client.

`GRPC.Message` recognizes trailers and parses them appropriately: it extracts partial-buffer messages using the run-length encoding bytes (which it was previously quietly ignoring, which would also allow malformed buffers due to e.g. network problems sneak through anyways), it respects the trailers flag, and returns appropriate data in each of these cases.

The GRPC client now also works with embedded trailers.

Overhead for non-grpcweb should be nominal as new code paths are hidden behind grpcweb checks, while the additional binary checks are placed in front the error paths (so errors may be nominally slower to be reached, but the happy paths should be untouched).

This has been tested with both Google's own `grpc-web` lbirary as well as buf.build's `connect`/`connect-web` libraries with a real-world API being served by elixir-grpc's grpc libraries. 

This does need more testing (what doesn't!), and there are some decisions made in the details of the code that could be discussed.


grpc_core/lib/grpc/message.ex
```diff
@@ -11,9 +11,11 @@ defmodule GRPC.Message do
     Message -> *{binary octet}
   """
 
+  import Bitwise
```
polvalente, line 14:
```suggestion
  use Bitwise, only_operators: true
```

Stylistic nit: I'd rather just import the operators as those are generally easily visually recognizable as being imported from Bitwise. The function identifiers not so much.

grpc_core/lib/grpc/message.ex
```diff
@@ -78,12 +81,14 @@ defmodule GRPC.Message do
   ## Examples
 
       iex> GRPC.Message.from_data(<<0, 0, 0, 0, 8, 1, 2, 3, 4, 5, 6, 7, 8>>)
-      <<1, 2, 3, 4, 5, 6, 7, 8>>
+      {<<1, 2, 3, 4, 5, 6, 7, 8>>, <<>>}
   """
-  @spec from_data(binary) :: binary
+  @spec from_data(binary) :: {message :: binary, rest :: binary}
```
polvalente, line 86: This is technically a breaking change. I don't think it impacts as much if we merge before releasing 1.0, but we need to keep this in mind.

grpc_core/lib/grpc/message.ex
```diff
@@ -107,24 +115,39 @@ defmodule GRPC.Message do
...
       _ ->
         {:error, RPCError.exception(status: :invalid_argument, message: "Message is malformed.")}
     end
   end
 
+  defp parse_trailers(data) do
+    data
+    |> String.split("\r\n")
+    |> Enum.reduce(%{}, fn line, acc ->
+      [k, v] = String.split(line, ":")
```
polvalente:
```suggestion
      [k, v] = String.split(line, ":", parts: 2)
```

Otherwise this can raise

grpc_core/lib/grpc/message.ex
```diff
@@ -107,24 +115,39 @@ defmodule GRPC.Message do
...
+      Map.put(acc, k, String.trim(v))
```
polvalente, line 147: Do we need to trim the key as well?

aseigo: Yes, otherwise we end up with unintended spaces, such as " 0" which is intended to represent a number such as in the `grpc-status` headers. Parsing that as an integer is no longer trivially possible using either `String.to_integer` or `Integer.parse`. The code that does this check for proper HTTP trailers would then need to have a `String.trim` added to it in order to work, which would penalize *all* such checks. :/

grpc_core/lib/grpc/message.ex
aseigo: see 054f370

grpc_core/lib/grpc/message.ex
aseigo: Unfortunately on newer versions of Elixir we get this during compile: `warning: use Bitwise is deprecated. import Bitwise instead`

... which is why I removed the `use`, and used an `import` even though it does not support the `only_operators` option :/

I can put it back as `Bitwise.bsl` if you prefer, of course, but the `import` should probably stay.

polvalente: I personally prefer avoiding imports as much as possible to reduce compilation time, but given this is just an stdlib module we can keep it as is.

I must have read an outdated version of the docs.

polvalente: Just found it in history. Apparently Google sent me to 1.12 and I didn't check that.

grpc_core/lib/grpc/message.ex
polvalente: I meant to ask if we needed:

```suggestion
      Map.put(acc, String.trim(k), String.trim(v))
```

Agreed with the reasoning for trimming the value :)

aseigo: Ah! I see, I misread ... blah.

I don't think it is needed, as the BNF documentation of headers all deny any use of spaces before the colon with optional whitespace only being allowed *after* the colon:

```
      h1-common-structure-header =
             h1-common-structure-legacy-header /
             h1-common-structure-self-identifying-header

     h1-common-structure-legacy-header =
             field-name ":" OWS h1-common-structure

     h1-common-structure-self-identifying-header:
             field-name ":" OWS ">" h1-common-structure "<"
```

(Where "OWS" is "optional white space".)

Or, in standardese: "Each field line consists of a case-insensitive field name followed by a colon (":"), optional leading whitespace, the field line value, and optional trailing whitespace." (see https://datatracker.ietf.org/doc/html/rfc9112 ).

We could add a trim to the field name in case of non-conformant implementations, but I wouldn't bother as it should never actually occur as per the standard.

polvalente: Alright, let's keep this as is then! Thanks for the clarification

polvalente: Alright, let's keep this as is then! Thanks for the clarification

polvalente:
> > Great! So the decision is whether we want to accept "incorrect" payloads regardless when grpcweb-formatted data is sent outside that scope. I'm ok with just keeping the single code path.
> 
> That's a good question.
> 
> It would probably mean having to add another param to `GRPC.Message.get_message/1` (e.g. the codec), and in practice this should only appear when the source is a server in grpc_web mode which means it would be another check that doesn't do anything in practice ... so personally I don't think even more API changes is really worth it?

Yeah, I'm leaning towards that as well. @sleipnir do you agree?

sleipnir:
@aseigo Thanks for the PR, excellent work.

I'm just still checking which PR will go into the main branch first.

aseigo:
Working on another fix for GRPC errors (general problem, not related to webrpc), I refactored this a small amount, pulling the grpcweb trailer handling into its own function. Same code, no functional changes, just a bit cleaner and easier to work around. Hopefully it will also make it a bit clearer what would need to be done in the new adapter as well?

Cheers!

byhemechi:
Does you have a very rough timeframe for when this is likely to be merged? i need to make some architectural decisions for a product using the buf grpc library and need to know whether i'll need to get envoy onto developer environments.

Appreciate that hard work you guys put in presumably for free 🙏, please don't take this as a "hurry up" message

aseigo:
> Does you have a very rough timeframe for when this is likely to be merged? i need to make some architectural decisions for a product using the buf grpc library and need to know whether i'll need to get envoy onto developer environments.

I don't know when it will be merged exactly (that's someone else's decision), but I have some work to do on this tomorrow (a pair of tests are not passing), and then it should hopefully be ready.

If the pressing need is for development envs, you can always include my fork with this branch as a dependency in the short term until this does get merged. That is what I am currently doing, in fact (also working with the buf.build connect libs)

sleipnir:
> Does you have a very rough timeframe for when this is likely to be merged? i need to make some architectural decisions for a product using the buf grpc library and need to know whether i'll need to get envoy onto developer environments.
> 
> Appreciate that hard work you guys put in presumably for free 🙏, please don't take this as a "hurry up" message

Once the tests pass, We will merge it.

@aseigo, let me know if I can help with anything else. I'll handle merging this later in the PR for the new adapter.

aseigo:
> let me know if I can help with anything else

Thanks, I appreciate it! I think it's all good, though ... I just need to work out those last two tests, which I will do tomorrow.  :)

byhemechi:
> If the PR is itself ready, you can always use a git dependency in Mix: {:grpc, github: "aseigo", branch: "feature/grpcweb-trailers-in-message"} or :ref instead of :branch if you want to point to a specific commit hash.

Yep, that's what I've been doing, was just wondering about the package because the reset of my team is very new to elixir and I don't trust them to understand how git deps work

Strangely, this gives me a bunch of dependency errors when I compile deps, everything seems to work just fine if I manually add the package subdirs. Almost certainly a layer 8 issue on my end and if things merge soon I won't put any thought into it.

Thanks for the quick replies :)

.gitignore
```diff
@@ -23,6 +23,6 @@ erl_crash.dump
 /log
 
 .elixir_ls
-
+.expert
```
polvalente: Can we revert this line? ElixirLS is only there due to inertia and should be removed too. I don't think external tooling like this should be added to the project's gitignore, but to each user's global gitignore instead.

aseigo: see d627860

polvalente: Thanks!

sleipnir:
> Ok, all tests are passing, and the errors are appearing both with native and grpcweb clients for me, including the improvements @sleipnir pushed the other day to make the `{:error, ...}` tuple return style work.
> 
> This is ready for a final review! :)

Thank you @aseigo 
@polvalente and I will discuss the best way to proceed, thank you again for your contribution.

aseigo:
Thanks! :) :heart: : :rocket:

PR #482, opened 2025-12-08
[Feat] new server adapter

sleipnir:
Hi everyone, I'd like to talk a little about what I'm currently working on here.

I'm almost finished implementing a new server-side adapter entirely based on Elixir using the `thousand_island` library. And the results are promising, as we can see below.

### Performance Metrics
| Metric                                  | Cowboy        | ThousandIsland | Improvement                     |
|-------------------------------------|-------------------|-----------------------|----------------------------------|
| Requests Processed          | 1,475             | 3,115                  | +111% (2.11x)                 |
| Total Time                           | 1.46s             | 0.86s                  | -41% (1.71x faster)         |
| Throughput                         | ~1,008 req/s  | ~3,637 req/s      | +261%                             |
| Minimum Latency               | 618µs            | 177µs                | -71% (3.5x faster)            |
| Maximum Latency              | 89.1ms          | 39.6ms              | -56% (2.25x faster)          |
| Average Latency                | ~991µs          | ~274µs              | -72% (3.6x faster)            |
| CPU Time (User)               | 3.12s             | 1.72s                 | -45%                                 |
| CPU Time (System)          | 0.50s             | 0.48s                 | -4%                                    |

### Analysis
Advantages of ThousandIsland:

* Dramatically higher throughput:
  Processed more than double the number of requests in the same test period.
* Significantly lower latencies:
  * Minimum latency 3.5x lower
  * Maximum latency 2.25x lower
  * Average latency approximately 3.6x lower
* CPU efficiency:
  Used 45% less CPU time (user time).
* Reduced total execution time:
  Completed the tests in almost half the time
 
The ThousandIsland adapter demonstrates substantially superior performance compared to `Cowboy` across all measured aspects. The implementation delivers:

* Better throughput (~3.6x)
* Lower latencies (3–4x)
* Better resource efficiency

This is still a draft and needs a lot of refinement. I opened the PR just to document the work and share it with everyone.


benchmark/bin/profile.exs
```diff
@@ -0,0 +1,87 @@
...
+
+# Start the server
+server_config =
+  Grpc.Testing.ServerConfig.new(
+    async_server_threads: 1,
+    port: 10000
+  )
+
+{:ok, _server_pid} = Benchmark.ServerManager.start_server(server_config)
+Process.sleep(1000)
```
polvalente: Why is this needed?

benchmark/bin/profile.exs
```diff
@@ -0,0 +1,87 @@
...
+
+# Connect client
+{:ok, ch} = GRPC.Stub.connect("localhost:10000")
+
+payload_type = Grpc.Testing.PayloadType.value(:COMPRESSABLE)
```
polvalente:
```suggestion
payload_type = Grpc.Testing.PayloadType.value(:COMPRESSIBLE)
```

minor nit: I don't think compressable is gramatically correct

.gitignore
```diff
@@ -25,5 +25,8 @@ erl_crash.dump
 .elixir_ls
 
 .elixir_tools
+grpc_client/.elixir-tools/
+grpc_core/.elixir-tools/
+grpc_server/.elixir-tools/
```
polvalente:
```suggestion
```

Same comment as usual, you should be adding these to your global gitignore

benchmark/config/config.exs
```diff
@@ -1,5 +1,6 @@
 import Config
 
-config :logger, level: :info
+# Disable logging by default for better performance
```
polvalente:
```suggestion
# Reduce logging overhead by default for better performance
```

Although I don't think this is actually true. You might be measuring the wrong thing if logging has this much of an impact. I'm curious about the difference in performance.

benchmark/config/config.exs
```diff
@@ -1,5 +1,6 @@
...
+config :logger, level: :error
 
 import_config "#{Mix.env()}.exs"
```
polvalente:
```suggestion
import_config "#{config_env()}.exs"
```

benchmark/bin/profile.exs
sleipnir: I removed the file.

benchmark/bin/profile.exs
sleipnir: I removed the file.

benchmark/config/config.exs
sleipnir: The development of this feature was entirely based on benchmarks; I don't remember the exact numbers now, but the impact was significant when running with logs. However, given the numerous modifications, anything could have been the case. I don't think this will impact the PR, and we can continuously test it even after release.

benchmark/profile_script.exs
```diff
@@ -0,0 +1,34 @@
...
+{:ok, _} = Application.ensure_all_started(:grpc_server)
+{:ok, _} = Application.ensure_all_started(:benchmark)
+
+server = %Grpc.Testing.ServerConfig{
+  async_server_threads: 1,
+  port: 10000
+}
+
+_server = Benchmark.ServerManager.start_server(server, adapter: GRPC.Server.Adapters.ThousandIsland)
+Process.sleep(500)
```
polvalente: Why is this needed? I believe I asked a similar question on a file that has been removed.

benchmark/profile_script.exs
```diff
@@ -0,0 +1,34 @@
...
+
+# Connect client
+{:ok, ch} = GRPC.Stub.connect("localhost:10000")
+
+# Prepare request
+payload_type = Grpc.Testing.PayloadType.value(:COMPRESSABLE)
```
polvalente:
```suggestion
payload_type = Grpc.Testing.PayloadType.value(:COMPRESSIBLE)
```

If we have control over the values, we need to fix spelling

grpc_client/test/grpc/integration/thousand_island_adapter_test.exs
```diff
@@ -0,0 +1,230 @@
...
+        {:ok, channel} = GRPC.Stub.connect("localhost:#{port}")
+
+        request = %Helloworld.HelloRequest{name: "ThousandIsland"}
+        {:ok, response} = channel |> Helloworld.Greeter.Stub.say_hello(request)
+
+        assert response.message == "Hello ThousandIsland!"
+
+        GRPC.Stub.disconnect(channel)
+      end)
+    end
```
polvalente, line 94: What exactly about this test (the file as a whole) makes it thousandisland-specific?

grpc_core/lib/grpc/message.ex
```diff
@@ -15,6 +15,9 @@ defmodule GRPC.Message do
 
   @max_message_length Bitwise.bsl(1, 32 - 1)
 
+  # Inline hot path functions
+  @compile {:inline, to_data: 2, from_data: 1}
```
polvalente: what's the impact of having this vs not having?

polvalente: The comment should explain more on the reasoning behind inlining

grpc_core/lib/grpc/message.ex
```diff
@@ -42,33 +45,41 @@ defmodule GRPC.Message do
...
 
-    length = IO.iodata_length(message)
+    length = IO.iodata_length(compressed_message)
+
+    case length > max_length do
+      true ->
+        {:error, "Encoded message is too large (#{length} bytes)"}
+
+      false ->
+        result = [compress_flag, <<length::size(4)-unit(8)>>, compressed_message]
```
polvalente: please rewrite to if-else

grpc_core/lib/grpc/message.ex
```diff
@@ -42,33 +45,41 @@ defmodule GRPC.Message do
...
+                do: codec.pack_for_channel(result),
+                else: result
+          end
 
-      result = if iolist, do: result, else: IO.iodata_to_binary(result)
+        result =
+          case opts[:iolist] do
+            true -> result
+            _ -> IO.iodata_to_binary(result)
+          end
```
polvalente: let's use if-else here too

grpc_core/lib/grpc/transport/http2/frame.ex
```diff
@@ -0,0 +1,104 @@
...
+      0x6 -> GRPC.Transport.HTTP2.Frame.Ping.deserialize(flags, stream_id, payload)
+      0x7 -> GRPC.Transport.HTTP2.Frame.Goaway.deserialize(flags, stream_id, payload)
+      0x8 -> GRPC.Transport.HTTP2.Frame.WindowUpdate.deserialize(flags, stream_id, payload)
+      0x9 -> GRPC.Transport.HTTP2.Frame.Continuation.deserialize(flags, stream_id, payload)
+      _unknown -> GRPC.Transport.HTTP2.Frame.Unknown.deserialize(type, flags, stream_id, payload)
+    end
+    |> case do
+      {:ok, frame} -> {{:ok, frame}, rest}
+      {:error, error_code, reason} -> {{:error, error_code, reason}, rest}
+    end
```
polvalente: let's rewrite this so the first case is a separate defp. Legibility is hurt by piping 2 cases

polvalente: also let's use 0 instead 0x0, 9 instead of 0x9 and so on

grpc_core/lib/grpc/transport/http2/frame.ex
```diff
@@ -0,0 +1,104 @@
...
+    {{:more, msg}, <<>>}
+  end
+
+  defmodule Flags do
+    @moduledoc false
+
+    import Bitwise
+
+    defguard set?(flags, bit) when band(flags, bsl(1, bit)) != 0
+    defguard clear?(flags, bit) when band(flags, bsl(1, bit)) == 0
```
polvalente:
```suggestion
```

remove this and use `not set?` for legibility

grpc_core/lib/grpc/transport/http2/frame.ex
```diff
@@ -0,0 +1,104 @@
...
+
+    @spec set([0..255]) :: 0..255
```
polvalente:
```suggestion
    @spec set([non_neg_integer()]) :: non_neg_integer() 
```

The spec was expecting the range, not a list of numbers contained in said range.

polvalente: Also, shouldn't the input range actually be [0, 7] instead? The input implies 256 bits for the output, but the output range you spec's implies 8 bits

grpc_core/lib/grpc/transport/http2/frame.ex
```diff
@@ -0,0 +1,104 @@
...
+    def set([]), do: 0x0
+    def set([bit | rest]), do: bor(bsl(1, bit), set(rest))
```
polvalente: Let's rewrite into a reduce, as this one is not using tail recursion:

`Enum.reduce(bits, 0, fn bit, acc -> acc + 2 ** bit end)`

grpc_core/lib/grpc/transport/http2/frame.ex
```diff
@@ -0,0 +1,104 @@
...
+  # nil is used to indicate for Stream.unfold/2 that the frame deserialization is finished
+  def deserialize(<<>>, _max_frame_size) do
+    nil
+  end
+
+  def deserialize(msg, _max_frame_size) do
+    {{:more, msg}, <<>>}
+  end
+
+  defmodule Flags do
```
polvalente: Let's extract this to its own file to help with discoverability

grpc_core/lib/grpc/transport/http2/frame/continuation.ex
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  defimpl GRPC.Transport.HTTP2.Frame.Serializable do
+    @end_headers_bit 2
+
+    def serialize(%GRPC.Transport.HTTP2.Frame.Continuation{} = frame, max_frame_size) do
+      fragment_length = IO.iodata_length(frame.fragment)
+
+      if fragment_length <= max_frame_size do
+        [{0x9, set([@end_headers_bit]), frame.stream_id, frame.fragment}]
```
polvalente: As a general comment, let's avoid 0x notation as much as possible

grpc_core/lib/grpc/transport/http2/frame/headers.ex
```diff
@@ -0,0 +1,147 @@
...
+        ) do
+      flags = if frame.end_stream, do: [@end_stream_bit], else: []
+
+      fragment_length = IO.iodata_length(frame.fragment)
+
+      if fragment_length <= max_frame_size do
+        [{0x1, set([@end_headers_bit | flags]), frame.stream_id, frame.fragment}]
+      else
+        <<this_frame::binary-size(max_frame_size), rest::binary>> =
+          IO.iodata_to_binary(frame.fragment)
```
polvalente, line 132: I suspect there's room for keeping this as iodata a while longer and do a single iodata_to_binary after processing all frames

grpc_core/lib/grpc/transport/http2/frame/ping.ex
```diff
@@ -0,0 +1,49 @@
...
+  @ack_bit 0
+
+  @spec deserialize(
+          GRPC.Transport.HTTP2.Frame.flags(),
+          GRPC.Transport.HTTP2.Stream.stream_id(),
+          iodata()
+        ) ::
+          {:ok, t()} | {:error, GRPC.Transport.HTTP2.Errors.error_code(), binary()}
+  def deserialize(flags, 0, <<payload::binary-size(8)>>) when set?(flags, @ack_bit) do
+    {:ok, %__MODULE__{ack: true, payload: payload}}
```
polvalente: I think this would be more readable if `ack: set?(flags, @ack_bit)` was used in just 1 clause. Right now you're both selecting on stream_id == 0 and the flag and this hurts legibility

grpc_core/lib/grpc/transport/http2/frame.ex
```diff
@@ -0,0 +1,104 @@
+defmodule GRPC.Transport.HTTP2.Frame do
```
polvalente, line 1: Why are these in grpc_core? Are they used by both client and server? I expected to see something change in grpc_client to leverage this

grpc_server/lib/grpc/server/adapters/thousand_island.ex
```diff
@@ -0,0 +1,639 @@
...
+
+  ThousandIsland is a modern, pure Elixir socket server that provides:
+  - Built-in connection pooling
+  - Efficient resource management  
+  - Better integration with Elixir/OTP ecosystem
+  - Simpler architecture than Cowboy/Ranch
+
+  ## Advantages over Cowboy
+
+  - **Pure Elixir**: No Erlang dependencies, better integration
```
polvalente: I wouldn't call this an advantage

grpc_server/lib/grpc/server/adapters/thousand_island.ex
```diff
@@ -0,0 +1,639 @@
...
+      ├── Connection State (HTTP2.Connection)
+      │   └── Stream States (HTTP2.StreamState per stream_id)
+      └── User Handler Tasks (spawned per RPC)
+          └── BidiStream GenServer (only for bidi streaming)
+  ```
+
+  ### 1. Unary RPC (request -> response)
+
+  #### Request Path
+  1. **Client sends HTTP/2 frames** → TCP socket
```
polvalente, line 77:
```suggestion
  #### Request Path

  1. **Client sends HTTP/2 frames** → TCP socket
```

grpc_server/lib/grpc/server/adapters/thousand_island.ex
```diff
@@ -0,0 +1,639 @@
...
+  4. **Connection.process_grpc_request/4** extracts complete request
+     - Decodes 5-byte length-prefixed message
+     - Looks up RPC method from path
+  5. **Dispatcher.dispatch/4** routes to handler
+     - Calls `Dispatcher.call_unary/5`
+     - Directly invokes user handler function: `MyServer.my_method(request, stream)`
+     - Handler runs **synchronously** in Handler process
+
+  #### Response Path
+  1. **Handler returns response** (or calls `GRPC.Server.send_reply/2`)
```
polvalente, line 94:
```suggestion
  #### Response Path

  1. **Handler returns response** (or calls `GRPC.Server.send_reply/2`)
```

grpc_server/lib/grpc/server/adapters/thousand_island.ex
```diff
@@ -0,0 +1,639 @@
...
+     - Sets END_STREAM flag
+  5. **Connection.send_trailers/4** encodes final HEADERS frame
+     - Sets END_HEADERS + END_STREAM flags
+
+  **Process Model**: Single Handler process handles entire request synchronously
+
+  ### 2. Client Streaming RPC (stream of requests -> response)
+
+  #### Request Path
+  1. **Client sends multiple DATA frames** (END_STREAM on last)
```
polvalente, line 113:
```suggestion
  #### Request Path

  1. **Client sends multiple DATA frames** (END_STREAM on last)
```

grpc_server/lib/grpc/server/adapters/thousand_island.ex
```diff
@@ -0,0 +1,639 @@
...
+     - Messages accumulated in StreamState.message_buffer
+  3. **When END_STREAM received** → **process_grpc_request/4**
+     - All messages decoded
+  4. **Dispatcher.call_client_streaming/5**
+     - Creates `Stream.unfold` from buffered messages
+     - Calls handler: `MyServer.my_method(request_enum, stream)`
+     - Handler **synchronously** consumes stream
+
+  #### Response Path
+  Same as Unary (single response at end)
```
polvalente, line 126:
```suggestion
  #### Response Path

  Same as Unary (single response at end)
```

grpc_server/lib/grpc/server/adapters/thousand_island.ex
```diff
@@ -0,0 +1,639 @@
...
+
+  **Process Model**: Single Handler process, synchronous handler execution
+
+  ### 3. Server Streaming RPC (request -> stream of responses)
+
+  #### Request Path
+  Same as Unary (single request)
+
+  #### Response Path
+  1. **Handler calls `GRPC.Server.send_reply/2` multiple times**
```
polvalente, line 138:
```suggestion
  #### Request Path

  Same as Unary (single request)

  #### Response Path

  1. **Handler calls `GRPC.Server.send_reply/2` multiple times**
```

grpc_server/lib/grpc/server/adapters/thousand_island.ex
```diff
@@ -0,0 +1,639 @@
...
+  User Handler Task (#PID<0.XXX.0>)
+  ├── Runs: MyServer.full_duplex_call(request_enum, stream)
+  ├── Consumes: request_enum (lazy, pulls from BidiStream)
+  └── Sends: {:grpc_send_data, ...} messages to Handler
+
+  BidiStream GenServer (#PID<0.YYY.0>)
+  ├── Queue: Buffered incoming requests
+  ├── Receives: {:add_message, msg} from Handler (when DATA arrives)
+  └── Provides: Lazy enumerable to User Task
+  ```
```
polvalente, line 179: If we're keeping these, these should be Mermaid-written sequence diagrams

grpc_server/lib/grpc/server/adapters/thousand_island.ex
```diff
@@ -0,0 +1,639 @@
...
+  **Why this works**:
+  - Handler process is single-threaded message loop
+  - Messages processed in order: accumulate_headers → send_data → send_trailers
+  - Accumulated headers guaranteed to be merged before first DATA
+  - User Task sends messages asynchronously, Handler serializes them
+
+  ## Message Flow Diagram (Bidi Streaming)
+
+  ```
+  Client                Handler Process              User Task              BidiStream
```
polvalente: let's rewrite this with mermaid

grpc_server/lib/grpc/server/adapters/thousand_island.ex
```diff
@@ -0,0 +1,639 @@
...
+  def start(endpoint, servers, port, opts) do
+    GRPC.Server.Cache.init()
+
+    case Process.whereis(GRPC.Server.StreamTaskSupervisor) do
+      nil ->
+        {:ok, _} = Task.Supervisor.start_link(name: GRPC.Server.StreamTaskSupervisor)
+
+      _pid ->
+        :ok
+    end
```
polvalente, line 389: Add this to the application tree

grpc_server/lib/grpc/server/adapters/thousand_island.ex
```diff
@@ -0,0 +1,639 @@
...
+
+  ## Options
+    * `:ip` - The IP to bind the server to (default: listen on all interfaces)
+    * `:port` - The port to listen on (required)
+    * `:num_acceptors` - Number of acceptor processes (default: 100)
+    * `:num_connections` - Maximum concurrent connections (default: 16384)
+    * `:transport_options` - Additional transport options to pass to ThousandIsland
+  """
+  @impl true
+  def start(endpoint, servers, port, opts) do
```
polvalente, line 379: this should be start_link

it's a bit misleading a start function that links a process (or 2 in the current version)

grpc_server/lib/grpc/server/adapters/thousand_island.ex
```diff
@@ -0,0 +1,639 @@
...
+
+  @doc """
+  Return a child_spec to start server under a supervisor.
+  """
+  @spec child_spec(atom(), %{String.t() => [module()]}, non_neg_integer(), Keyword.t()) ::
+          Supervisor.child_spec()
+  def child_spec(endpoint, servers, port, opts) do
+    server_opts = build_server_opts(endpoint, servers, port, opts)
+
+    GRPC.Server.Cache.init()
```
polvalente: move this to application.ex

grpc_server/lib/grpc/server/adapters/thousand_island.ex
```diff
@@ -0,0 +1,639 @@
...
+
+  def set_headers(%{handler_pid: pid, stream_id: stream_id}, headers) do
+    send(pid, {:grpc_accumulate_headers, stream_id, headers})
+    :ok
+  end
+
+  def set_resp_trailers(%{handler_pid: _pid, stream_id: stream_id}, trailers) do
+    # Store in process dictionary (runs in handler context during dispatch)
+    current_custom_trailers = Process.get({:grpc_custom_trailers, stream_id}, %{})
+    Process.put({:grpc_custom_trailers, stream_id}, Map.merge(current_custom_trailers, trailers))
```
polvalente: It would be nice if instead of using a process dict we could have a genserver with explicit state

polvalente: If this is always called by the handler, let's have it only live there

grpc_server/lib/grpc/server/adapters/thousand_island.ex
```diff
@@ -0,0 +1,639 @@
...
+  @impl true
+  def send_reply(%{handler_pid: pid, stream_id: stream_id}, data, opts) do
+    # Encode message with gRPC framing (compressed flag + length + data)
+    compressor = Keyword.get(opts, :compressor)
+    codec = Keyword.get(opts, :codec)
+
+    case GRPC.Message.to_data(data, compressor: compressor, codec: codec, iolist: true) do
+      {:ok, framed_data, _size} ->
+        # Send data frame - handler will send accumulated headers first if needed
+        send(pid, {:grpc_send_data, stream_id, framed_data})
```
polvalente, line 535: we should have the handler be a genserver and this be a genserver call

polvalente: all of these fire-and-forget sends should become calls

grpc_server/lib/grpc/server/adapters/thousand_island.ex
```diff
@@ -0,0 +1,639 @@
...
+    case Keyword.get(adapter_opts, :ip) do
+      nil -> transport_opts
+      ip -> Keyword.put(transport_opts, :ip, ip)
+    end
+  end
+
+  defp maybe_add_ssl(transport_opts, nil), do: transport_opts
+
+  defp maybe_add_ssl(transport_opts, cred_opts) do
+    transport_opts ++ cred_opts.ssl
```
polvalente: Keword.merge

grpc_server/lib/grpc/server/adapters/thousand_island.ex
```diff
@@ -0,0 +1,639 @@
...
+  defp transport_module(opts) do
+    if cred_opts(opts) do
+      ThousandIsland.Transports.SSL
+    else
+      ThousandIsland.Transports.TCP
+    end
+  end
+
+  defp cred_opts(opts) do
+    Kernel.get_in(opts, [:cred])
```
polvalente:
```suggestion
    opts[:cred]
```

grpc_server/lib/grpc/server/adapters/thousand_island.ex
```diff
@@ -0,0 +1,639 @@
...
+  end
+
+  defp servers_name(nil, servers) do
+    servers |> Map.values() |> Enum.map(fn s -> inspect(s) end) |> Enum.join(",")
```
polvalente:
```suggestion
    Enum.map_join(servers, ",", fn _k, s -> inspect(s) end)
```

grpc_server/lib/grpc/server/adapters/thousand_island.ex
```diff
@@ -0,0 +1,639 @@
...
+    else
+      ThousandIsland.Transports.TCP
+    end
+  end
+
+  defp cred_opts(opts) do
+    Kernel.get_in(opts, [:cred])
+  end
+
+  defp servers_name(nil, servers) do
```
polvalente: let's rename to servers_names or server_names

grpc_server/lib/grpc/server/adapters/thousand_island/handler.ex
```diff
@@ -0,0 +1,284 @@
...
+
+  # Inline hot path functions
+  @compile {:inline, handle_data: 3, handle_preface: 3, handle_frames_loop: 5}
+
+  @impl ThousandIsland.Handler
+  def handle_connection(socket, handler_options) do
+    Logger.debug("New HTTP/2 connection established - socket: #{inspect(socket)}")
+
+    # Initialize ETS cache for codecs/compressors lookup  
+    GRPC.Server.Cache.init()
```
polvalente, line 25: move to application.ex

grpc_server/lib/grpc/server/adapters/thousand_island/handler.ex
```diff
@@ -0,0 +1,284 @@
...
+
+    # Support both keyword list and map formats
+    {servers_list, endpoint, opts} =
+      if is_map(handler_options) do
+        {Map.get(handler_options, :servers, []), Map.get(handler_options, :endpoint),
+         Map.get(handler_options, :opts, [])}
+      else
+        {Keyword.get(handler_options, :servers, []), Keyword.get(handler_options, :endpoint),
+         Keyword.get(handler_options, :opts, [])}
+      end
```
polvalente, line 35: rewrite these to do handler_options[key] for each variable

polvalente: if performance is a concern, just standardize on map

grpc_server/lib/grpc/server/adapters/thousand_island/handler.ex
```diff
@@ -0,0 +1,284 @@
...
+
+    servers =
+      cond do
+        is_map(servers_list) and not is_struct(servers_list) -> servers_list
+        is_list(servers_list) -> GRPC.Server.servers_to_map(servers_list)
+        true -> %{}
+      end
```
polvalente, line 42: just enforce that the map should be passed directly

grpc_server/lib/grpc/server/adapters/thousand_island/handler.ex
```diff
@@ -0,0 +1,284 @@
...
+      preface_received: false,
+      accumulated_headers: %{}
+    }
+
+    {:continue, new_state}
+  end
+
+  @impl ThousandIsland.Handler
+  def handle_data(data, socket, %{preface_received: false, buffer: buffer} = state) do
+    new_buffer = buffer <> data
```
polvalente, line 64: let's accumulate as an iolist [buffer, data]

grpc_server/lib/grpc/server/adapters/thousand_island/handler.ex
```diff
@@ -0,0 +1,284 @@
...
+  @impl ThousandIsland.Handler
+  def handle_error(reason, _socket, state) do
+    Logger.error("Connection error: #{inspect(reason)}")
+    {:close, state}
+  end
+
+  def handle_info({:grpc_accumulate_headers, stream_id, headers}, {socket, state}) do
+    current_headers = Map.get(state.accumulated_headers, stream_id, %{})
+    updated_headers = Map.merge(current_headers, headers)
+    new_accumulated = Map.put(state.accumulated_headers, stream_id, updated_headers)
```
polvalente, line 88: a clear case for update_in

grpc_server/lib/grpc/server/adapters/thousand_island/handler.ex
```diff
@@ -0,0 +1,284 @@
...
+    Logger.debug(
+      "[Handler] Updating stream_state for stream #{stream_id}, bidi_pid=#{inspect(updated_stream_state.bidi_stream_pid)}"
+    )
+
+    connection = state.connection
+
+    updated_connection = %{
+      connection
+      | streams: Map.put(connection.streams, stream_id, updated_stream_state)
+    }
```
polvalente, line 174: put_in

grpc_server/lib/grpc/server/adapters/thousand_island/handler.ex
```diff
@@ -0,0 +1,284 @@
...
+  end
+
+  defp handle_preface(buffer, _socket, state) when byte_size(buffer) < 24 do
+    # Wait for more data (preface is 24 bytes)
+    {:continue, %{state | buffer: buffer}}
+  end
+
+  defp handle_preface(<<@connection_preface, remaining::binary>>, socket, state) do
+    # Valid preface, initialize connection  
+    try do
```
polvalente, line 240: since this is the outer block, let's use the rescue for defp instead of try to remove indentation

grpc_server/lib/grpc/server/bidi_stream.ex
```diff
@@ -0,0 +1,168 @@
...
+      {:put_messages, new_messages} ->
+        Logger.info(
+          "[BidiStream #{state.stream_id}] Received #{length(new_messages)} new messages"
+        )
+
+        # Add messages to queue
+        new_queue =
+          Enum.reduce(new_messages, state.messages, fn msg, queue ->
+            :queue.in(msg, queue)
+          end)
```
polvalente, line 117:
```suggestion
          Enum.reduce(new_messages, state.messages, &:queue.in/2)
```

grpc_server/lib/grpc/server/cache.ex
```diff
@@ -0,0 +1,90 @@
+defmodule GRPC.Server.Cache do
+  @moduledoc """
+  ETS-based cache for frequently accessed server metadata.
```
polvalente, line 3: How frequently is this being written to? Does it make sense to use :persistent_term for better read performance?

grpc_server/lib/grpc/server/cache.ex
```diff
@@ -0,0 +1,90 @@
...
+  Improves performance by avoiding repeated Enum.find operations.
```
polvalente, line 4: If this data is scoped per connection, we should have one cache per connection. We get this benefit while also having better read peformance

grpc_server/lib/grpc/server/cache.ex
```diff
@@ -0,0 +1,90 @@
...
+    case :ets.lookup(@table_name, cache_key) do
+      [{^cache_key, rpc}] ->
+        rpc
+
+      [] ->
+        rpc_calls = server.__meta__(:service).__rpc_calls__()
+
+        rpc =
+          Enum.find(rpc_calls, nil, fn {name, _, _, _} ->
+            Atom.to_string(name) == method_name
```
polvalente, line 73: We should either only allow atom method names in the input or store string names in the key

grpc_server/lib/grpc/server/http2/connection.ex
```diff
@@ -0,0 +1,1064 @@
...
+  require Logger
+
+  alias GRPC.Transport.HTTP2.Frame
+  alias GRPC.Transport.HTTP2.Settings
+  alias GRPC.Transport.HTTP2.Errors
+  alias GRPC.Server.HTTP2.StreamState
+
+  # Inline hot path functions for performance
+  @compile {:inline,
+            extract_messages: 2, send_frame: 3, handle_headers_frame: 3, send_grpc_trailers: 4}
```
polvalente, line 14: We need to evaluate the impact of these inlines and avoid as many as possible

grpc_server/lib/grpc/server/http2/connection.ex
```diff
@@ -0,0 +1,1064 @@
...
+    )
+
+    {messages, remaining} = extract_messages(stream_state.data_buffer, [])
+    # Reverse since we build list backwards for performance
+    extracted_count = length(messages)
+
+    Logger.debug(
+      "[Connection] Extracted #{extracted_count} messages, #{byte_size(remaining)} bytes remaining"
+    )
+
```
polvalente, line 913:
```suggestion
    Logger.debug(fn ->
        extracted_count = length(messages)
      "[Connection] Extracted #{extracted_count} messages, #{byte_size(remaining)} bytes remaining"
    end)

    # Reverse since we build list backwards for performance
```

grpc_server/lib/grpc/server/http2/connection.ex
```diff
@@ -0,0 +1,1064 @@
...
+    end
+
+    %{connection | send_hpack_state: final_hpack}
+  end
+
+  defp extract_messages_from_buffer(stream_state) do
+    # Extract 5-byte length-prefixed messages from data_buffer
+    Logger.debug(
+      "[Connection] Extracting messages from data_buffer (#{byte_size(stream_state.data_buffer)} bytes)"
+    )
```
polvalente, line 904:
```suggestion
    Logger.debug(fn ->
      "[Connection] Extracting messages from data_buffer (#{byte_size(stream_state.data_buffer)} bytes)"
    end)
```

grpc_server/lib/grpc/server/http2/connection.ex
```diff
@@ -0,0 +1,1064 @@
...
+        Logger.debug(
+          "[SEND_GRPC_ERROR] stream=#{stream_id}, status=#{status_code}, message=#{message}, headers_sent=#{headers_sent}, end_stream_received=#{end_stream_received}"
+        )
+
+        # RFC 7540 Section 5.1: Se já enviamos END_STREAM (via headers_sent=true em resposta anterior),
+        # o stream está "closed" e NÃO podemos enviar mais frames (exceto PRIORITY)
+        # Nesse caso, apenas remover o stream e não enviar nada
+        if headers_sent && end_stream_received do
+          Logger.debug(
+            "[SEND_GRPC_ERROR] stream=#{stream_id} SKIPPED - stream is fully closed (both sides sent END_STREAM)"
```
polvalente, line 961: let's unnest these with cond

grpc_server/lib/grpc/server/http2/dispatcher.ex
```diff
@@ -0,0 +1,821 @@
...
+  ## Returns
+  - `{:ok, response_headers, response_data, trailers}` - Success
+  - `{:error, rpc_error}` - gRPC error to send to client
+  """
+  @spec dispatch(StreamState.t(), map(), atom(), GRPC.Server.HTTP2.Connection.t()) ::
+          {:ok, list(), binary(), map()} | {:error, GRPC.RPCError.t()}
+  def dispatch(%StreamState{} = stream_state, servers, endpoint, connection) do
+    Logger.debug(
+      "[dispatch] path=#{stream_state.path}, messages=#{length(stream_state.message_buffer)}"
+    )
```
polvalente, line 47:
```suggestion
    Logger.debug(fn ->
      "[dispatch] path=#{stream_state.path}, messages=#{length(stream_state.message_buffer)}"
    end)
```

grpc_server/lib/grpc/server/http2/dispatcher.ex
```diff
@@ -0,0 +1,821 @@
...
+    if StreamState.deadline_exceeded?(stream_state) do
+      now = System.monotonic_time(:microsecond)
+
+      Logger.debug(
+        "[dispatch] Deadline exceeded for path=#{stream_state.path}, deadline=#{stream_state.deadline}, now=#{now}, diff=#{now - (stream_state.deadline || now)}us"
+      )
+
+      {:error, GRPC.RPCError.exception(status: :deadline_exceeded, message: "Deadline exceeded")}
+    else
+      with {:ok, service_name, method_name} <- parse_path(stream_state.path),
```
polvalente, line 59: extract this else to a defp do readability

grpc_server/lib/grpc/server/http2/dispatcher.ex
```diff
@@ -0,0 +1,821 @@
...
+        {:error, %GRPC.RPCError{} = error} ->
+          error
+
+        {:error, _reason} = err ->
+          Logger.error("Dispatch error: #{inspect(err)}")
+          {:error, GRPC.RPCError.exception(status: :internal, message: "Internal server error")}
+      end
+    end
+
+    # Fecha o if deadline_exceeded?
```
polvalente:
```suggestion
```

grpc_server/lib/grpc/server/http2/dispatcher.ex
```diff
@@ -0,0 +1,819 @@
...
+      end
+    end
+  end
+
+  defp decode_messages(message_buffer, rpc, codec, compressor) do
+    # Extract request type from RPC definition
+    # RPC format: {name, {request_module, is_stream?}, {reply_module, is_stream?}, options}
+    {_name, {request_module, _is_stream?}, _reply, _options} = rpc
+
+    try do
```
polvalente, line 222: use defp rescue to remove nesting

grpc_server/lib/grpc/server/http2/dispatcher.ex
```diff
@@ -0,0 +1,819 @@
...
+      codec ->
+        {:ok, codec}
+    end
+  end
+
+  defp get_compressor(server, metadata) do
+    # Check grpc-encoding header
+    encoding = Map.get(metadata, "grpc-encoding", "identity")
+
+    if encoding == "identity" do
```
polvalente, line 200: can we unnest this?

grpc_server/lib/grpc/server/http2/dispatcher.ex
```diff
@@ -0,0 +1,819 @@
...
+      base_headers =
+        if stream.compressor do
+          Map.put(base_headers, "grpc-encoding", stream.compressor.name())
+        else
+          base_headers
+        end
+
+      GRPC.Server.set_headers(stream, base_headers)
+
+      try do
```
polvalente, line 268: extract to a defp

grpc_server/lib/grpc/server/http2/dispatcher.ex
```diff
@@ -0,0 +1,819 @@
...
+
+      GRPC.Server.Adapters.ThousandIsland.send_trailers(stream.payload, error_trailers)
+      {:ok, :streaming_done}
+    end
+  end
+
+  defp call_client_streaming(server, func_name, requests, stream) do
+    # Check if function is implemented
+    if function_exported?(server, func_name, 2) do
+      try do
```
polvalente, line 344: extract to a defp

grpc_server/lib/grpc/server/http2/dispatcher.ex
```diff
@@ -0,0 +1,819 @@
...
+          if stream.compressor do
+            Map.put(base_headers, "grpc-encoding", stream.compressor.name())
+          else
+            base_headers
+          end
+
+        # Accumulate base headers without sending
+        GRPC.Server.set_headers(stream, base_headers)
+
+        # Convert list to stream
```
polvalente, line 362: stream to list? also, Enum.into -> Enum.to_list would communicate this and make this comment unnecessary

grpc_server/lib/grpc/server/http2/dispatcher.ex
```diff
@@ -0,0 +1,819 @@
...
+
+      GRPC.Server.Adapters.ThousandIsland.send_trailers(stream.payload, error_trailers)
+      {:ok, :streaming_done}
+    end
+  end
+
+  defp call_server_streaming(server, func_name, request, stream) do
+    # Check if function is implemented
+    if function_exported?(server, func_name, 2) do
+      try do
```
polvalente, line 426: extract to defp

grpc_server/lib/grpc/server/http2/dispatcher.ex
```diff
@@ -0,0 +1,819 @@
...
+  defp call_bidi_streaming(server, func_name, requests, stream, _connection) do
+    # Check if function is implemented
+    if function_exported?(server, func_name, 2) do
+      stream_id = stream.payload.stream_state.stream_id
+
+      Logger.info(
+        "[call_bidi_streaming] Starting bidi stream #{stream_id} with #{length(requests)} initial requests"
+      )
+
+      try do
```
polvalente, line 510: extract to defp

grpc_server/lib/grpc/server/http2/dispatcher.ex
```diff
@@ -0,0 +1,819 @@
...
+            base_headers
+          end
+
+        # Accumulate base headers without sending
+        GRPC.Server.set_headers(stream, base_headers)
+
+        # CRITICAL: Run handler in a separate task to not block the connection handler
+        # The connection handler MUST continue processing incoming DATA frames
+        # and feed them to the BidiStream while the handler is consuming messages
+        Task.start(fn ->
```
polvalente: Should we have a task supervisor? I don't like that this is just a dangling task with no links.

grpc_server/lib/grpc/server/http2/dispatcher.ex
```diff
@@ -0,0 +1,819 @@
...
+      # Build response headers
+      base_headers = [
+        {":status", "200"},
+        {"content-type", "application/grpc+proto"}
+      ]
+
+      # Only add grpc-encoding if there's actual compression (not identity)
+      headers =
+        if compressor do
+          base_headers ++ [{"grpc-encoding", compressor.name()}]
```
polvalente, line 746: can we prepend here?

grpc_server/lib/grpc/server/http2/dispatcher.ex
```diff
@@ -0,0 +1,819 @@
...
+
+    # Calculate length
+    length = byte_size(data_binary)
+
+    # Build 5-byte length-prefixed message
+    <<compressed_flag::8, length::32, data_binary::binary>>
+  end
+
+  defp encode_response(response, codec, compressor, stream_state) do
+    try do
```
polvalente, line 733: use defp rescue

grpc_server/lib/grpc/server/http2/dispatcher.ex
```diff
@@ -0,0 +1,819 @@
...
+      # Only add grpc-encoding if there's actual compression (not identity)
+      headers =
+        if compressor do
+          base_headers ++ [{"grpc-encoding", compressor.name()}]
+        else
+          base_headers
+        end
+
+      # Merge custom headers from handler (for custom_metadata test)
+      headers =
```
polvalente, line 752: why are headers a list while trailers are kept as a map below?

grpc_server/lib/grpc/server/http2/stream_state.ex
```diff
@@ -0,0 +1,360 @@
...
+            headers_sent: false,
+            codec: nil,
+            compressor: nil,
+            rpc: nil,
+            deadline: nil,
+            end_stream_received: false,
+            error_sent: false
+
+  # Parses grpc-timeout header and converts to absolute deadline
+  # Format: "1H" (hour), "1M" (minute), "1S" (second), "1m" (millisecond), "1u" (microsecond), "1n" (nanosecond)
```
polvalente, line 84: here is this format defined? Is 1H20M a valid value?

grpc_server/lib/grpc/server/http2/stream_state.ex
```diff
@@ -0,0 +1,360 @@
...
+  defp parse_timeout_to_deadline(nil), do: nil
+
+  defp parse_timeout_to_deadline(timeout_str) do
+    case Integer.parse(timeout_str) do
+      {value, "H"} -> System.monotonic_time(:microsecond) + value * 3_600_000_000
+      {value, "M"} -> System.monotonic_time(:microsecond) + value * 60_000_000
+      {value, "S"} -> System.monotonic_time(:microsecond) + value * 1_000_000
+      {value, "m"} -> System.monotonic_time(:microsecond) + value * 1_000
+      {value, "u"} -> System.monotonic_time(:microsecond) + value
+      {value, "n"} -> System.monotonic_time(:microsecond) + div(value, 1_000)
```
polvalente, line 94: use System.convert_time_unit instead to avoid magic numbers

polvalente: Also, let's convert, extract to a variable and have System.monotonic_time be called only at the end outside of the case

grpc_server/lib/grpc/server/http2/stream_state.ex
```diff
@@ -0,0 +1,360 @@
...
+      _ -> nil
```
polvalente, line 95: This should raise

grpc_server/lib/grpc/server/http2/stream_state.ex
```diff
@@ -0,0 +1,360 @@
...
+    %__MODULE__{
+      stream_id: stream_id,
+      state: :idle,
+      window_size: initial_window_size
+    }
+  end
+
+  @doc """
+  Creates a stream state from decoded HTTP/2 headers.
+  Extracts pseudo-headers and metadata from header list.
```
polvalente, line 113: what are pseudo headers?

grpc_server/lib/grpc/server/http2/stream_state.ex
```diff
@@ -0,0 +1,360 @@
...
+
+        {key, value}, {p, m, a, ct, t, meta} ->
+          {p, m, a, ct, t, Map.put(meta, key, value)}
+      end)
+
+    # Parse timeout and calculate deadline
+    deadline = parse_timeout_to_deadline(timeout_str)
+
+    if timeout_str do
+      require Logger
```
polvalente, line 148: move this to the top of the module

grpc_server/lib/grpc/server/http2/stream_state.ex
```diff
@@ -0,0 +1,360 @@
...
+
+      Logger.info(
+        "[StreamState] Parsed grpc-timeout: #{timeout_str} -> deadline: #{inspect(deadline)} (now: #{System.monotonic_time(:microsecond)})"
+      )
```
polvalente, line 152:
```suggestion
      Logger.info(fn ->
        "[StreamState] Parsed grpc-timeout: #{timeout_str} -> deadline: #{inspect(deadline)} (now: #{System.monotonic_time(:microsecond)})"
      end)
```

grpc_server/lib/grpc/server/http2/stream_state.ex
```diff
@@ -0,0 +1,360 @@
...
+
+      _other ->
+        {:error, :protocol_error}
+    end
+  end
+
+  @doc """
+  Processes DATA frame and accumulates gRPC messages.
+
+  gRPC uses 5-byte length-prefixed framing:
```
polvalente, line 207:
```suggestion
  gRPC uses run-length encoding:
```

grpc_server/lib/grpc/server/http2/stream_state.ex
```diff
@@ -0,0 +1,360 @@
...
+    content_type = Map.get(headers_map, "content-type")
+
+    # Validate gRPC request
+    with :ok <- validate_method(method),
+         :ok <- validate_path(path),
+         :ok <- validate_content_type(content_type) do
+      # Extract metadata (non-pseudo headers)
+      metadata =
+        headers_map
+        |> Enum.filter(fn {name, _value} -> !String.starts_with?(name, ":") end)
```
polvalente, line 287:
```suggestion
        |> Enum.reject(fn {name, _value} -> String.starts_with?(name, ":") end)
```

grpc_server/lib/grpc/server/http2/stream_state.ex
```diff
@@ -0,0 +1,360 @@
...
+    {:ok, stream, messages}
+  end
+
+  # Extract 5-byte length-prefixed messages
+  defp extract_messages(
+         <<compressed::8, length::32, payload::binary-size(length), rest::binary>>,
+         acc
+       ) do
+    message = %{compressed: compressed == 1, data: payload}
+    extract_messages(rest, acc ++ [message])
```
polvalente, line 353: can we prepend+reverse for this?

grpc_server/lib/grpc/server/stream.ex
```diff
@@ -98,10 +98,8 @@ defmodule GRPC.Server.Stream do
          data,
          opts
        ) do
-    opts =
-      opts
-      |> Keyword.put(:codec, codec)
-      |> Keyword.put(:http_transcode, access_mode == :http_transcoding)
+    # Optimize opts construction - avoid multiple Keyword operations
+    opts = [{:codec, codec}, {:http_transcode, access_mode == :http_transcoding} | opts]
```
polvalente, line 102: Revert this, as this is introducing a very subtle bug. We generally expect keys to be unique. If opts contains :codec or :http_transcode, we'll end up with a situation where Keyword.get will return the values set here, but if this list is converted to a map, then the original values are the ones kept.

polvalente (changes requested):
Blocking on the GRPC.Server.Stream opts bug

grpc_server/test/grpc/server/http2/frame_test.exs.bak
```diff
@@ -0,0 +1,259 @@
+defmodule GRPC.Server.HTTP2.FrameTest do
```
polvalente, line 1: stray file

grpc_client/test/grpc/integration/thousand_island_adapter_test.exs
sleipnir: Resolved!

grpc_core/lib/grpc/message.ex
sleipnir: done, i will send soon a new commit with this fix

grpc_core/lib/grpc/transport/http2/frame/headers.ex
sleipnir: Ok, but I feel safer not doing this right now.

grpc_core/lib/grpc/transport/http2/frame.ex
sleipnir: I didn't want generic http2 stuff on the server, I think it's better this way.

polvalente: The client apparently doesn't need to know anything about this abstraction, so it belongs in the server.

grpc_server/lib/grpc/server/adapters/thousand_island.ex
sleipnir: Since this is from the adapter API and we'll have to change it in Cowboy as well, I would leave this for after all the other review points. First, I want it to work as it is, then we'll change it to implement this improvement.

grpc_server/lib/grpc/server/adapters/thousand_island.ex
sleipnir: The server doesn't have an application; it uses a supervisor. Only the client has an application. I moved it to the supervisor.

polvalente: Then we should create one

polvalente: Then we should create one

grpc_server/lib/grpc/server/adapters/thousand_island.ex
sleipnir: This was intentional to avoid message exchange; it will most likely impact performance and the process model I envisioned if we change it. I would only do this if it justifies a performance improvement. In any case, I'll leave this for when I have a better-revised version (that is, I'm keeping this open for further review later).

grpc_server/lib/grpc/server/adapters/thousand_island.ex
sleipnir: This was deliberately avoided; we can review this further later, okay?

grpc_server/lib/grpc/server/adapters/thousand_island.ex
polvalente: If we're keeping this in a process dict, wrapping a genserver that does the same things but with state updates is most likely equivalent.

Generally using the process dictionary is a code smell

polvalente: If we're keeping this in a process dict, wrapping a genserver that does the same things but with state updates is most likely equivalent.

Generally using the process dictionary is a code smell

polvalente: If we're keeping this in a process dict, wrapping a genserver that does the same things but with state updates is most likely equivalent.

Generally using the process dictionary is a code smell

polvalente: In the worst case scenario, you're using the users own process dictionary, and then if things go wrong you might end up leaving garbage that impacts a follow-up call.

On the other hand, if this is always a process we own and not the user, then I don't see any reason to not turn it into a Genserver

polvalente: In the worst case scenario, you're using the users own process dictionary, and then if things go wrong you might end up leaving garbage that impacts a follow-up call.

On the other hand, if this is always a process we own and not the user, then I don't see any reason to not turn it into a Genserver

polvalente: In the worst case scenario, you're using the users own process dictionary, and then if things go wrong you might end up leaving garbage that impacts a follow-up call.

On the other hand, if this is always a process we own and not the user, then I don't see any reason to not turn it into a Genserver

polvalente: In the worst case scenario, you're using the users own process dictionary, and then if things go wrong you might end up leaving garbage that impacts a follow-up call.

On the other hand, if this is always a process we own and not the user, then I don't see any reason to not turn it into a Genserver

polvalente: In the worst case scenario, you're using the users own process dictionary, and then if things go wrong you might end up leaving garbage that impacts a follow-up call.

On the other hand, if this is always a process we own and not the user, then I don't see any reason to not turn it into a GenServer

grpc_server/lib/grpc/server/adapters/thousand_island.ex
polvalente: Deliberately avoiding a genserver call to use a send doesn't make any sense. With the send you don't have delivery guarantees and the the perf gain comes at the cost of correctness.

grpc_core/lib/grpc/transport/http2/frame/headers.ex
sleipnir: What I mean is that yes, I will review all the points for revision, but points that might be too much work I'll leave for the end, just to make it easier for myself. Rest assured, everything will be reviewed.

grpc_core/lib/grpc/transport/http2/frame.ex
sleipnir: Okay, I'll remove it to the server as agreed.

grpc_server/lib/grpc/server/adapters/thousand_island.ex
sleipnir: I'll leave this for another PR for now; I've already moved it to the Supervisor for convenience. It would be good to avoid impacting the devs, it would be a good change, but in any case, this PR will only come up in a 1.x.x update, but I would think about it carefully.

polvalente: Just think abou whether this supports more than 1 supervisor instance as is.

Given the lack of arguments, it seems to me it should be called in application.ex

grpc_server/lib/grpc/server/adapters/thousand_island.ex
sleipnir: I eliminated the dictionary and became much more defensive in my use of `send`, but I kept the process model. This reduced the difference in benchmarks between this adapter and the cowboy adapter, meaning it reduced performance for unary requests. But the intention is to make the code more secure.

grpc_server/lib/grpc/server/adapters/thousand_island.ex
sleipnir: What I'm saying is that this API has been like this since the library was released, and there has never been such a request from users on the server side.

But as I said, I don't object to any of this; I'll just deal with it in another PR.

grpc_server/lib/grpc/server/adapters/thousand_island.ex
sleipnir: I think we've already discussed this privately; another change was made to these items and sent a little while ago.

polvalente:
> I have an umbrella app with different apps starting their own `GRPC.Server.Supervisor`s in their own supervision trees (with different Endpoint modules), and it does not start with this branch with the following error:
> 
> ```elixir
> ** (Mix) Could not start application discovery: Discovery.Application.start(:normal, []) returned an error: shutdown: failed to start child: GRPC.Server.Supervisor
>     ** (EXIT) shutdown: failed to start child: GRPC.Server.StreamTaskSupervisor
>         ** (EXIT) already started: #PID<0.871.0>
> ```
> 
> The `Application` module with the supervision tree:
> 
> ```
> defmodule Discovery.Application do
>   @moduledoc false
> 
>   use Application
> 
>   @impl true
>   def start(_type, _args) do
>     children = [
>       Discovery.Repo,
>       Discovery.RateLimit,
>       {
>         GRPC.Server.Supervisor,
>         endpoint: Discovery.Endpoint,
>         port: Application.get_env(:discovery, :grpc_port),
>         start_server: true,
>         adapter_opts: [
>           cred: GRPC.Credential.new(ssl: Application.get_env(:discovery, :ssl))
>         ]
>       }
>     ]
> 
>     opts = [strategy: :one_for_one, name: Discovery.Supervisor]
>     Supervisor.start_link(children, opts)
>   end
> end
> ```
> 
> There is a nearly identical one in another app in the same umbrella but with a different endpoint, port, etc. So there are multiple GRPC servers and multiple Server.Supervisors running in the same BEAM instance.
> 
> This works with current releases of the grpc library, as well as with the `master` branch, but fails to start with this branch. Is this a known issue, or an expected change?

This is something we already mapped out during code review yesterday. This branch is very much a work in progress.

aseigo:
> is gonna be important when we move from the current state of flux we're in.

Feel free to ping me when you are at that point and I'll kick some of the tires and look more closely at the implementation as well at that point. Cheers!

sleipnir:
> great feat! any news on this?

Hi Alex, we'll continue this work soon, unfortunately I'm dealing with some personal and work management issues. But it's on my to-do list and I'll get back to it soon.

aseigo:
Checking in here to see if there's anything which others can do to move this forward in any meaningful way? I see some review comments, such as moving the http2 modules from `grpc_core` into `server`, but it is hard to know what remains to be done and therefore what could usefully be contributed ... 

With some insight into what remains, I could probably find some time to work in a branch based on this PR's branch to move some parts forward, if that would be useful and used. Let me know ... would really like to see this get to a state where it can be merged so we can start working with it :)

sleipnir:
Hello @aseigo, since this PR is very large and we had a very high demand for PRs right when we started, especially in grpc-client, we naturally had to pause this work to handle the volume of PR submissions. Unfortunately, this review and analysis of the other PRs has not yet been completed due to the limited time we have to work with OSS.
As far as I remember, there wasn't much missing here, but a good part of the rationale behind it was only in our heads and would need to be reviewed again for us to be able to move forward.
If you want to continue in this area, we can answer your questions and see together what the best way to proceed is.
Note: One of the changes we had to make during the review slightly degraded performance in exchange for "security," and I personally didn't like that as much because it strongly impacted performance. This is something to keep in mind in this work; anything we do can have a negative impact on overall performance, and the initial idea was to have a much more performant adapter than the current one.

Issue #479, opened 2025-12-02
Starting two `GRPC.Client.Supervisor`s prevents the VM from shutting down gracefully


ollien:
**Describe the bug**
If there are two `GRPC.Client.Supervisor`s running, with default OTP configurations, the VM will never shut down.

I believe this is caused by https://github.com/elixir-grpc/grpc/pull/466 - it allows for two unrelated supervisors to supervise the same pid, which leads to this unfortunate situation.

**To Reproduce**
Create a mix project with this Application. If you run `mix run --no-halt`, and then send a SIGTERM to the VM process (`kill -s TERM $(pgrep beam)`), the process will log `[notice] SIGTERM received - shutting down`, and then hang forever. If you remove one of the children of the supervisor (e.g. `sup2` below), the VM shuts down rather quickly after logging the SIGTERM message.

This application is obviously a contrived example, but it can be maddeningly hard to debug if, for instance, two dependency's OTP applications start `GRPC.Client.Supervisor`s for you.

This can be worked around if you set the [`shutdown_timeout` parameter](https://www.erlang.org/doc/apps/kernel/kernel_app.html) to some reasonable timeout _before_ sending the SIGTERM (e.g. `Application.put_env(:kernel, :shutdown_timeout, 1000)`), this causes the VM to brutally kill any process that doesn't shut down in time, which is likely not a good long-term fix 

```ex
defmodule GrpcDualSupTest.Application do
  @moduledoc false

  use Application

  defmodule TestSupervisor do
    use Supervisor

    def start_link(arg) do
      Supervisor.start_link(__MODULE__, arg)
    end

    @impl true
    def init(_) do
      children = [
        {GRPC.Client.Supervisor, []}
      ]

      Supervisor.init(children, strategy: :one_for_one)
    end
  end

  @impl true
  def start(_type, _args) do
    children = [
      Supervisor.child_spec({TestSupervisor, []}, id: :sup1),
      Supervisor.child_spec({TestSupervisor, []}, id: :sup2)
    ]

    opts = [strategy: :one_for_one, name: GrpcDualSupTest.Supervisor]
    Supervisor.start_link(children, opts)
  end
end
```

**Expected behavior**
The client supervisor should not prevent the VM from being shutdown.

**Versions:**
- OS: Ubuntu 22.04
- Erlang: OTP 25
- Elixir: Elixir 1.15.5
- mix.lock(grpc, gun, cowboy, cowlib):

<details>
[...]

PR #483, opened 2025-12-08
[Refactor]: GRPC client setup

sleipnir:
This is a break from recent versions but reverts to the previous behavior for setting up client connections.

This closes https://github.com/elixir-grpc/grpc/issues/479


grpc_client/lib/grpc/client/connection.ex
```diff
@@ -153,24 +153,17 @@ defmodule GRPC.Client.Connection do
...
-          ]
-
-          opts = [strategy: :one_for_one, name: MyApp.Supervisor]
-          Supervisor.start_link(children, opts)
+    case Process.whereis(GRPC.Client.Supervisor) do
+      nil ->
+        # For compatibility purposes only. In normal operation,
+        # the supervisor should be started by the application.
+        {:ok, pid} =
+          DynamicSupervisor.start_link(__MODULE__, [], name: GRPC.Client.Supervisor)
```
polvalente: This will likely crash the application. A bit hard to reproduce, but if GRPC.Client.Supervisor crashes in the main tree, and this is called before it restarts, I think the main app tree will fail with the :already_started return value.

polvalente: Furthermore, there's no retrocompatibility given that the client supervisor being added to the app tree is being done atomically to this change.

polvalente: Also, I just realized, grpc_client _technically_ doesn't have any published versions hahahah

polvalente: Let's just remove this Process.whereis search and refer to it by name where needed.

sleipnir: I should have been high (but I don't even use drugs).

.gitignore
```diff
@@ -25,5 +25,8 @@ erl_crash.dump
 .elixir_ls
 
 .elixir_tools
+grpc_client/.elixir-tools/
+grpc_core/.elixir-tools/
+grpc_server/.elixir-tools/
```
polvalente: This shouldn't be included in the library, but in the dev's global gitignore IMO
```suggestion
```

benchmark/lib/benchmark/application.ex
```diff
@@ -4,9 +4,7 @@ defmodule Benchmark.Application do
 
   @impl true
   def start(_type, _args) do
-    children = [
-      {GRPC.Client.Supervisor, []}
-    ]
+    children = []
```
polvalente: I think in this case the whole app module should be removed

grpc_client/lib/grpc/client/application.ex
```diff
@@ -0,0 +1,13 @@
+defmodule GRPC.Client.Application do
+  @moduledoc false
+  use Application
+
+  def start(_type, _args) do
+    children = [
+      {GRPC.Client.Supervisor, [name: GRPC.Client.Supervisor]}
+    ]
```
polvalente, line 8:
```suggestion
    children = [
      GRPC.Client.Supervisor
    ]
```

polvalente:
```
def start_link(opts) do
    case DynamicSupervisor.start_link(__MODULE__, opts, name: __MODULE__) do
      {:ok, _pid} = started ->
        started

      {:error, {:already_started, pid}} ->
        {:ok, pid}

      other ->
        other
    end
  end
```

I just realized this setup in the client supervisor needs to change too.
In fact, given that this is the only thing we have in that module, we can change to the code below and remove that file.

```suggestion
    children = [
      {DynamicSupervisor, [name: GRPC.Client.Supervisor]}
    ]
```

grpc_client/lib/grpc/client/application.ex
```diff
@@ -0,0 +1,13 @@
...
+
+    opts = [strategy: :one_for_one, name: Grpc.Supervisor]
```
polvalente:
```suggestion
    opts = [strategy: :one_for_one, name: GRPC.Supervisor]
```

PR #486, opened 2025-12-16
Send GRPC errors in trailers in all cases

aseigo:
In the documentation, there are several examples such as this one from the error handling guide:

```elixir
GRPC.Stream.from([1, 2])
|> GRPC.Stream.map(fn
  2 -> raise "boom"
  x -> x
end)
|> GRPC.Stream.map_error(fn
  {:error, {:exception, _reason}} ->
    {:error, GRPC.RPCError.exception(message: "Booomm")}
end)
```

However, we have found that this does not actually work in practice and instead errors are thrown on the server side with no errors arriving on the client side. Which is unfortunate, because this would be a very nice API pattern :)

Thankfully, it's a pretty simple fix: `GRPC.Server.Adapter.send_reply/2` needs to send errors to the client. This does mean adding one more function to the `Adapter` behaviour to facilitate access to the error sending capabilities of the adapter, however.

While pursuing this problem, I stumbled on a semi-related "foot-gun" in that the error exception did not require a status, and that would result in yet other errors being thrown. To address that, the status field is set to an unknown error by default, so that developers who only set a message don't end up with things blowing up on them.

With this PR, we now see errors appear in trailers. I think there are still discussions to be had in terms of multi-message streams and how errors could/should handled there. As it currently stands, the stream finishes when an error occurs. Of course, the developer can use an in-band error signalling message instead, leaving "actual" GRPC errors a communication-terminating event.

### DIFF
--- a/grpc_server/lib/grpc/server/stream.ex
+++ b/grpc_server/lib/grpc/server/stream.ex
@@ -70,6 +70,11 @@ defmodule GRPC.Server.Stream do
     do_send_reply(stream, [], opts)
   end
 
+  def send_reply(%{adapter: adapter} = stream, {:error, %GRPC.RPCError{} = error}, _opts) do
+    adapter.send_error(stream.payload, error)
+    stream
+  end
+
   def send_reply(
         %{grpc_type: :server_stream, codec: codec, access_mode: :http_transcoding, rpc: rpc} =
           stream,
--- a/grpc_client/test/grpc/integration/server_test.exs
+++ b/grpc_client/test/grpc/integration/server_test.exs
@@ -268,7 +268,7 @@ defmodule GRPC.Integration.ServerTest do
         end)
       end)
 
-    assert logs =~ "Exception raised while handling /helloworld.Greeter/SayHello"
+    assert logs =~ "(GRPC.RPCError) Please authenticate"
   end
 
   test "return errors for unknown errors" do
@@ -284,7 +284,7 @@ defmodule GRPC.Integration.ServerTest do
         end)
       end)
 
-    assert logs =~ "Exception raised while handling /helloworld.Greeter/SayHello"
+    assert logs =~ "unknown error(This is a test, please ignore it)"
   end
 
   test "logs error if exception_log_filter returns true" do
@@ -329,7 +329,6 @@ defmodule GRPC.Integration.ServerTest do
 
       {pid, ref} = :erlang.binary_to_term(data)
       send(pid, {:exception_log_filter, ref, exception})
-
       true
     end
   end
@@ -396,6 +395,7 @@ defmodule GRPC.Integration.ServerTest do
       {:ok, channel} = GRPC.Stub.connect("localhost:#{port}")
       rect = %Routeguide.Rectangle{}
       error = %GRPC.RPCError{message: "Please authenticate", status: 16}
+
       assert {:error, ^error} = channel |> Routeguide.RouteGuide.Stub.list_features(rect)
     end)
   end
--- a/grpc_server/lib/grpc/server/adapter.ex
+++ b/grpc_server/lib/grpc/server/adapter.ex
@@ -24,4 +24,6 @@ defmodule GRPC.Server.Adapter do
   @callback send_reply(state, content :: binary(), opts :: keyword()) :: any()
 
   @callback send_headers(state, headers :: map()) :: any()
+
+  @callback send_error(state, headers :: map()) :: any()
 end
--- a/grpc_server/lib/grpc/server/adapters/cowboy.ex
+++ b/grpc_server/lib/grpc/server/adapters/cowboy.ex
@@ -168,6 +168,11 @@ defmodule GRPC.Server.Adapters.Cowboy do
     Handler.set_resp_trailers(pid, trailers)
   end
 
+  @impl true
+  def send_error(%{pid: pid}, error) do
+    Handler.send_error(pid, error)
+  end
+
   def send_trailers(%{pid: pid}, trailers) do
     Handler.stream_trailers(pid, trailers)
   end
--- a/grpc_server/lib/grpc/server/adapters/cowboy/handler.ex
+++ b/grpc_server/lib/grpc/server/adapters/cowboy/handler.ex
@@ -222,6 +222,13 @@ defmodule GRPC.Server.Adapters.Cowboy.Handler do
   end
 
   @doc """
+  Asynchronously send an error to the client.
+  """
+  def send_error(pid, error) do
+    send(pid, {:send_error, error})
+  end
+
+  @doc """
   Asynchronously send back to client a chunk of `data`, when `http_transcode?` is true, the
   data is sent back as it's, with no transformation of protobuf binaries to http2 data frames.
   """
@@ -485,6 +492,16 @@ defmodule GRPC.Server.Adapters.Cowboy.Handler do
     {:ok, req, state}
   end
 
+  def info({:send_error, error}, req, state) do
+    req = send_error(req, error, state, :rpc_error)
+
+    [req: req]
+    |> ReportException.new(error)
+    |> maybe_log_error(state.exception_log_filter)
+
+    {:stop, req, state}
+  end
+
   def info({:handling_timeout, _}, req, state) do
     error = %RPCError{status: GRPC.Status.deadline_exceeded(), message: "Deadline expired"}
     req = send_error(req, error, state, :timeout)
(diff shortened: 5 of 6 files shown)


grpc_server/lib/grpc/server/stream.ex
```diff
@@ -70,6 +70,11 @@ defmodule GRPC.Server.Stream do
     do_send_reply(stream, [], opts)
   end
 
+  def send_reply(%{adapter: adapter} = stream, {:error, %GRPC.RPCError{} = error}, _opts) do
+    adapter.send_error(stream.payload, error)
```
polvalente, line 74: Do we really need a new send_error callback? What does send_reply not have?

grpc_server/lib/grpc/server/stream.ex
```diff
@@ -70,6 +70,11 @@ defmodule GRPC.Server.Stream do
     do_send_reply(stream, [], opts)
   end
 
+  def send_reply(%{adapter: adapter} = stream, {:error, %GRPC.RPCError{} = error}, _opts) do
```
polvalente, line 73: One thing we need to check is if the only errors that can reach here are in the form of this struct, of if we need to add another clause

grpc_server/lib/grpc/server/stream.ex
aseigo: The function already exists, it's just not accessible via the Adaptor behaviour.

The difference to send_reply is that the error path doesn't try to GRPC-encode the message, puts the error information into the trailers, and ends the connection immediately.

It would be possible to change the contract of `Adapter.send_reply/3` so that implementations must pattern match on errors being passed in as the data, but that means overloading the purpose of `send_reply`: sometimes it sends a reply, sometimes it sends an error.

I felt it cleaner to keep these two paths clearly separated, and as the adapter will need to track its own errors anyways and have a way to deal with those, this is isn't actually introducing new complexity.

But if you'd prefer to overload `send_reply` and make it a requirement for adapters to handle RPCError structs there, that can also be done!

polvalente: Second paragraph sold me on the new callback!

polvalente: Second paragraph sold me on the new callback!

grpc_server/lib/grpc/server/stream.ex
aseigo: That's a very good question.

I've oscillated between having one and not having one, as it would be "easy" enough to try and match on any error .. or even just errors with strings (an error message) and wrap those in an RPCError. 

The flip side of that is should that occur, it is almost certainly a developer error. Either an RPCError , a thrown exception, or a protobuf-encodable struct should be the result. Returning Something Else(tm) is probably something like a raw Ecto query or whatever that has been returned, and simply passing that back silently to the client, even as an error, feels somewhere between dangerous and too much magic.

In those cases, I'd rather see a request crash on the server side that can be tracked and addressed.

sleipnir:
> In the documentation, there are several examples such as this one from the error handling guide:
> 
> ```elixir
> GRPC.Stream.from([1, 2])
> |> GRPC.Stream.map(fn
>   2 -> raise "boom"
>   x -> x
> end)
> |> GRPC.Stream.map_error(fn
>   {:error, {:exception, _reason}} ->
>     {:error, GRPC.RPCError.exception(message: "Booomm")}
> end)
> ```
> 
> However, we have found that this does not actually work in practice and instead errors are thrown on the server side with no errors arriving on the client side. Which is unfortunate, because this would be a very nice API pattern :)
> 

@aseigo Thank you for bringing this up! I'd like to clarify that the current behavior is actually intentional by design, not a bug. Let me explain the flow and the reasoning behind it.

## Current Error Flow (Working as Designed)

When using `GRPC.Stream.map_error/2` in streaming handlers, here's what happens:

* Handler calls run_with (inside GRPC.Server.send_reply for streaming)
* Flow executes and map_error transforms errors into {:error, GRPC.RPCError{}}
* run_with detects the error and calls send_response
* send_response raises the GRPC.RPCError (intentional control flow mechanism)
* Adapter catches the exception:
  * Cowboy: catches via catch and does exit({:shutdown, {error, []}})
  * Stream process/handler sends error trailers to client
  * Stream terminates with proper gRPC error status
  * This flow successfully sends errors to clients through HTTP/2 trailers with the correct grpc-status and grpc-message.

Why We Terminate the Stream on Error
The decision to terminate the stream when an error occurs is intentional for several important reasons:

1. Trailer Accumulation Problem
  In HTTP/2 and gRPC protocol, trailers are sent once at the end of a stream with the END_STREAM flag. If we were to continue streaming after an error:
  * Problematic scenario:[msg1, ERROR, msg2, ERROR2, msg3] # Which trailers do we send? # grpc-status from ERROR or ERROR2?# We can't send multiple trailers - protocol limitation!
HTTP/2 only allows one HEADERS frame with END_STREAM at the end. We cannot send:
Multiple trailer frames (protocol violation)
Accumulate errors indefinitely (memory/state management nightmare)

2. gRPC Protocol Semantics
  Per gRPC protocol specification:
  "Status is sent as a single trailer at the end of the stream"
  The protocol is designed around the concept of one status per RPC. Errors are meant to be terminating events.

3. Clear Error Semantics
  When an error occurs in a stream transformation:

```elixir
GRPC.Stream.from([1, 2, 3])
|> GRPC.Stream.map(fn x ->   if x == 2, do: raise("error"), else: xend)
|> GRPC.Stream.map_error(fn _ ->   GRPC.RPCError.exception(status: :internal, message: "Processing failed")end)
```

The client receives:

✅ msg1 (successful)
✅ Trailers: **grpc-status=13, grpc-message="Processing failed"**
✅ Stream closes

The alternative (continuing after error) would be confusing:

❌ What does receiving more messages after an error mean?
❌ Was the error "recovered"?
❌ How does the client know the final status?

Alternative: In-Band Error Signaling
If you need to handle recoverable errors within a stream, the recommended pattern is in-band error signaling:

Define a response type that includes success/error variants, in other words, map the errors to common data types within your application's business logic.

This way:

✅ Errors are part of the message protocol
✅ Stream continues after recoverable errors
✅ Final grpc-status=0 indicates successful completion
✅ Client can handle per-message errors

The current behavior is:

raise → adapter catches → sends trailers works perfectly for both Cowboy and ThousandIsland
Stream termination on error prevents trailer accumulation issues and follows gRPC protocol semantics
map_error does work - it transforms errors before sending them to clients as terminating trailers
For non-terminating errors, use in-band error messages in your protobuf definitions
The key insight is: gRPC errors are meant to be stream-terminating events. If you need to continue streaming despite errors, those "errors" should be modeled as regular response messages, not gRPC status codes.

Does this clarify the design? I'm happy to discuss further if you have a specific use case that this pattern doesn't address well!
But do you have any evidence to show us that the client didn't receive the error via the trailer? Couldn't it be a bug in the clients? What you're implementing seems to be exactly the original design, just in a different way. I'd like to discuss this a bit more before proceeding with the review.

sleipnir:
> > All I had to do was adjust the patterns of the `send_response/3` function
> 
> With that adjustment, on the server-side I see the exception in the logs rather than random errors (yes! nice!) but I am still not seeing them on the client side. I applied the patch to the `master` branch as well as my grpcweb-with-trailers branch, and tested with a native client and a grpcweb client and I'm still seeing `grpc-message: '', grpc-status: 0`
> 
> So maybe handling this in `Stream.send_response` is a good way forward, but there's still getting the response to the client it seems ...

Did you run the tests? They're important to know if it's something the clients are handling or a server error. In my case, I believe it's a client issue because the integration tests worked with the standard elixir client, correctly retrieving the errors.

@aseigo How can I replicate your problem scenario here on my end?

sleipnir:
> > Did you run the tests?
> 
> The tests in the `feat/new-server-adapter` work. I suspect (though haven't confirmed due to time restrictions here) that it is due to the ThousandIslands-based adapter doing this correctly while the existing cowboy adapter in the `master` branch is not?
> 

Those tests ran with Cowboy; `run_server` without passing options will use the default adapter, which in this case is Cowboy. To run the tests with ThousandIsland, you would need to pass the option:

```elixir
run_server([MyService], fn port ->

# ...
end, 0, adapter: GRPC.Server.Adapters.ThousandIsland)
```

I only used that branch because it was convenient to be able to commit to it without impacting anything in production.
I also will continue investigating the issue before reviewing your PR. I don't like approving things without understanding the real reason why they are trying to solve. 😸

sleipnir:
> Fixed in #487, closing! Nice work @sleipnir !

I'm the one who should be thanking everyone for their hard work.

PR #487, opened 2025-12-16
[Fix]: Stream map error 

sleipnir:
This PR corrects a pattern matching issue in the send_response function and adds integration tests for the map_error function. There was a suspicion that this function was not correctly sending errors in the trailers, so I decided to add new tests to specifically validate this.

@aseigo I've brought the changes to the master branch in this PR, and I see that everything works. I haven't been able to reproduce your `map_error` error aside from the pattern matching issue. The tests are going according to what the Elixir customer expects.

Is there another way I can reproduce the problem?

Captura de tela de 2025-12-16 12-39-49

Captura de tela de 2025-12-16 12-41-09

### DIFF
--- a/grpc_client/test/grpc/integration/stream_test.exs
+++ b/grpc_client/test/grpc/integration/stream_test.exs
@@ -33,4 +33,326 @@ defmodule GRPC.StreamTest do
       end)
     end
   end
+
+  describe "map_error/2" do
+    defmodule MapErrorService do
+      use GRPC.Server, service: Routeguide.RouteGuide.Service
+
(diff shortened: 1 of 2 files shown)


sleipnir:
> With a very simple (~10 lines) change to the `grpcweb` branch rebased onto this branch, we now see errors on both native and web clients with returning `{:error, ...}` tuples from `map` or `map_error`. Very nice!
> 
> I suggest merging this branch, and I can then close #486 and ammend #481

Done!

sleipnir:
@aseigo Thank you for your patience and continued contribution.

Issue #464, opened 2025-11-10
What happens to the `GRPC.Client.Connection` when the server disconnects?


mrmicahcooper:
**Describe the question**
Hi,

I'm looking at using the new `GRPC.Client.Connection` for managed connections instead of managing them myself and I can't tell from reading through the code or the docs:

What happens when the server disconnects as it does often with Google's GRPC apis? Is this handled by the `Grpc.Client.Supervisor` for me? 

Thanks! ❤️ 

**Versions:**
- mix.lock(grpc, gun, cowboy, cowlib): grpc 0.11.4.

sleipnir:
Hi! @mrmicahcooper  Thanks for the question — and for using elixir-grpc!

The `GRPC.Client.Connection` module is not intended to be used directly.
It is internally supervised by its own supervisor, and `GRPC.Stub` only starts and stops connections through `Stub.connect/2` — so you typically don’t need to interact with `GRPC.Client.Connection` yourself.

If you have a specific scenario in mind (like Google’s gRPC APIs closing connections frequently), feel free to share more context — I’d be happy to clarify the expected behavior or discuss the best approach for your use case.

mrmicahcooper:
No this is excellent. Thank you for the explanation.
I will attempt to create working gist to exemplify the behavior I've seen. If I'm able to, I'll put it in a separate issue.

I'll close this with 1 more opinion/piece of feedback.
If the channel is really just a key to reference the underlying managed GRPC connections, that could be different, it'd be awesome if I could give that a name to reference instead. Something like:

```elixir
GRPC.Stub.connect("pubsub.googleapis.com", [name: :my_pubsub_reference])

Pubsub.Stub.publish(:my_pubsub_reference, request, opts)
```

That way, I don't need the response of `GRPC.Stub.connect()` to send data through that managed connection.
Kinda like how `Finch` manages its HTTP1 connections.

Anyways. Thanks again!

sleipnir:
> No this is excellent. Thank you for the explanation. I will attempt to create working gist to exemplify the behavior I've seen. If I'm able to, I'll put it in a separate issue.
> 
> I'll close this with 1 more opinion/piece of feedback. If the channel is really just a key to reference the underlying managed GRPC connections, that could be different, it'd be awesome if I could give that a name to reference instead. Something like:
> 
> GRPC.Stub.connect("pubsub.googleapis.com", [name: :my_pubsub_reference])
> 
[...]

PR #491, opened 2026-01-23
Support naming a virtual channel

mrmicahcooper:
## Allow a GRPC.Channel to be created with a name.
This removes the need for developers to store the `Channel` that's returned from `GRPC.Client.Connection.connect`.

This is one attempt to solve the feedback I previously mentioned in a separate issue:
https://github.com/elixir-grpc/grpc/issues/464#issuecomment-3514204429

I'll explain how I do things now and what this PR is attempting to do:

### Current
I have an elixir app that talks to a GRPC endpoint. 
So I say `{:ok, channel} = GRPC.Stub.connect("https://grpc.endpoint.com")`
Now, I need to hold on to this channel somewhere so I can use it later so I create a GenServer with this channel in its state.
Now I call the GenServer which uses the channel in its state to forward the request to the GRPC stub.
Something like `HelloWorld.Stub.say_hello(channel, "hello")`
This works and has the ability to ensure the GRPC.Client.Supervisor is started before we start this GenServer starts.
The downside is now I have to have an entire process just to store a virtual channel.

### Proposed
Use a module attribute to store the name of the channel I want.
`@endpoint_channel :my_service_channel`
Now create a channel with that name and ignore the response:
`{:ok, _channel} = GRPC.Client.Connection.connect("https://grpc.endpoint.com", name: @enpdoint_channel)`
When I want to use that channel, I use the predetermined name:
Something like `HelloWorld.Stub.say_hello(%Channel{ref: @endpoint_channel}), "hello")`
Now I don't need to store that channel anywhere.

This PR implements the "proposed" and ignores the fact that you'd have to manually create this connection somewhere in your app after the `GRPC.Client.Supervisor` is started.

Something that might be cool is the ability to create these named channels in `config`. 
But this PR seems like a really low risk way add this feature.

I'd love any feedback or thoughts about a different approach.

Thanks!

### DIFF
--- a/grpc_client/lib/grpc/client/connection.ex
+++ b/grpc_client/lib/grpc/client/connection.ex
@@ -153,26 +153,6 @@ defmodule GRPC.Client.Connection do
   """
   @spec connect(String.t(), keyword()) :: {:ok, Channel.t()} | {:error, any()}
   def connect(target, opts \\ []) do
-    supervisor_pid = Process.whereis(GRPC.Client.Supervisor)
-
-    if is_nil(supervisor_pid) or !Process.alive?(supervisor_pid) do
-      raise """
-      GRPC.Client.Supervisor is not running. Please ensure it is started as part of your
-      application's supervision tree:
-
-          children = [
-            {GRPC.Client.Supervisor, []}
-          ]
-
-          opts = [strategy: :one_for_one, name: MyApp.Supervisor]
-          Supervisor.start_link(children, opts)
-
-      You can also start it manually in scripts or test environments:
-
-          {:ok, _pid} = DynamicSupervisor.start_link(strategy: :one_for_one, name: GRPC.Client.Supervisor)
-      """
-    end
-
     case build_initial_state(target, opts) do
       {:ok, initial_state} ->
         ch = initial_state.virtual_channel


test/grpc/integration/stub_test.exs
```diff
@@ -68,6 +68,21 @@ defmodule GRPC.Integration.StubTest do
...
+    run_server(HelloServer, fn port ->
+      {:ok, _channel} =
+        GRPC.Client.Connection.connect("localhost:#{port}",
+          interceptors: [GRPC.Client.Interceptors.Logger],
+          name: :foobar
+        )
+
+      name = "GRPC user!"
+      req = %Helloworld.HelloRequest{name: name}
+      {:ok, reply} = %GRPC.Channel{ref: :foobar} |> Helloworld.Greeter.Stub.say_hello(req)
```
sleipnir: I think the test could reflect the use case you mentioned in the description of this PR

mrmicahcooper: Updated to use a module attribute.

lib/grpc/client/connection.ex
```diff
@@ -173,7 +173,7 @@ defmodule GRPC.Client.Connection do
       """
     end
 
-    ref = make_ref()
+    {ref, opts} = Keyword.pop(opts, :name, make_ref())
```
polvalente: Is it possible for us to use Keyword.validate in this function to help document the options? If so, we can just have `name: make_ref()` in the option arguments there

polvalente: @sleipnir

test/grpc/integration/stub_test.exs
```diff
@@ -68,6 +68,22 @@ defmodule GRPC.Integration.StubTest do
...
   end
 
+  @channel_name :my_channel
+  test "use a channel name to send a message" do
+    run_server(HelloServer, fn port ->
+      {:ok, _channel} =
+        GRPC.Client.Connection.connect("localhost:#{port}",
+          interceptors: [GRPC.Client.Interceptors.Logger],
+          name: @channel_name
+        )
```
polvalente:
```suggestion
  test "use a channel name to send a message" do
    run_server(HelloServer, fn port ->
      {:ok, _channel} =
        GRPC.Client.Connection.connect("localhost:#{port}",
          interceptors: [GRPC.Client.Interceptors.Logger],
          name: :my_channel
        )
```

polvalente: No need for the module attribute. Maybe if this was a macro behavior could be different, but given that it's a function, we can just rely on the atom value.

lib/grpc/client/connection.ex
mrmicahcooper: Done!

sleipnir: broken here https://github.com/mrmicahcooper/grpc/blob/e0c7bd19978d430ce182481b8fd338dd153bfae0/grpc_client/test/grpc/integration/stub_test.exs#L81

grpc_client/lib/grpc/client/connection.ex
```diff
@@ -153,9 +153,27 @@ defmodule GRPC.Client.Connection do
   """
   @spec connect(String.t(), keyword()) :: {:ok, Channel.t()} | {:error, any()}
   def connect(target, opts \\ []) do
-    ref = make_ref()
+    supervisor_pid = Process.whereis(GRPC.Client.Supervisor)
```
polvalente: @sleipnir remind me again why we don't have this in an application.ex for grpc_client?

mrmicahcooper: Hmm I must have pulled this in during a rebase. It doesn't belong. 
Fixed!

PR #495, opened 2026-02-09
Add IPv6 support for gRPC gun adapter

KalleJoP:
## Summary

Improves IPv6 support for gRPC client connections by (1) falling back to AAAA DNS lookups when no A records exist, and (2) passing IP addresses to Gun as tuples instead of strings.

## Why

IPv6 connections were failing with the Gun adapter for two reasons:

1. **Missing AAAA lookup** — For `dns://` targets, only A (IPv4) records were queried. Hosts that expose only IPv6 or that we want to use over IPv6 had no resolved addresses, so the client never attempted an IPv6 connection.
2. **Host format for Gun** — When an IPv6 address was used (e.g. from a manual config), it was passed to `:gun.open/3` as a string. Gun expects IP addresses as `inet:ip_address()` tuples (e.g. an 8-element tuple for IPv6). Passing a string could lead to failed or unreliable IPv6 connections.

This PR fixes both: DNS now falls back to AAAA when A returns nothing, and we parse IP strings to tuples before calling Gun so IPv6 addresses are in the correct form.

### 1. DNS resolver: AAAA fallback (`grpc_client/lib/grpc/client/resolver/dns.ex`)

- If the A-record lookup returns no addresses, the resolver now falls back to an AAAA lookup so IPv6-only or dual-stack targets can be resolved.
- Logic: try A first; if the result is an empty list, call `lookup_addresses(host, :aaaa)`.

### 2. Gun adapter: parse host to tuple before `open` (`grpc_client/lib/grpc/client/adapters/gun.ex`)

- Before calling `:gun.open/3`, the host is passed through a new `parse_address/1` helper.
- If the host is a valid IP string (IPv4 or IPv6), `:inet.parse_address/1` is used and the resulting **tuple** is passed to `:gun.open`.
- If parsing fails (e.g. hostname), the host is passed as a charlist as before.
- This ensures IPv6 addresses (e.g. from AAAA) are given to Gun in the form it expects

### DIFF
--- a/grpc_client/lib/grpc/client/resolver/dns.ex
+++ b/grpc_client/lib/grpc/client/resolver/dns.ex
@@ -48,7 +48,7 @@ defmodule GRPC.Client.Resolver.DNS do
   defp lookup_addresses(host) do
     case lookup_addresses(host, :a) do
       {:ok, [_ | _] = addrs} -> {:ok, addrs}
-      {:ok, addrs} when is_list(addrs) and length(addrs) == 0 -> lookup_addresses(host, :aaaa)
+      {:ok, []} -> lookup_addresses(host, :aaaa)
       other -> other
     end
   end


grpc_client/lib/grpc/client/resolver/dns.ex
```diff
@@ -46,7 +46,16 @@ defmodule GRPC.Client.Resolver.DNS do
   end
 
   defp lookup_addresses(host) do
-    case adapter().lookup(host, :a) do
+    case lookup_addresses(host, :a) do
+      {:ok, addrs} when is_list(addrs) and length(addrs) > 0 -> {:ok, addrs}
+      # if no A records, try IPv6
```
sleipnir: This comment is unnecessary and can be removed.

sleipnir (changes requested):
Could you add a test to verify this fallback?

grpc_client/lib/grpc/client/resolver/dns.ex
```diff
@@ -46,7 +46,15 @@ defmodule GRPC.Client.Resolver.DNS do
   end
 
   defp lookup_addresses(host) do
-    case adapter().lookup(host, :a) do
+    case lookup_addresses(host, :a) do
+      {:ok, addrs} when is_list(addrs) and length(addrs) > 0 -> {:ok, addrs}
```
polvalente:
```suggestion
      {:ok, [_ | _] = addrs} -> {:ok, addrs}
```

length is O(n)

grpc_client/lib/grpc/client/resolver/dns.ex
```diff
@@ -46,7 +46,15 @@ defmodule GRPC.Client.Resolver.DNS do
...
+      {:ok, addrs} when is_list(addrs) and length(addrs) == 0 -> lookup_addresses(host, :aaaa)
```
polvalente:
```suggestion
      {:ok, []} -> lookup_addresses(host, :aaaa)
```

polvalente (approved):

PR #498, opened 2026-02-21
fix(mint): fully drain buffer when an HTTP/2 frame carries multiple gRPC messages

arctarus:
## Summary

`StreamResponseProcess` would only ever decode **one** gRPC message per incoming
HTTP/2 data frame, leaving any additional complete messages silently stranded in
the internal buffer. Consumers would then block waiting for the next network event
to arrive, even though the data they needed was already sitting inside the process.
This commit replaces the single-pass decode with a recursive `drain_buffer/5`
helper that keeps decoding until the buffer holds no more complete messages.

---

## Background

The gRPC-over-HTTP/2 wire format wraps each protobuf payload in a 5-byte
length-prefixed envelope (1 compressed-flag byte + 4 big-endian length bytes).
A single HTTP/2 DATA frame has no obligation to carry exactly one such envelope —
it can carry many, or a fraction of one. The Mint adapter receives raw DATA frames
via `handle_info/2` in `ConnectionProcess`, which forwards each chunk to
`StreamResponseProcess` by calling `StreamResponseProcess.consume/3`.

`StreamResponseProcess` maintains an internal binary `buffer` to handle the case
where a chunk ends in the middle of a message. When new data arrives it
concatenates the buffer with the incoming chunk and attempts to extract a gRPC
message with `GRPC.Message.get_message/2`.

---

## The Bug

The previous implementation called `GRPC.Message.get_message/2` **once** and then
stopped, regardless of how many complete messages were present in the combined
buffer:

```elixir
# BEFORE — single-pass, only ever decodes the first message
case GRPC.Message.get_message(buffer <> data, state.compressor) do
  {{_, message}, rest} ->
    response = codec.decode(message, res_mod)
    new_responses = :queue.in({:ok, response}, responses)
    new_state = %{state | buffer: rest, responses: new_responses}
    {:reply, :ok, new_state, {:continue, :produce_response}}

  _ ->
    new_state = %{state | buffer: buffer <> data}
    {:reply, :ok, new_state, {:continue, :produce_response}}
end
```

Consider a server that sends two small responses packed into one DATA frame — a
common optimisation for short replies like health-checks or acknowledgements. The
flow before this fix was:

```
DATA frame arrives (msg_1 <> msg_2)
  → get_message decodes msg_1, stores msg_2 in buffer
  → consumer receives msg_1 ✓
  → consumer calls :get_response again
  → responses queue is empty, done = false
  → consumer BLOCKS ← bug: msg_2 is already in the buffer
  → (no new data is coming; stream stalls indefinitely, or until a timeout)
```

The same problem applies to any number of messages per frame: N messages in one
chunk required N separate network events to be fully consumed. In streaming RPCs
with high message rates and small payloads — exactly the workloads where
performance matters most — this could cause significant latency spikes or
permanent consumer hangs if the last DATA frame contained more than one message
and no further frames were expected.

Additionally, the `_ ->` branch recomputed `buffer <> data` a second time when
storing the incomplete remainder, allocating an unnecessary duplicate binary.

---

### `drain_buffer/5` — recursive buffer draining

A new private recursive helper exhausts the buffer completely in a single
`consume` call:

```elixir
# After — drain_buffer/5 recurses until no complete message remains
defp drain_buffer(buffer, compressor, codec, res_mod, responses_queue) do
  case GRPC.Message.get_message(buffer, compressor) do
    {{_, message}, rest} ->
      response = codec.decode(message, res_mod)
      drain_buffer(rest, compressor, codec, res_mod, :queue.in({:ok, response}, responses_queue))

    _ ->
      {buffer, responses_queue}
  end
end
```

The updated `handle_call` clause for `:data` events:

```elixir
# After — single concatenation, full drain
def handle_call({:consume_response, {:data, data}}, _from, state) do
  %{
    buffer: buffer,
    grpc_stream: %{response_mod: res_mod, codec: codec},
    responses: responses
  } = state

  combined = buffer <> data
  {new_buffer, new_responses} =
    drain_buffer(combined, state.compressor, codec, res_mod, responses)

  {:reply, :ok, %{state | buffer: new_buffer, responses: new_responses},
   {:continue, :produce_response}}
end
```

Key properties of this approach:

- **Correctness** — every complete message reachable in the buffer is decoded and
  enqueued before control is returned. Consumers can never block on data that is
  already present.
- **Single allocation** — `buffer <> data` is evaluated exactly once and bound to
  `combined`. The old code allocated the same binary twice in the no-match branch.
- **Tail-call friendly** — each recursive call receives the `rest` binary returned
  directly by `GRPC.Message.get_message/2`, which is a sub-binary slice with no
  additional allocation.
- **Termination guaranteed** — `GRPC.Message.get_message/2` only returns
  `{{flag, message}, rest}` when `rest` is strictly shorter than the input.
  The buffer strictly shrinks on every recursive step, so the recursion always
  terminates.

---

## Files Changed

| File | Change |
|------|--------|
| `grpc_client/lib/grpc/client/adapters/mint/stream_response_process.ex` | Added `drain_buffer/5`; rewrote the `:data` handler; expanded module comment with a *Buffer draining* section |
| `grpc_client/test/grpc/adapters/mint/stream_response_process_test.exs` | Added 3 unit tests and 1 integration test |
| `grpc_client/CHANGELOG.md` | Added `## Unreleased` entry |

---

## Tests Added

All new tests live in `stream_response_process_test.exs`.

### Unit tests — `"handle_call/3 - data"`

| Test | What it verifies |
|------|-----------------|
| `"decodes all complete messages when a single chunk contains two full messages"` | Two messages packed in one chunk are both decoded; buffer is empty after the call |
| `"decodes all complete messages and retains the incomplete tail in the buffer"` | Two complete messages plus a partial third are handled correctly; only the partial tail stays in the buffer |
| `"decodes all complete messages when a single chunk contains three full messages"` | Guards against off-by-one errors in the recursion; all three messages decoded, empty buffer |

### Integration test — `"build_stream/1"`

| Test | What it verifies |
|------|-----------------|
| `"emits all messages when a single chunk contains multiple complete gRPC messages"` | Starts a real `StreamResponseProcess`, calls `consume/3` once with two serialised messages concatenated, then `done/1`, and asserts `Enum.to_list/1` returns both decoded responses — exactly the end-to-end scenario that was broken before |

All pre-existing tests in the file continue to pass without modification (30 → 30,
no failures).

---

## How to Reproduce the Bug (before this fix)

```elixir
{:ok, pid} = StreamResponseProcess.start_link(stream, true)

hello = <<0, 0, 0, 0, 12, 10, 10, 72, 101, 108, 108, 111, 32, 76, 117, 105, 115>>
bye   = <<0, 0, 0, 0, 10, 10,  8, 66, 121, 101,  32,  76, 117, 105, 115>>

stream = StreamResponseProcess.build_stream(pid)

# Simulates a single DATA frame carrying both messages
StreamResponseProcess.consume(pid, :data, hello <> bye)
StreamResponseProcess.done(pid)

# Before fix: returns only [{:ok, %HelloReply{message: "Hello Luis"}}]
# After fix:  returns both decoded replies
Enum.to_list(stream)
```

---

## Risk Assessment

**Low.** The change is purely additive in the happy path — the existing
single-decode behaviour is a strict subset of the new drain loop. Any binary that
previously produced zero or one decoded message will produce the same result. The
only difference is that buffers containing two or more complete messages will now
be fully drained instead of partially consumed. No public API, no process
lifecycle, and no inter-process communication protocol has been modified.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,3 @@ erl_crash.dump
 /log
 
 grpc-*.tar
-
-# language server
-.expert/
--- a/grpc_client/CHANGELOG.md
+++ b/grpc_client/CHANGELOG.md
@@ -1,15 +1,5 @@
 # Changelog
 
-## Unreleased
-
-### Fixed
-
-  * `StreamResponseProcess`: buffer is now fully drained when a single HTTP/2
-    data frame contains multiple gRPC messages, preventing consumers from
-    blocking unnecessarily while decoded data is already available in the
-    process buffer. A new private `drain_buffer/5` recursive helper replaces
-    the previous single-pass `GRPC.Message.get_message/2` call.
-
 ## v1.0.0-rc.1 (2025-12-02)
 
 ### Enhancements


.gitignore
```diff
@@ -23,3 +23,6 @@ erl_crash.dump
 /log
 
 grpc-*.tar
+
+# language server
+.expert/
```
polvalente: Please remove this. This should live in your global gitignore

grpc_client/CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Changelog
 
+## Unreleased
+
+### Fixed
+
+  * `StreamResponseProcess`: buffer is now fully drained when a single HTTP/2
```
polvalente: We don't generally update the changelog until the release. 

Also, I think that the rc version hasn't even been released

grpc_client/lib/grpc/client/adapters/mint/stream_response_process.ex
```diff
@@ -96,18 +106,17 @@ defmodule GRPC.Client.Adapters.Mint.StreamResponseProcess do
...
+    # combined buffer in a single pass, avoiding repeated allocation and
+    # preventing consumers from blocking when multiple messages arrive in
+    # the same HTTP/2 data frame.
+    combined = buffer <> data
 
-      _ ->
-        new_state = %{state | buffer: buffer <> data}
-        {:reply, :ok, new_state, {:continue, :produce_response}}
-    end
+    {new_buffer, new_responses} =
```
polvalente, line 115: Let's redefine new_state instead of an in-place update so that the return format is easier to read

polvalente (approved):

PR #510, opened 2026-03-12
Feat/release 1.0.0 rc.1

sleipnir:


grpc/CHANGELOG.md
```diff
@@ -1,15 +1,34 @@
 # Changelog
 
-## v1.0.0-rc.1 (2025-12-02)
+## v1.0.0-rc.1 (2026-03-12)
```
polvalente, line 3: I think the changelog should follow the previous format, where we have the spirit of what was done rather than a list of commits. We can leave the list to the autogenerated release notes or tag compares

grpc_client/mix.exs
```diff
@@ -29,8 +29,8 @@ defmodule GrpcClient.MixProject do
 
   defp deps do
     [
-      {:grpc_core, path: "../grpc_core"},
-      # {:grpc_core, "~> 1.0"}, # Uncomment for hex release
+      #{:grpc_core, path: "../grpc_core"},
+      {:grpc_core, "~> 1.0"}, # Uncomment for hex release
```
polvalente: I think we might need to set 1.0.0-rc.1, I'm not sure if ~> accepts RCs.
We might need to update mix.lock before publishing client and server as well

sleipnir: Yeap

polvalente (approved):

PR #513, opened 2026-03-30
feat: periodic DNS re-resolution for client-side service discovery

cgreeno:
## Summary

Adds periodic DNS re-resolution to the gRPC client, enabling automatic discovery of new backends and removal of stale ones. This is essential for Kubernetes deployments where pods scale up/down behind headless services, matching the behaviour of grpc-go and grpc-java.

## What's included

**New module: `GRPC.Client.DnsResolver`** — a dedicated GenServer linked to each Connection that owns the entire DNS resolve lifecycle:

- Periodic re-resolution on a configurable interval (default 30s)
- Exponential backoff on DNS failure (doubles per failure, caps at 5min, resets on success)
- Rate limiting to prevent DNS flooding from `resolve_now/1` calls (default 5s floor)
- Telemetry events: `[:grpc, :client, :resolve, :stop]` on success, `[:grpc, :client, :resolve, :error]` on failure
- Sends `{:dns_result, result}` to the Connection process for channel reconciliation

**Changes to `GRPC.Client.Connection`:**

- Channel reconciliation on re-resolution: new backends get channels created, removed backends get disconnected, unchanged backends are left as-is
- Load balancer re-initialized with updated address list after reconciliation
- Fallback to any healthy channel when LB picks a failed one
- Retry previously failed channels on subsequent re-resolution cycles
- `resolve_now/1` public API for on-demand re-resolution (e.g. from health checks)
- Fixed `:refresh` handler to handle `{:error, _}` channels without crashing

**Opt-in via `dns://` scheme** — only DNS targets trigger re-resolution. Static targets (`ipv4:`, `ipv6:`, `unix:`) are unaffected. No existing behaviour changes for `http://`, `https://`, or bare `host:port` targets.

## New options for `connect/2`

| Option | Default | Description |
|--------|---------|-------------|
| `:resolve_interval` | 30000 | Base interval between DNS re-resolutions (ms) |
| `:max_resolve_interval` | 300000 | Backoff cap (ms) |
| `:min_resolve_interval` | 5000 | Rate-limit floor for `resolve_now/1` (ms) |

## Note on `persistent_term`

The existing codebase uses `:persistent_term` for zero-copy channel passing to client stubs. This works well for infrequent updates, but `persistent_term` updates trigger a **global garbage collection pass across all BEAM processes** ([Erlang docs](https://www.erlang.org/doc/apps/erts/persistent_term.html)). With periodic re-resolution, channel reconciliation runs every 30s+ per connection.

To mitigate this, we guard `persistent_term.put` calls to only write when the picked channel actually changes — skipping redundant writes on no-op re-resolution cycles. A future improvement would be migrating to ETS with `read_concurrency: true`, which has no global GC cost on writes while maintaining the same(or close) read performance for `pick_channel`.

## Test coverage (28 tests)

| # | Category | Test |
|---|----------|------|
| 1-4 | Core | Scale-up, scale-down, no-op, complete replacement |
| 5-8 | Errors | DNS failure, recovery, empty addresses, recovery after empty |
| 9-10 | Stability | pick_channel during/after re-resolution |
| 11 | Timer | Repeated cycles accumulate changes |
| 12 | Edge | Non-DNS targets skip re-resolution |
| 13 | Edge | Disconnect kills linked resolver cleanly |
| 14 | Edge | LB error during re-resolution — connection survives |
| 15 | Edge | Port change on same host detected |
| 16 | Backoff | Doubles on failure, resets on success, caps at max |
| 17 | Rate limit | resolve_now calls coalesced within min_resolve_interval |
| 18 | Telemetry | :stop on success, :error on failure/empty |
| 19-20 | Persistence | Fallback to healthy channel, retry failed channels |
| 21 | Async | Connection stays responsive during slow DNS |
| 22 | Refresh | GenServer survives :refresh picking a failed channel |

## Test plan

- [x] All 28 re-resolution tests pass (10 consecutive runs, 0 flakes)
- [x] Full test suite: 244 tests, 0 failures
- [x] End-to-end tested with Docker Compose (scale up/down/outage/recovery)
- [x] End-to-end tested on Kubernetes with minikube + headless service + CoreDNS

### DIFF
--- a/grpc_client/lib/grpc/client/connection.ex
+++ b/grpc_client/lib/grpc/client/connection.ex
@@ -101,7 +101,8 @@ defmodule GRPC.Client.Connection do
           base_resolve_interval: non_neg_integer(),
           max_resolve_interval: non_neg_integer(),
           min_resolve_interval: non_neg_integer(),
-          last_resolve_at: integer() | nil
+          last_resolve_at: integer() | nil,
+          resolve_timer_ref: reference() | nil
         }
 
   defstruct virtual_channel: nil,
@@ -116,7 +117,8 @@ defmodule GRPC.Client.Connection do
             base_resolve_interval: 30_000,
             max_resolve_interval: 300_000,
             min_resolve_interval: 5_000,
-            last_resolve_at: nil
+            last_resolve_at: nil,
+            resolve_timer_ref: nil
 
   def child_spec(initial_state) do
     %{
@@ -145,9 +147,12 @@ defmodule GRPC.Client.Connection do
 
     # Only schedule periodic re-resolution for DNS targets — static targets
     # (ipv4:, ipv6:, unix:) always resolve to the same addresses.
-    if state.resolver && state.target && dns_target?(state.target) do
-      Process.send_after(self(), :re_resolve, state.resolve_interval)
-    end
+    state =
+      if state.resolver && state.target && dns_target?(state.target) do
+        schedule_re_resolve(state)
+      else
+        state
+      end
 
     {:ok, state}
   end
@@ -336,8 +341,7 @@ defmodule GRPC.Client.Connection do
         do_re_resolve(resolver, target, adapter, opts, %{state | last_resolve_at: now})
       end
 
-    Process.send_after(self(), :re_resolve, state.resolve_interval)
-    {:noreply, state}
+    {:noreply, schedule_re_resolve(state)}
   end
 
   def handle_info(:re_resolve, state), do: {:noreply, state}
@@ -485,20 +489,48 @@ defmodule GRPC.Client.Connection do
               }
 
             _ ->
-              %{state | real_channels: real_channels, lb_state: picked_lb_state}
+              # LB picked a channel that failed to connect. Fall back to any
+              # healthy channel so persistent_term doesn't hold a dead ref.
+              fallback_to_healthy_channel(state, real_channels, picked_lb_state)
           end
 
         {:error, _} ->
-          %{state | real_channels: real_channels}
+          fallback_to_healthy_channel(state, real_channels, state.lb_state)
       end
     else
-      %{state | real_channels: real_channels}
+      fallback_to_healthy_channel(state, real_channels, state.lb_state)
     end
   end
 
+  defp fallback_to_healthy_channel(state, real_channels, lb_state) do
+    ref = state.virtual_channel.ref
+
+    case Enum.find_value(real_channels, fn {_k, v} -> match?({:ok, _}, v) && v end) do
+      {:ok, healthy_channel} ->
+        :persistent_term.put({__MODULE__, :lb_state, ref}, healthy_channel)
+
+        %{
+          state
+          | real_channels: real_channels,
+            lb_state: lb_state,
+            virtual_channel: healthy_channel
+        }
+
+      nil ->
+        Logger.warning("No healthy channels available after re-resolution")
+        :persistent_term.erase({__MODULE__, :lb_state, ref})
+        %{state | real_channels: real_channels, lb_state: lb_state}
+    end
+  end
+
+  defp schedule_re_resolve(state) do
+    if state.resolve_timer_ref, do: Process.cancel_timer(state.resolve_timer_ref)
+    ref = Process.send_after(self(), :re_resolve, state.resolve_interval)
+    %{state | resolve_timer_ref: ref}
+  end
+
   defp dns_target?(target) do
-    uri = URI.parse(target)
-    uri.scheme == "dns" or is_nil(uri.scheme)
+    URI.parse(target).scheme == "dns"
   end
 
   defp via(ref) do
@@ -708,13 +740,16 @@ defmodule GRPC.Client.Connection do
 
         {"ipv4:#{uri.host}:#{uri.port}", opts, @insecure_scheme}
 
-      # Compatibility mode: host:port or unix:path
+      # Compatibility mode: host:port or unix:path.
+      # Bare host:port defaults to dns:// scheme (matching the Resolver docs:
+      # "If no scheme is specified, dns is assumed"), so DNS-based targets
+      # get proper resolution and periodic re-resolution.
       uri.scheme in [nil, ""] ->
         scheme = if opts[:cred], do: @secure_scheme, else: @insecure_scheme
 
         case String.split(target, ":") do
           [host, port] ->
-            {"ipv4:#{host}:#{port}", opts, scheme}
+            {"dns://#{host}:#{port}", opts, scheme}
 
           [path] ->
             {"unix://#{path}", opts, "unix"}
--- a/grpc_client/lib/grpc/client/connection.ex
+++ b/grpc_client/lib/grpc/client/connection.ex
@@ -80,6 +80,13 @@ defmodule GRPC.Client.Connection do
   @insecure_scheme "http"
   @secure_scheme "https"
   @refresh_interval 15_000
+  @default_resolve_interval 30_000
+  @default_max_resolve_interval 300_000
+  @default_min_resolve_interval 5_000
+
+  # Telemetry event names
+  @resolve_stop_event [:grpc, :client, :resolve, :stop]
+  @resolve_error_event [:grpc, :client, :resolve, :error]
 
   @type t :: %__MODULE__{
           virtual_channel: Channel.t(),
@@ -87,7 +94,14 @@ defmodule GRPC.Client.Connection do
           lb_mod: module() | nil,
           lb_state: term() | nil,
           resolver: module() | nil,
-          adapter: module()
+          adapter: module(),
+          target: String.t() | nil,
+          connect_opts: keyword(),
+          resolve_interval: non_neg_integer(),
+          base_resolve_interval: non_neg_integer(),
+          max_resolve_interval: non_neg_integer(),
+          min_resolve_interval: non_neg_integer(),
+          last_resolve_at: integer() | nil
         }
 
   defstruct virtual_channel: nil,
@@ -95,7 +109,14 @@ defmodule GRPC.Client.Connection do
             lb_mod: nil,
             lb_state: nil,
             resolver: nil,
-            adapter: GRPC.Client.Adapters.Gun
+            adapter: GRPC.Client.Adapters.Gun,
+            target: nil,
+            connect_opts: [],
+            resolve_interval: 30_000,
+            base_resolve_interval: 30_000,
+            max_resolve_interval: 300_000,
+            min_resolve_interval: 5_000,
+            last_resolve_at: nil
 
   def child_spec(initial_state) do
     %{
@@ -121,6 +142,13 @@ defmodule GRPC.Client.Connection do
     )
 
     Process.send_after(self(), :refresh, @refresh_interval)
+
+    # Only schedule periodic re-resolution for DNS targets — static targets
+    # (ipv4:, ipv6:, unix:) always resolve to the same addresses.
+    if state.resolver && state.target && dns_target?(state.target) do
+      Process.send_after(self(), :re_resolve, state.resolve_interval)
+    end
+
     {:ok, state}
   end
 
@@ -227,6 +255,20 @@ defmodule GRPC.Client.Connection do
     end
   end
 
+  @doc """
+  Triggers an immediate DNS re-resolution, subject to rate limiting.
+
+  Intended for use by health checks or heartbeat mechanisms that detect
+  a backend has gone away and want to force a fresh DNS lookup.
+  """
+  @spec resolve_now(Channel.t()) :: :re_resolve | {:error, :no_connection}
+  def resolve_now(%Channel{ref: ref}) do
+    case :global.whereis_name({__MODULE__, ref}) do
+      :undefined -> {:error, :no_connection}
+      pid -> send(pid, :re_resolve)
+    end
+  end
+
   @impl GenServer
   def handle_call({:disconnect, %Channel{adapter: adapter} = channel}, _from, state) do
     resp = {:ok, %Channel{channel | adapter_payload: %{conn_pid: nil}}}
@@ -279,6 +321,27 @@ defmodule GRPC.Client.Connection do
 
   def handle_info(:refresh, state), do: {:noreply, state}
 
+  def handle_info(
+        :re_resolve,
+        %{resolver: resolver, target: target, adapter: adapter, connect_opts: opts} = state
+      )
+      when not is_nil(resolver) and not is_nil(target) do
+    now = System.monotonic_time(:millisecond)
+
+    state =
+      if rate_limited?(state, now) do
+        Logger.debug("DNS re-resolution for #{target} rate-limited, skipping")
+        state
+      else
+        do_re_resolve(resolver, target, adapter, opts, %{state | last_resolve_at: now})
+      end
+
+    Process.send_after(self(), :re_resolve, state.resolve_interval)
+    {:noreply, state}
+  end
+
+  def handle_info(:re_resolve, state), do: {:noreply, state}
+
   def handle_info({:DOWN, _ref, :process, pid, reason}, state) do
     Logger.warning(
       "#{inspect(__MODULE__)} received :DOWN from #{inspect(pid)} with reason: #{inspect(reason)}"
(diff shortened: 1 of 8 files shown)


grpc_client/lib/grpc/client/dns_resolver.ex
```diff
@@ -0,0 +1,157 @@
+defmodule GRPC.Client.DnsResolver do
```
polvalente:
```suggestion
defmodule GRPC.Client.DNSResolver do
```

I generally prefer to not downcase acronyms

grpc_client/lib/grpc/client/connection.ex
```diff
@@ -227,6 +259,28 @@ defmodule GRPC.Client.Connection do
...
 
+  @doc """
+  Triggers an immediate DNS re-resolution, subject to rate limiting.
+
+  Intended for use by health checks or heartbeat mechanisms that detect
+  a backend has gone away and want to force a fresh DNS lookup.
+  """
+  @spec resolve_now(Channel.t()) :: :ok | {:error, :no_connection}
+  def resolve_now(%Channel{ref: ref}) do
+    case :global.whereis_name({__MODULE__, ref}) do
```
polvalente: Not sure about :global vs Process here. What's the reasoning for using :global?

grpc_client/lib/grpc/client/connection.ex
```diff
@@ -80,22 +80,31 @@ defmodule GRPC.Client.Connection do
...
         }
 
   defstruct virtual_channel: nil,
             real_channels: %{},
             lb_mod: nil,
             lb_state: nil,
             resolver: nil,
-            adapter: GRPC.Client.Adapters.Gun
+            adapter: GRPC.Client.Adapters.Gun,
+            target: nil,
```
polvalente:
```suggestion
            resolver_target: nil,
```

`target` by itself seems a bit too broad for the Connection struct.

We could maybe have `dns_resolver_opts` as the single new field. Thoughts?

grpc_client/lib/grpc/client/connection.ex
```diff
@@ -308,6 +370,143 @@ defmodule GRPC.Client.Connection do
...
+    added = MapSet.difference(new_keys, old_keys)
+    removed = MapSet.difference(old_keys, new_keys)
+
+    # Disconnect removed channels
+    real_channels =
+      Enum.reduce(MapSet.to_list(removed), state.real_channels, fn key, channels ->
+        case Map.get(channels, key) do
+          {:ok, ch} -> do_disconnect(adapter, ch)
+          _ -> :ok
+        end
```
polvalente: Is this expecting the interface of `Map.fetch` or is it just a coincidence that `Map.get` returns `{:ok, ch}` instead of `ch`?

grpc_client/lib/grpc/client/connection.ex
```diff
@@ -308,6 +370,143 @@ defmodule GRPC.Client.Connection do
...
+              # healthy channel so persistent_term doesn't hold a dead ref.
+              fallback_to_healthy_channel(state, real_channels, picked_lb_state)
+          end
+
+        {:error, _} ->
+          fallback_to_healthy_channel(state, real_channels, state.lb_state)
+      end
+    else
+      fallback_to_healthy_channel(state, real_channels, state.lb_state)
+    end
```
polvalente, line 483: I think there's too much nesting here. Could we extract some of this to defps?

grpc_client/lib/grpc/client/connection.ex
```diff
@@ -308,6 +370,143 @@ defmodule GRPC.Client.Connection do
...
+
+  # Only write to persistent_term when the channel actually changed.
+  # persistent_term updates trigger a global GC pass, so we skip
+  # redundant writes on no-change re-resolution cycles.
+  defp maybe_update_persistent_term(current_channel, new_channel) do
+    if current_channel != new_channel do
+      :persistent_term.put(
+        {__MODULE__, :lb_state, new_channel.ref},
+        new_channel
+      )
```
polvalente, line 515: Is this a risk of memory leaks? For instance, should we delete the current_channel entry?

polvalente (commented):
Gave this a very cursory pass. @sleipnir is probably the best to review this one in more depth, but I'll take another look later!

grpc_client/lib/grpc/client/dns_resolver.ex
cgreeno: cool will fix

grpc_client/lib/grpc/client/connection.ex
cgreeno: I thought global was already used by the upstream code for Connection registration. my bad. I will change it to match the style in the rest of the codebase. 

`GenServer.cast(via(ref), :resolve_now)               `                                                                                                                           
                                                                                                                                                                                  
  The only issue is error handling — GenServer.call(via(ref), ...) will raise if the process doesn't exist, but GenServer.cast is fire-and-forget now anyway so it silently so nobody is listening anyway I suppose.

grpc_client/lib/grpc/client/connection.ex
cgreeno: ya that is dumb - missed that - will edit to something like {:connected, ch} to make it more obvious.

grpc_client/lib/grpc/client/connection.ex
cgreeno: will fix

grpc_client/lib/grpc/client/connection.ex
cgreeno: I considered bundling into a single dns_resolver_opts field, but looking at the existing struct pattern it keeps things flat (each field is a top-level key, no nested maps). I was trying to match the existing style as much as possible. The three fields we added each serve a distinct purpose:

- resolver_target — checked in init to decide whether to start the DNS resolver
- connect_opts — used by reconcile_channels when connecting real channels (needs adapter_opts etc.), so it can't live inside a DNS-specific bundle
- dns_resolver_pid — the process reference for forwarding resolve_now

Bundling would only save resolver_target moving into a nested map, which adds indirection without reducing complexity. Happy to change if you feel strongly though. :)

polvalente: I'm ok with keeping it flat if we rename targets to resolver_targets!

grpc_client/lib/grpc/client/dns_resolver.ex
```diff
@@ -0,0 +1,157 @@
+defmodule GRPC.Client.DNSResolver do
```
sleipnir, line 1: I have a small concern about this module, mainly because it seems to introduce some overlap with responsibilities that were previously handled in a stateless way.

From what I understand, we might not need to introduce a new resolver here, and possibly not even a stateful approach. The `Connection` module already manages state and has access to both resolvers and load balancing, which seems aligned with the original design.

Initially, resolvers were intended to remain stateless, with `Connection` acting as the orchestrator—handling inputs, outputs, and coordination. It feels like a good portion of the logic being added here might already fit naturally within that existing structure.

Before moving forward with this PR, it might be worth revisiting this design choice to see if we can keep things simpler and more consistent with the original architecture.

sleipnir (changes requested):

grpc_client/lib/grpc/client/dns_resolver.ex
cgreeno: Thanks for the feedback @sleipnir. I appreciate you looking at this with the original architecture in mind. I actually started with exactly what you're describing: the resolve loop, backoff, rate limiting, and timer management all lived inside Connection as handle_info clauses. That was the first iteration.                                        
   
The problem I ran into was that Connection's GenServer was accumulating a lot of state and responsibility that had nothing to do with its core job of managing channels and load balancing. The re-resolution feature added 8 new state fields (resolve_interval, base_resolve_interval, max_resolve_interval, min_resolve_interval, last_resolve_at, resolve_timer_ref, resolve_task, resolve_start_time), 3 new handle_info clauses (for the timer, task completion, and task crashes), and 5 private functions (backoff, reset_backoff, rate_limited?, schedule_re_resolve, handle_resolve_result with telemetry). Connection went from being a clean orchestrator to also being a DNS retry state machine. 

  You can see that version here: https://github.com/cgreeno/grpc/commit/c5f10d1   --> still needs a bit of work but you get the idea of what I was trying.                                                                                               
   
 Moving it to a linked process made Connection simpler....  it went back to handling one message type (`{:dns_result, result}) ` and doing what it's good at: reconciling channels and updating the LB. The resolver itself stays stateless. DNSResolver doesn't replace the Resolver behaviour, it just wraps it with the periodic retry loop. The actual DNS lookup still goes through the existing stateless Resolver.resolve/1 contract.                                                                                                          
                                                                                                      
The benefit, at least to me, is separation of concerns: DNSResolver owns "when and how often to call resolve" (timers, backoff, rate limiting, telemetry). Connection owns "what to do with  the result" (channel reconciliation, LB re-init, persistent_term). Neither needs to know about the other's internals, and because they're linked, lifecycle management is automatic.                                                                                                                                                                      
                                                                                                      
That said, I'm happy to discuss further or adjust the approach if you feel strongly.  I can rollback the change to my previous commit.

sleipnir: What do you think about a mixed approach using delegation?

We could add a DNS attribute to the connection state map and handle it similarly to how we already deal with the load balancing module. In other words, within the `Connection` GenServer loop, we would delegate to a stateless DNS module.

My concern with introducing another process linked to `Connection` is that it creates an additional point of failure and adds more complexity in terms of supervision and lifecycle management. By using delegation instead, we avoid that overhead, don’t need to handle potential failures of a separate DNS process, and keep the overall design simpler.

Curious to hear your thoughts on this approach.

cgreeno: Thanks @sleipnir, I like the delegation pattern and I can see how it mirrors the `lb_mod `pattern. One thing I'd note is that either way I think we would need a separate process for the DNS I/O to avoid blocking Connection. That was actually one of the original reason I moved to an async approach. Sorry, I should have mentioned this more clearly in my initial reply above. The very first version had everything inline in Connection, and during testing with slow/unreachable DNS,  the GenServer became unresponsive.

NOW whether that's a Task per resolve call or a long-lived linked process is for sure debateable... I went with the linked process because it felt cleaner to have one stable process own the full lifecycle (timer, backoff, rate limiting, I/O) rather than Connection managing short-lived Tasks and tracking their refs alongside the delegation state.
                                                                                                                                                                           
 Keeping this in mind as well as DNS re-resolution is just the first piece...  the gRPC spec also defines: 

-  **SRV record resolution** — _grpc-lb._tcp.<host> lookups for port/priority/weight discovery                                                                                   
-   **Health checking** — periodic grpc.health.v1.Health/Check RPCs to each backend, removing unhealthy ones between DNS cycles
-  **xDS resolver** — a persistent gRPC stream to the control plane (Istio/Envoy/Traffic Director) for real-time endpoint updates

I planned to have a stab at implementing these as well. _sidebar_: Fresha is planning to move to a service mesh(possibly) and I just wanted to be ahead of the curve if we decide to go that way. :) 

With the delegation pattern, all of these would add their state maps and handle_info clauses to Connection. Health checking alone needs per-backend timers and RPC calls. xDS needs a persistent streaming connection. Delegation works for DNS but doesn't generalize to the other features in the gRPC spec, so we'd end up with mixed patterns. At least, that is how I am thinking about it. I think, once all 3 others are implemented, you would end up with some annoying complex routing logic. 

With the linked process pattern, each concern is its own process sending a common {:resolve_result, addresses} message back to Connection. Connection stays focused on channel  reconciliation and LB,  it doesn't care whether addresses came from DNS, SRV, health checks, or xDS. And if a DNS lookup hangs or an xDS stream crashes, Connection keeps serving traffic from its existing channels. 

I haven't given it a HUGE amount of thought but roughly this is what connection _could_ look like after DNS + SRV + health checking + xDS is added with delegation: 

  ```
defstruct virtual_channel: nil,                                                                                                                                                 
            real_channels: %{},                                                                                                                                                   
            lb_mod: nil,                                                                                                                                                          
            lb_state: nil,                                                                                                                                                        
            resolver: nil,                                                                            
            adapter: GRPC.Client.Adapters.Gun,
            # DNS re-resolution state                                                                                                                                             
            dns_state: %{},                                                                                                                                                       
            dns_task: nil,                                                                                                                                                        
            dns_timer_ref: nil,                                                                                                                                                   
            # SRV resolution state                                                                    
            srv_state: %{},                                                                                                                                                       
            srv_task: nil,
            srv_timer_ref: nil,                                                                                                                                                   
            # Health checking state (per backend)                                                     
            health_state: %{},
            health_tasks: %{},      # one per backend                                                                                                                             
            health_timer_refs: %{}, # one per backend
            # xDS control plane state                                                                                                                                             
            xds_state: %{},                                                                                                                                                       
            xds_stream_pid: nil,
            xds_subscriptions: %{}                                                                                                                                                
                                                                                                                                                                                  
  And the handle_info clauses:
                                                                                                                                                                                  
  # --- Existing upstream ---                                                                         
  def handle_info(:refresh, state)                                                                                                                                                
   
  # --- DNS re-resolution ---                                                                                                                                                     
  def handle_info(:dns_resolve, state)           # timer fires                                        
  def handle_info({ref, result}, state)          # dns task completed                                                                                                             
    when ref == state.dns_task.ref                                                                    
  def handle_info({:DOWN, ref, ...}, state)      # dns task crashed                                                                                                               
    when ref == state.dns_task.ref                                                                                                                                                
                                                                                                                                                                                  
  # --- SRV resolution ---                                                                                                                                                        
  def handle_info(:srv_resolve, state)           # timer fires                                        
  def handle_info({ref, result}, state)          # srv task completed
    when ref == state.srv_task.ref                                                                                                                                                
  def handle_info({:DOWN, ref, ...}, state)      # srv task crashed
    when ref == state.srv_task.ref                                                                                                                                                
                                                                                                      
  # --- Health checking (per backend) ---                                                                                                                                         
  def handle_info({:health_check, address}, state)        # timer fires for one backend
  def handle_info({ref, result}, state)                    # health task completed                                                                                                
    when Map.has_key?(state.health_tasks, ref)                                                                                                                                    
  def handle_info({:DOWN, ref, ...}, state)                # health task crashed
    when Map.has_key?(state.health_tasks, ref)                                                                                                                                    
                                                                                                                                                                                  
  # --- xDS control plane ---
  def handle_info(:xds_connect, state)                     # initial connect                                                                                                      
  def handle_info({:xds_update, resource_type, data}, state) # stream update                          
  def handle_info({:xds_error, reason}, state)             # stream error                                                                                                         
  def handle_info(:xds_reconnect, state)                   # reconnect timer                                                                                                      
                                                                                                                                                                                  
  # --- Catch-all ---                                                                                                                                                             
  def handle_info({:DOWN, _ref, ...}, state)    # now you need to figure out  WHICH task/stream this belongs to
```
 
**That's ~15 handle_info clauses, ~20 state fields, and the {ref, result} / {:DOWN, ...} handlers need to disambiguate which subsystem the task belongs to. Every time you add a  feature, Connection gets wider.   

As I said, have not given this a huge amount of thought so take it with a HUGE grain of salt and that I am using this example more to make my point. 😄** 

Compare that to my proposal of a separate process:

```
  defstruct virtual_channel: nil,                                                                                                                                                 
            real_channels: %{},                                                                       
            lb_mod: nil,
            lb_state: nil,
            resolver: nil,
            adapter: GRPC.Client.Adapters.Gun,
            dns_resolver_pid: nil,                                                                                                                                                
            srv_resolver_pid: nil,
            health_checker_pid: nil,                                                                                                                                              
            xds_resolver_pid: nil                                                                     
                                                                                                                                                                                  
  # Connection only handles results....  doesn't care who sent them                                                                                                                  
  def handle_info({:resolve_result, addresses}, state)                                                                                                                            
  def handle_info({:health_update, address, :healthy | :unhealthy}, state) 
```                                                                                                       
Two handle_info clauses. Four pid fields. Each subsystem manages its own timers, tasks, backoff, and failure recovery internally.

BUT I am not looking at the wider project, only what I feel is best in this instance. AND I am happy to take your lead and what the plan is for the long term architecture. I am open to whatever direction you think is best for the project long-term.  So honestly happy with either approach.  ❤️

Comment (cgreeno) [reply grpc_client/lib/grpc/client/dns_resolver.ex:1]: FYI - going on holiday tomorrow for a week so if I don't reply it is because my wife took away my laptop ;)

Comment (sleipnir) [reply grpc_client/lib/grpc/client/dns_resolver.ex:1]: When I originally designed the address resolver, my intention was to move us closer to an xDS-based approach. What seems to be missing now is an improvement to the GRPC.Client.Resolver callback: specifically, introducing a function that allows initializing the resolver implementation when needed.

This would enable resolvers that require long-lived or stateful interactions to be properly initialized within the Connection, without exposing unnecessary implementation details—whether they rely on a GenServer, Task, or any internal orchestration mechanism.

In addition, we would likely need another callback to handle state updates triggered by Connection events (such as disconnections, rebalancing, etc.). This update function should remain generic enough to support different resolver implementations.

With this approach, I believe we can achieve both goals: maintaining a clean design while ensuring each resolver has the state management capabilities it needs.

@cgreeno wdyt?

Comment (sleipnir) [reply grpc_client/lib/grpc/client/dns_resolver.ex:1]: This way, we keep Connection as the main orchestrator and central point of coordination, while delegating specific responsibilities to other modules only when necessary.

Connection remains responsible for lifecycle and high-level orchestration, while resolvers encapsulate their own logic and state. At the same time, cohesion is maintained through well-defined contracts and behaviors, ensuring that all components interact consistently without tight coupling. This avoids leaking implementation details and makes the system easier to evolve and reason about over time.

Comment (sleipnir) [reply grpc_client/lib/grpc/client/dns_resolver.ex:1]: > FYI - going on holiday tomorrow for a week so if I don't reply it is because my wife took away my laptop ;)

Enjoy your vacation!

Comment (cgreeno) [reply grpc_client/lib/grpc/client/dns_resolver.ex:1]: Ahh OK I get you. Ya let me have a stab at that....

How do you see Health checker working? Do you see Health checking as a different concern from Resolution? 

Something like:

```
defstruct virtual_channel: nil,                                                                     
            real_channels: %{},                                                                                                                                           
            lb_mod: nil,
            lb_state: nil,                                                                                                                                                
            resolver: nil,                                                                            
            adapter: GRPC.Client.Adapters.Gun,
            resolver_state: nil,      # opaque — DNS worker, xDS stream, whatever
            health_checker: nil,      # module                                                                                                                            
            health_checker_state: nil, # opaque — per-backend check processes                                                                                             
            connect_opts: [] 
```

cgreeno: Everything is addressed except disconnection event forwarding, which I will follow-up after this PR... the plumbing (update/2 callback) is in place but we haven't wired `:refresh` to call it on dead channels yet. Assume it is OK to address after?

cgreeno:
Closing in favour of a rebased PR onto current master (grpc_client/ was renamed to grpc/ in the 1.0.0-rc.1 release). New PR incoming.

PR #519, opened 2026-04-10
fix(gun): keep named channels alive after original caller exits

### LINKED ISSUE #518: Named Gun-backed channels die after the original caller exits
## Problem

Named Gun-backed channels outlive the process that created them, while the underlying Gun connection still follows the lifetime of that short-lived caller.

## Expected behavior

A named Gun-backed connection should behave like a true long-lived library-managed connection.

This should work reliably even if the original caller exits:

```elixir
# process 1
GRPC.Stub.connect(address, name: stable_name, adapter: GRPC.Client.Adapters.Gun)

...

# process 2
GRPC.Stub.connect(address, name: stable_name, adapter: GRPC.Client.Adapters.Gun)
```

Later calls through the same named channel should continue to use a healthy underlying Gun connection managed by the library, not by the transient process that originally called `connect/2`.

## Observed behavior

1. A short-lived process calls `GRPC.Stub.connect/2` with a `name`.
2. The first non-streaming RPC succeeds.
3. The caller process exits.
4. Later RPCs through the same named connection hit a Gun connection that is already closing or dead.

## Repro

The following regression test reproduces the issue on `master`.

```elixir
# grpc/test/grpc/integration/stub_test.exs

test "named Gun connections survive the original caller exiting" do
    parent = self()
    channel_name = {:named_gun_connection, make_ref()}

    {caller_pid, caller_ref} =
      spawn_monitor(fn ->
        {:ok, channel} = GRPC.Stub.connect("localhost:#{port}", name: channel_name)
        request = %Helloworld.HelloRequest{name: "first caller"}
        {:ok, reply} = Helloworld.Greeter.Stub.say_hello(channel, request)
        send(parent, {:initial_call_succeeded, reply.message})
      end)

    assert_receive {:initial_call_succeeded, "Hello, first caller"}
    assert_receive {:DOWN, ^caller_ref, :process, ^caller_pid, :normal}

    {:ok, channel} = GRPC.Stub.connect("localhost:#{port}", name: channel_name)

    request = %Helloworld.HelloRequest{name: "second caller"}
    assert {:ok, reply} = Helloworld.Greeter.Stub.say_hello(channel, request)
    assert reply.message == "Hello, second caller"
  end)
end
```

## Suspected root cause

The named channel abstraction is long-lived, but the real Gun connection still appears to be tied to the process that originally opened it.

Gun monitors the owner process of the connection and shuts the connection down gracefully when that owner exits, so the issue appears to be that the library-managed named connection does not fully align Gun ownership with the intended long-lived named channel lifecycle.

## Notes

This issue is specifically about named Gun-backed channels and connection ownership/lifecycle.

I would lean toward keeping the fix scoped to the Gun adapter, so the underlying Gun connection is no longer owned by the short-lived caller and Gun-specific lifecycle handling remains inside the adapter boundary rather than in `GRPC.Client.Connection`. This also aligns better with the Mint adapter.

smartinio:
## Problem

Named Gun-backed channels outlive the process that created them at the API level, while the underlying Gun socket still follows the short-lived caller's lifecycle.

In practice, this shows up as:

1. A short-lived process calls `GRPC.Stub.connect/2` with a `name`.
2. The first unary RPC succeeds.
3. The caller process exits.
4. Later RPCs through the same named connection hit a Gun connection that is already closing or effectively dead.

## Root cause

The named channel abstraction is long-lived, but the Gun connection is still tied to the process that originally opened it. That means the library-managed named connection does not fully control the lifecycle of the underlying Gun transport.

## Fix

Keep the fix scoped to the Gun adapter. Fixes #518

- Add `GRPC.Client.Adapters.Gun.ConnectionProcess` to own long-lived Gun connections for the adapter.
- Add `GRPC.Client.Adapters.Gun.StreamResponseProcess` to own per-stream Gun response state.
- Route Gun request/response lifecycle through adapter-owned processes instead of relying on caller mailboxes or caller-owned Gun connections.
- Keep `GRPC.Client.Connection` generic and avoid introducing Gun-specific handling there.

With this change, named Gun connections behave like true long-lived library-managed connections.

## Test plan

- [x] Added regression test that shows named Gun connection survives original caller process exiting and can be used later
- [x] Added focused tests for `Gun.StreamResponseProcess` lifecycle and unexpected-message handling
- [x] Ran the full `grpc` test suite (`mix test`)


sleipnir:
Hi @smartinio I will review this soon.

smartinio:
@sleipnir Hi, I noticed CI failed. Missed running CI on my fork before opening the PR. Amended a fix and re-ran CI in my fork. Should be green now.

interop/mix.lock
```diff
@@ -5,6 +5,7 @@
...
   "gen_stage": {:hex, :gen_stage, "1.3.2", "7c77e5d1e97de2c6c2f78f306f463bca64bf2f4c3cdd606affc0100b89743b7b", [:mix], [], "hexpm", "0ffae547fa777b3ed889a6b9e1e64566217413d018cabd825f786e843ffe63e7"},
   "googleapis": {:hex, :googleapis, "0.1.0", "13770f3f75f5b863fb9acf41633c7bc71bad788f3f553b66481a096d083ee20e", [:mix], [{:protobuf, "~> 0.12", [hex: :protobuf, repo: "hexpm", optional: false]}], "hexpm", "1989a7244fd17d3eb5f3de311a022b656c3736b39740db46506157c4604bd212"},
+  "grpc_core": {:hex, :grpc_core, "1.0.0-rc.1", "d82957bca32937bb52df06596cca7550783acc139a06b70202a982ef8b59490e", [:mix], [{:googleapis, "~> 0.1.0", [hex: :googleapis, repo: "hexpm", optional: false]}, {:jason, ">= 0.0.0", [hex: :jason, repo: "hexpm", optional: false]}, {:protobuf, "~> 0.14", [hex: :protobuf, repo: "hexpm", optional: false]}, {:telemetry, "~> 1.0", [hex: :telemetry, repo: "hexpm", optional: false]}], "hexpm", "c76233ea374421da562b5b022c22614e81f9cf862da93543cff93c37c085f136"},
```
smartinio, line 8: This lockfile was stale and got updated when I installed the interop deps locally to run the interop tests. Can drop if you want.

sleipnir:
Thanks @smartinio

PR #521, opened 2026-04-13
fix: extract target normalisation and fix IPv6 parsing crashes

arctarus:
## The bug

`GRPC.Client.Connection` contained two private functions, `normalize_target_and_opts/2` and `split_host_port/1`, responsible for parsing and normalising the raw target string passed to `GRPC.Stub.connect/2`.

Both functions had untested edge cases around IPv6 addresses that caused runtime crashes in production:

### 1. `normalize_target_and_opts/2` — `CaseClauseError` on IPv6 targets

The function delegated schemeless targets to a `cond` block that only handled two shapes: "a string with one colon" (host:port) and "a string with no colon" (unix socket). When an IPv6 address was passed — which always contains multiple colons — no branch matched and Elixir raised a `CaseClauseError`, crashing the `Connection` process before a connection was even attempted.

Affected inputs included any of:

```elixir
GRPC.Stub.connect("[::1]:50051", [])       # bracketed IPv6 — crash
GRPC.Stub.connect("::1:50051", [])         # bare IPv6 with port — crash
GRPC.Stub.connect("2001:db8::1:8080", [])  # full IPv6 address — crash
```

### 2. `split_host_port/1` — incorrect parsing of bracketed IPv6 without port

When a bracketed IPv6 address was provided with no port (e.g. `"[::1]"`), the regex for the `[addr]:port` pattern did not match. The fallback called `strip_scheme/1`, which splits on the first `:` character. Applied to `"[::1]"` this produced the host `":1]"` — a silently wrong value that would cause a connection failure at the adapter level with a confusing error.

## Why these bugs matter

Both failures happen before any network activity. The `Connection` process is a `GenServer`: a `CaseClauseError` during `init/1` means the process never starts, and the supervision tree restarts it immediately with the same arguments, causing a restart loop. The bracketed-IPv6-without-port bug silently constructs a malformed host and attempts to connect to it, producing a cryptic adapter error. Neither failure points back to IPv6 as the root cause, making both hard to diagnose in production.

While most Kubernetes clusters today connect to gRPC services via DNS names rather than raw IP addresses, there are real and growing scenarios where a caller passes an IPv6 address directly to `GRPC.Stub.connect/2`: local development and testing (`[::1]:50051` is the natural loopback on any IPv6-enabled machine), dual-stack clusters (stable since Kubernetes 1.21, increasingly enabled by cloud providers), telco and edge deployments (5G core network functions are specified over IPv6, where gRPC is a primary protocol), and bare-metal clusters going IPv6-first to avoid RFC 1918 exhaustion.

A bug that causes a silent restart loop with no actionable error message warrants a fix even if it affects a minority of users today.

### Extraction to `GRPC.Client.Connection.Target`

The normalisation logic was extracted from `Connection` into a dedicated internal module, `GRPC.Client.Connection.Target`, with two public functions:

- `normalize/2` — takes a raw target string and an optional `%GRPC.Credential{}`, returns `{norm_target, scheme, cred}`.
- `split_host_port/1` — takes a normalised target string, returns `{host, port}`.

This separation keeps `Connection` focused on lifecycle concerns and makes the parsing logic independently testable.

### IPv6 handling in `normalize/2`

The schemeless branch now explicitly handles three shapes before falling through:

1. **Bracketed IPv6 with port** — `"[::1]:50051"` matched by `~r/^\[([^\]]+)\]:(\d+)$/`, yielding `"ipv4:::1:50051"`.
2. **Bracketed IPv6 without port** — `"[::1]"` matched by a fallback `~r/\[([^\]]+)\]/` regex, stripping the brackets and returning the default port.
3. **Bare IPv6 with port** — `"2001:db8::1:8080"` handled by inspecting whether the last colon-separated segment is an integer, treating it as the port and joining the remaining segments as the address.

### Bracketed IPv6 fallback in `split_host_port/1`

The `_ ->` fallback when the `[addr]:port` regex does not match now uses a second regex `~r/\[([^\]]+)\]/` to extract the address from the brackets, rather than calling `strip_scheme/1` which blindly splits on the first `:`.

### Interface simplification

`normalize/2` previously accepted a `keyword()` opts list and returned the full opts back to the caller with `:cred` injected. The function only ever read one key from opts (`opts[:cred]`). The return type now reflects the module's actual concern:

```
# before
normalize(target, opts) :: {norm_target, norm_opts, scheme}

# after
normalize(target, cred) :: {norm_target, scheme, cred}
```

`Connection` now calls `Target.normalize(target, opts[:cred])` and reassembles its own state directly, keeping `opts` management in `Connection` where it belongs.

The `nimble_options` dependency that was previously added to validate `opts[:cred]` was removed, as it is no longer needed — the type is enforced by the `is_struct(cred, GRPC.Credential)` guard clause.

## Tests

A new test file, `grpc/test/grpc/client/connection/target_test.exs`, provides exhaustive coverage of `Target.normalize/2` and `Target.split_host_port/1`, including all IPv6 input shapes, scheme inference, credential propagation, and the `ArgumentError` raised when a credential is passed alongside an `http://` target.

The test suite was written first as failing tests to reproduce the original crashes, then made green by the implementation. The bracketed-IPv6-without-port bug in `split_host_port/1` was itself discovered by the new tests during the review pass.

### DIFF
--- a/grpc/lib/grpc/client/connection/endpoint_resolver.ex
+++ b/grpc/lib/grpc/client/connection/endpoint_resolver.ex
@@ -8,11 +8,11 @@ defmodule GRPC.Client.Connection.EndpointResolver do
   #
   #   * `"https://host:port"` – TLS, implicit :cred injection
   #   * `"http://host:port"`  – plain-text, rejects :cred
-  #   * `"host:port"`         – compatibility shorthand → ipv4
+  #   * `"host:port"`         – compatibility shorthand → ipv4/ipv6
   #   * `"path"`              – bare path → unix socket
   #   * `"dns://…"`, `"ipv4:…"`, `"ipv6:…"`, `"unix:…"`, etc. – passed through
-  #   * `"[::1]:port"`        – bracketed IPv6 with port → ipv4 normalised
-  #   * `"::1:port"`          – bare IPv6 with port → ipv4 normalised
+  #   * `"[::1]:port"`        – bracketed IPv6 with port → ipv6 normalised
+  #   * `"::1:port"`          – bare IPv6 with port → ipv6 normalised
 
   @insecure_scheme "http"
   @secure_scheme "https"
@@ -21,7 +21,7 @@ defmodule GRPC.Client.Connection.EndpointResolver do
   @doc """
   Normalises `target` and `cred`, returning `{norm_target, scheme, cred}`.
 
-  - `norm_target` – canonical target string for the resolver (e.g. `"ipv4:1.2.3.4:50051"`)
+  - `norm_target` – canonical target string for the resolver (e.g. `"ipv4:1.2.3.4:50051"`, `"ipv6:::1:50051"`)
   - `scheme`      – `"http"`, `"https"`, or `"unix"`
   - `cred`        – resolved `%GRPC.Credential{}`, or `nil` for plain-text targets
 
@@ -38,7 +38,7 @@ defmodule GRPC.Client.Connection.EndpointResolver do
       {"ipv4:localhost:50051", "http", nil}
 
       iex> GRPC.Client.Connection.EndpointResolver.normalize("[::1]:50051", nil)
-      {"ipv4:::1:50051", "http", nil}
+      {"ipv6:::1:50051", "http", nil}
 
   """
   @spec normalize(String.t(), GRPC.Credential.t() | nil) ::
@@ -50,13 +50,15 @@ defmodule GRPC.Client.Connection.EndpointResolver do
     cond do
       uri.scheme == @secure_scheme and uri.host ->
         resolved_cred = cred || default_ssl_option()
-        {"ipv4:#{uri.host}:#{uri.port}", @secure_scheme, resolved_cred}
+        prefix = resolver_prefix(uri.host)
+        {"#{prefix}:#{uri.host}:#{uri.port}", @secure_scheme, resolved_cred}
 
       uri.scheme == @insecure_scheme and uri.host ->
         if cred,
           do: raise(ArgumentError, "invalid option for insecure (http) address: :cred")
 
-        {"ipv4:#{uri.host}:#{uri.port}", @insecure_scheme, nil}
+        prefix = resolver_prefix(uri.host)
+        {"#{prefix}:#{uri.host}:#{uri.port}", @insecure_scheme, nil}
 
       # Compatibility mode: "host:port", bare path, or raw IPv6
       uri.scheme in [nil, ""] ->
@@ -155,10 +157,11 @@ defmodule GRPC.Client.Connection.EndpointResolver do
       String.starts_with?(target, "[") ->
         case Regex.run(~r/^\[([^\]]+)\]:(\d+)$/, target) do
           [_, addr, port] ->
-            {"ipv4:#{addr}:#{port}", scheme, cred}
+            {"ipv6:#{addr}:#{port}", scheme, cred}
 
           _ ->
-            {"ipv4:#{String.trim_leading(target, "[") |> String.replace("]", "")}", scheme, cred}
+            addr = target |> String.trim_leading("[") |> String.replace("]", "")
+            {"ipv6:#{addr}", scheme, cred}
         end
 
       String.contains?(target, ":") ->
@@ -169,19 +172,26 @@ defmodule GRPC.Client.Connection.EndpointResolver do
             case Integer.parse(port_str) do
               {_port, ""} ->
                 addr = parts |> Enum.drop(-1) |> Enum.join(":")
-                {"ipv4:#{addr}:#{port_str}", scheme, cred}
+                prefix = resolver_prefix(addr)
+                {"#{prefix}:#{addr}:#{port_str}", scheme, cred}
 
               _ ->
-                {"ipv4:#{target}", scheme, cred}
+                prefix = resolver_prefix(target)
+                {"#{prefix}:#{target}", scheme, cred}
             end
 
           _ ->
-            {"ipv4:#{target}", scheme, cred}
+            prefix = resolver_prefix(target)
+            {"#{prefix}:#{target}", scheme, cred}
         end
 
       true ->
         {"unix://#{target}", "unix", nil}
     end
+  end
+
+  defp resolver_prefix(host) when is_binary(host) do
+    if String.contains?(host, ":"), do: "ipv6", else: "ipv4"
   end
 
   defp strip_scheme(target) do
--- a/grpc/lib/grpc/client/connection/endpoint_resolver.ex
+++ b/grpc/lib/grpc/client/connection/endpoint_resolver.ex
@@ -191,7 +191,11 @@ defmodule GRPC.Client.Connection.EndpointResolver do
   end
 
   defp resolver_prefix(host) when is_binary(host) do
-    if String.contains?(host, ":"), do: "ipv6", else: "ipv4"
+    case :inet.parse_address(String.to_charlist(host)) do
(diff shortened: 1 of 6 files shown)


grpc/test/grpc/client/connection/target_test.exs
```diff
@@ -0,0 +1,284 @@
...
+
+  # ---------------------------------------------------------------------------
+  # Helpers
+  # ---------------------------------------------------------------------------
+
+  defp cred(opts \\ [verify: :verify_none]), do: %GRPC.Credential{ssl: opts}
+
+  # ---------------------------------------------------------------------------
+  # normalize/2 — https:// scheme
+  # ---------------------------------------------------------------------------
```
sleipnir:
```suggestion
```

grpc/test/grpc/client/connection/target_test.exs
```diff
@@ -0,0 +1,284 @@
+defmodule GRPC.Client.Connection.TargetTest do
+  use ExUnit.Case, async: true
+
+  alias GRPC.Client.Connection.Target
+
+  doctest GRPC.Client.Connection.Target
+
+  # ---------------------------------------------------------------------------
+  # Helpers
+  # ---------------------------------------------------------------------------
```
sleipnir:
```suggestion
```

grpc/lib/grpc/client/connection/target.ex
```diff
@@ -0,0 +1,235 @@
...
+        case Regex.run(~r/^\[([^\]]+)\]:(\d+)$/, target) do
+          [_, addr, port] ->
+            {"ipv4:#{addr}:#{port}", scheme, cred}
+
+          _ ->
+            {"ipv4:#{String.trim_leading(target, "[") |> String.replace("]", "")}", scheme, cred}
+        end
+
+      # Bare IPv6 or host:port — any colon present
+      # Heuristic: last colon-separated segment is all digits → it is the port
```
sleipnir:
```suggestion
```

grpc/lib/grpc/client/connection/target.ex
```diff
@@ -0,0 +1,235 @@
...
+      String.contains?(target, ":") ->
+        parts = String.split(target, ":")
+
+        case List.last(parts) do
+          port_str when byte_size(port_str) > 0 ->
+            case Integer.parse(port_str) do
+              {_port, ""} ->
+                # Last segment is a number → treat as port, rest is the address
```
sleipnir:
```suggestion
```

grpc/lib/grpc/client/connection/target.ex
```diff
@@ -0,0 +1,235 @@
...
+                addr = parts |> Enum.drop(-1) |> Enum.join(":")
+                {"ipv4:#{addr}:#{port_str}", scheme, cred}
+
+              _ ->
+                # Last segment is not a number → no recognisable port, pass as-is
```
sleipnir:
```suggestion
```

grpc/lib/grpc/client/connection/target.ex
```diff
@@ -0,0 +1,235 @@
...
+
+      # No colon → bare host, use default port
+      true ->
+        {target, @default_port}
+    end
+  end
+
+  # ---------------------------------------------------------------------------
+  # Private
+  # ---------------------------------------------------------------------------
```
sleipnir:
```suggestion
```

sleipnir (changes requested):
Hi, thanks for the PR.

Could you reduce the number of comments in the code that don't translate into some high-level explanation? I've seen many comments that are self-explanatory within the code itself, below the comment.

sleipnir:
@arctarus We also have an error in the CI.

arctarus:
@sleipnir feedback applied and errors fixed! Let me know if there is something else I can do.

grpc/lib/grpc/client/connection/target.ex
```diff
@@ -0,0 +1,216 @@
+defmodule GRPC.Client.Connection.Target do
```
sleipnir: I would call this module EndpointResolver because it does more than just normalization. Is that okay?

grpc/lib/grpc/client/connection/target.ex
```diff
@@ -0,0 +1,216 @@
...
+
+      # URI.parse misreads "hostname:port" as scheme="hostname", host=nil.
+      # Detect this: a real resolver scheme always has a host OR uses "://"
+      # notation. If host is nil and the scheme is not a known gRPC resolver
+      # prefix, treat it as a schemeless host:port shorthand.
+      is_nil(uri.host) and uri.scheme not in ["ipv4", "ipv6", "dns", "unix", "xds"] ->
+        scheme = if cred, do: @secure_scheme, else: @insecure_scheme
+        normalize_schemeless(target, scheme, cred)
+
+      # Anything else (dns://host, ipv4:host:port, ipv6:…, unix:…, xds:…)
```
sleipnir: Self-explanatory comment. You can remove it.

sleipnir (changes requested):

grpc/lib/grpc/client/connection/endpoint_resolver.ex
```diff
@@ -0,0 +1,216 @@
...
+  @spec normalize(String.t(), GRPC.Credential.t() | nil) ::
+          {String.t(), String.t(), GRPC.Credential.t() | nil}
+  def normalize(target, cred)
+      when is_binary(target) and (is_nil(cred) or is_struct(cred, GRPC.Credential)) do
+    uri = URI.parse(target)
+
+    cond do
+      uri.scheme == @secure_scheme and uri.host ->
+        resolved_cred = cred || default_ssl_option()
+        {"ipv4:#{uri.host}:#{uri.port}", @secure_scheme, resolved_cred}
```
sleipnir: What happens if a port isn't provided?

grpc/lib/grpc/client/connection/endpoint_resolver.ex
```diff
@@ -0,0 +1,216 @@
...
+
+      uri.scheme == @insecure_scheme and uri.host ->
+        if cred,
+          do: raise(ArgumentError, "invalid option for insecure (http) address: :cred")
+
+        {"ipv4:#{uri.host}:#{uri.port}", @insecure_scheme, nil}
```
sleipnir: Why are we inferring IPv4 for everything?

sleipnir (changes requested):

grpc/lib/grpc/client/connection/endpoint_resolver.ex
arctarus: In the cases of `http` or `https` schemas, the port is set by `URI.parse/1`  otherwise it will go to a different branch which extracts the port from the raw string or omits it.

grpc/lib/grpc/client/connection/endpoint_resolver.ex
arctarus: You're right! the ipv4: prefix was being hardcoded for all addresses, including IPv6 ones. This was inherited from the original `Connection` code, but since IPv6 inputs crashed before reaching the resolver it was never visible.

I've added a resolver_prefix/1 helper that checks whether the host contains : (which only IPv6 addresses do) and emits the correct prefix:

"127.0.0.1:50051" → "ipv4:127.0.0.1:50051"
"[::1]:50051" → "ipv6:::1:50051"
"2001:db8::1:8080" → "ipv6:2001:db8::1:8080"
This ensures the target is routed to IPv6.resolve/1 instead of IPv4.resolve/1 downstream.

Thanks!

sleipnir (approved):

sleipnir:
@arctarus thank you 
Great job!

PR #523, opened 2026-04-17
Feature [client]: add supervised connection pool with legacy mode opt…

### LINKED ISSUE #522: [Continuation of #345] Supervised gRPC Connection Pool — actual implementation
**Background**

This issue continues the discussion from #345 , which was closed by #350 . That PR added documentation guidance for managing multiple gRPC connections, but left an actual in-library implementation as future work. This PR delivers that implementation.

@Nezteb (the original author of #345) and I had been planning to work on this together for very long time. The implementation is now ready and I'm submitting it for review.

**What this adds**

A supervised connection pool for gRPC channels built directly into the client, with:

  - Multiple concurrent HTTP/2 connections managed under an Erlang/OTP supervisor tree
  - Load distribution across connections, addressing the :max_concurrent_streams bottleneck that #345 highlighted (Mint's default cap of 100 concurrent streams per channel)
  - A legacy mode opt-out flag so existing users are not affected by the change — the pool is strictly opt-in
  - Automatic reconnection with exponential backoff (building on the work from #512)

**Motivation**

  The original issue correctly identified that a single reused channel becomes a bottleneck at scale. The recommended approach from that discussion — "keeping multiple connections open, monitoring their status and spreading the load across them" — is exactly what this implements, natively in the library rather than requiring users to roll their own solution or reach for external libraries like conn_grpc.

**Disclaimer**

  The core pooling logic predates this PR. It has been running in production as a standalone internal library for approximately two years. The integration work to fit it cleanly into the existing elixir-grpc codebase was assisted by AI tooling, but all original pooling logic and design decisions were written by a human. I want to be transparent about this so reviewers can weigh the maturity of the logic separately from the integration code as well as giving it a bit more attention to the review.

PR Ref: **https://github.com/elixir-grpc/grpc/pull/523**

VladoPlavsic:
Introduces a built-in HTTP/2 connection pool for every GRPC.Stub.connect/2
call. Connections are checked out before each RPC and returned afterwards,
with stream counts tracked per connection. Supports configurable pool size,
overflow capacity, per-connection stream limits, and optional health-check
pings — all via a :pool option on connect/2.

Adds legacy mode (config :grpc, pool_enabled: false) to restore the
pre-pool single-connection behaviour for gradual migration. Adds nil
max_overflow support for unbounded overflow connections. Includes full
unit and integration test coverage for pool, overflow, and legacy paths.

Closes: https://github.com/elixir-grpc/grpc/issues/522

<details>
    <summary>Full patch notes</summary>
    
# TL;DR - For legacy mode see Legacy Mode section at the bottom

# Connection Pool — Patch Notes

## Summary

This PR introduces a built-in **connection pool** for the Elixir gRPC client. Every call to
`GRPC.Stub.connect/2` (or `GRPC.Client.Connection.connect/2`) now starts a supervised pool of
long-lived HTTP/2 connections instead of opening a single raw connection. Callers share those
connections transparently: the client checks out a connection before each RPC and returns it
afterwards, keeping stream counts accurate. Pool size, overflow capacity, and per-connection
stream limits are all configurable via a single `:pool` option.

---

### New modules

| Module | Role |
|:-------|:-----|
| `GRPC.Client.Pool` | Public API: `start_for_address/4`, `stop_for_address/1`, `checkout/1`, `checkin/2` |
| `GRPC.Client.Pool.Config` | Struct that carries pool configuration (size, overflow, max streams) |
| `GRPC.Client.Pool.Supervisor` | OTP supervisor that owns one pool; restarted transiently |
| `GRPC.Client.Pool.Server` | `GenServer` that tracks channels, open-stream counts, and leases |
| `GRPC.Client.Pool.Implementation` | Pure functional core — all state transitions, no side effects |
| `GRPC.Client.Pool.HealthCheck.DynamicSupervisor` | Manages health-check workers for pool connections |
| `GRPC.Client.Pool.HealthCheck.Server` | Sends periodic gRPC health-check pings; exits cleanly when its connection drops |

### Modified modules

**`GRPC.Channel`** — added `pool: reference() | nil` field. The field is `nil` on raw channels and
holds a pool reference on virtual channels returned by `connect/2`.

**`GRPC.Client.Application`** — registers `GRPC.Client.Pool.Registry` at startup so pool
supervisors and servers can be found by `pool_ref` without additional setup.

**`GRPC.Client.Connection`** — replaced direct `adapter.connect` calls with
`GRPC.Client.Pool.start_for_address/4`. The `do_disconnect` private function now calls
`GRPC.Client.Pool.stop_for_address/1` instead of `adapter.disconnect/1`. Added `:pool` to the
`Keyword.validate!` options list with sensible defaults.

**`GRPC.Stub`** — the `call/5` implementation now goes through `acquire_channel/2` (pool
checkout) and `release_channel/2` (pool checkin) bracketing every RPC, replacing the previous
ad-hoc `pick_channel` + liveness check.

---

## Benefits

- **Reduced connection overhead** — HTTP/2 connections are long-lived and multiplexed; no
  per-call handshake cost.
- **Controlled concurrency** — `max_streams` caps how many concurrent requests share a single
  connection. When a connection is saturated, the pool opens an overflow connection instead of
  stacking unlimited streams.
- **Automatic recovery** — when a connection process exits unexpectedly, the pool removes it,
  clears its leases, and lets Gun re-establish the session in the background. Callers do not need
  to handle reconnection logic.
- **Back-pressure** — when the pool (including overflow) is fully exhausted,
  `GRPC.Status.resource_exhausted()` is returned immediately rather than stacking requests
  indefinitely.
- **Observability** — lease tracking (streams per channel, leases by PID and monitor ref) makes
  pool state inspectable at any time via `:sys.get_state/1`.

---

### 1. `GRPC.Stub.connect/2` returns a virtual channel, not a raw connection

Previously the returned `%GRPC.Channel{}` had `adapter_payload` populated with the underlying
connection PID (e.g. `%{conn_pid: pid}`). That field is now `nil` on the virtual channel —
internal connection details belong to the pool, not the caller.

**Before:**
```elixir
{:ok, %GRPC.Channel{adapter_payload: %{conn_pid: pid}}} = GRPC.Stub.connect("localhost:50051")
```

**After:**
```elixir
{:ok, %GRPC.Channel{pool: pool_ref}} = GRPC.Stub.connect("localhost:50051")
# pool_ref is a reference(); adapter_payload is nil on the virtual channel
```

### 2. `GRPC.Stub.disconnect/2` return value changed

Previously the disconnected channel carried `adapter_payload: %{conn_pid: nil}`. Now it carries
`pool: nil`.

**Before:**
```elixir
{:ok, %GRPC.Channel{adapter_payload: %{conn_pid: nil}}} = GRPC.Stub.disconnect(channel)
```

**After:**
```elixir
{:ok, %GRPC.Channel{pool: nil}} = GRPC.Stub.disconnect(channel)
```

### 3. Connection-process crash messages are no longer forwarded to the caller

Previously, if the underlying Gun connection process crashed, an `{:EXIT, pid, reason}` message
could propagate to the process that opened the connection (when it was trap-exiting). The pool
now owns all connections and handles those exits internally. Callers will no longer receive
connection-crash messages.

---

## New API: `:pool` option

`GRPC.Stub.connect/2` and `GRPC.Client.Connection.connect/2` accept a new `:pool` keyword
option:

```elixir
GRPC.Stub.connect("localhost:50051",
  pool: %{
    size: 2,           # number of persistent connections (default: 1)
    max_overflow: 5,   # max extra connections when pool is saturated (default: 0)
    max_streams: 100   # max concurrent streams per connection (default: nil = unlimited)
  }
)
```

All three keys are optional; omitting `:pool` entirely uses `%{size: 1, max_overflow: 0, max_streams: nil}`.

Setting `max_overflow: nil` removes the overflow cap entirely — the pool opens new connections on
demand whenever all existing connections are saturated, with no upper bound beyond what the server
and OS allow.

---

## Migration Guide

1. **Remove `adapter_payload` pattern matches on `connect` results.** If your code inspects
   `channel.adapter_payload.conn_pid` after connecting, remove that assertion. The pool now
   manages connection PIDs internally.

2. **Update `disconnect` pattern matches.** Replace checks for `adapter_payload: %{conn_pid: nil}`
   with `pool: nil`.

3. **Remove manual connection-crash handling.** If your process was trapping exits and handling
   `{:EXIT, conn_pid, _}` to detect drops, that logic is no longer needed. The pool handles
   reconnection automatically.

4. **Tune pool size for your workload** (optional). The default of one connection is conservative.
   For services with moderate to high RPC concurrency, consider increasing `:size` or setting
   `:max_overflow` to absorb traffic bursts without opening unbounded connections.

## Legacy Mode (opt-out)

If you need to temporarily disable the pool and restore the pre-pool behaviour, set the following
in your config:

```elixir
config :grpc, pool_enabled: false
```

With the pool disabled:

- `GRPC.Stub.connect/2` calls `adapter.connect` directly and returns a channel with
  `adapter_payload: %{conn_pid: pid}` and `pool: nil` — exactly as before.
- RPC calls use the old `pick_channel` + `Process.alive?` liveness check instead of pool
  checkout/checkin.
- `GRPC.Stub.disconnect/1` calls `adapter.disconnect` directly and returns a channel with
  `adapter_payload: %{conn_pid: nil}`.
- No pool supervisor, server, or registry entries are created.

The default is `pool_enabled: true`. This option is intended as a temporary escape hatch while
migrating — not as a permanent configuration.

---

</details>

### DIFF
--- a/grpc/lib/grpc/client/connection.ex
+++ b/grpc/lib/grpc/client/connection.ex
@@ -324,7 +324,7 @@ defmodule GRPC.Client.Connection do
     end
   end
 
-  defp pool_enabled?, do: Application.get_env(:grpc, :pool_enabled, true)
+  defp pool_enabled?, do: Application.get_env(:grpc, :pool_enabled, false)
 
   defp do_disconnect(adapter, %Channel{pool: nil} = ch), do: adapter.disconnect(ch)
 
--- a/grpc/lib/grpc/client/connection.ex
+++ b/grpc/lib/grpc/client/connection.ex
@@ -140,6 +140,14 @@ defmodule GRPC.Client.Connection do
     * `:codec` – request/response codec (default: `GRPC.Codec.Proto`)
     * `:compressor` / `:accepted_compressors` – message compression
     * `:headers` – default metadata headers
+    * `:pool` – connection pool options (map):
+      * `:size` – number of persistent connections (default: `1`)
+      * `:max_overflow` – max extra connections when pool is saturated; `nil` for no limit (default: `0`)
+      * `:max_streams` – max concurrent streams per connection; `nil` for no client-side limit (default: `nil`)
+      * `:health_check_enabled` – send periodic gRPC health-check pings (default: `false`)
+
+  The pool can be disabled entirely by setting `config :grpc, pool_enabled: false` in your
+  application config, which restores the pre-pool behaviour (single direct connection per `connect/2` call).
 
   Returns:
 
@@ -228,26 +236,15 @@ defmodule GRPC.Client.Connection do
   end
 
   @impl GenServer
-  def handle_call({:disconnect, %Channel{adapter: adapter} = channel}, _from, state) do
+  def handle_call({:disconnect, %Channel{adapter: adapter, pool: nil} = channel}, _from, state) do
     resp = {:ok, %Channel{channel | adapter_payload: %{conn_pid: nil}}}
-    :persistent_term.erase({__MODULE__, :lb_state, channel.ref})
+    do_handle_disconnect(adapter, channel, resp, state)
+  end
 
-    if Map.has_key?(state, :real_channels) do
-      Enum.map(state.real_channels, fn
-        {_key, {:ok, ch}} ->
-          do_disconnect(adapter, ch)
-
-        _ ->
-          :ok
-      end)
-
-      keys_to_delete = [:real_channels, :virtual_channel]
-      new_state = Map.drop(state, keys_to_delete)
-
-      {:reply, resp, new_state, {:continue, :stop}}
-    else
-      {:reply, resp, state, {:continue, :stop}}
-    end
+  @impl GenServer
+  def handle_call({:disconnect, %Channel{adapter: adapter} = channel}, _from, state) do
+    resp = {:ok, %Channel{channel | pool: nil}}
+    do_handle_disconnect(adapter, channel, resp, state)
   end
 
   @impl GenServer
@@ -312,15 +309,27 @@ defmodule GRPC.Client.Connection do
     {:global, {__MODULE__, ref}}
   end
 
-  defp do_disconnect(adapter, channel) do
-    adapter.disconnect(channel)
-  rescue
-    _ ->
-      :ok
-  catch
-    _type, _value ->
-      :ok
+  defp do_handle_disconnect(adapter, channel, resp, state) do
+    :persistent_term.erase({__MODULE__, :lb_state, channel.ref})
+
+    if Map.has_key?(state, :real_channels) do
+      Enum.map(state.real_channels, fn
+        {_key, {:ok, ch}} -> do_disconnect(adapter, ch)
+        _ -> :ok
+      end)
+
+      {:reply, resp, Map.drop(state, [:real_channels, :virtual_channel]), {:continue, :stop}}
+    else
+      {:reply, resp, state, {:continue, :stop}}
+    end
   end
+
+  defp pool_enabled?, do: Application.get_env(:grpc, :pool_enabled, true)
+
+  defp do_disconnect(adapter, %Channel{pool: nil} = ch), do: adapter.disconnect(ch)
+
+  defp do_disconnect(_adapter, %Channel{pool: pool_ref}),
+    do: GRPC.Client.Pool.stop_for_address(pool_ref)
 
   defp build_initial_state(target, opts) do
     opts =
@@ -333,7 +342,8 @@ defmodule GRPC.Client.Connection do
         codec: GRPC.Codec.Proto,
         compressor: nil,
         accepted_compressors: [],
-        headers: []
+        headers: [],
+        pool: %{size: 1, max_overflow: 0, max_streams: nil, health_check_enabled: false}
       )
 
     resolver = Keyword.get(opts, :resolver, GRPC.Client.Resolver)
@@ -369,10 +379,10 @@ defmodule GRPC.Client.Connection do
 
     case resolver.resolve(norm_target) do
       {:ok, %{addresses: addresses, service_config: config}} ->
-        build_balanced_state(base_state, addresses, config, lb_policy_opt, norm_opts, adapter)
+        build_balanced_state(base_state, addresses, config, lb_policy_opt, norm_opts)
 
       {:error, _reason} ->
-        build_direct_state(base_state, norm_target, norm_opts, adapter)
+        build_direct_state(base_state, norm_target, norm_opts)
     end
   end
 
@@ -397,8 +407,7 @@ defmodule GRPC.Client.Connection do
          addresses,
          config,
          lb_policy_opt,
-         norm_opts,
-         adapter
+         norm_opts
        ) do
     lb_policy =
       cond do
@@ -419,7 +428,7 @@ defmodule GRPC.Client.Connection do
         {:ok, {prefer_host, prefer_port}, new_lb_state} = lb_mod.pick(lb_state)
 
         real_channels =
-          build_real_channels(addresses, base_state.virtual_channel, norm_opts, adapter)
+          build_real_channels(addresses, base_state.virtual_channel, norm_opts)
 
         key = build_address_key(prefer_host, prefer_port)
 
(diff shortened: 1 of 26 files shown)


sleipnir:
Hello @VladoPlavsic, thank you for the PR.

We have some open PRs for the gRPC client and we need to review them before taking a look here.

grpc/lib/grpc/client/connection.ex
```diff
@@ -312,16 +309,28 @@ defmodule GRPC.Client.Connection do
...
+        _ -> :ok
+      end)
+
+      {:reply, resp, Map.drop(state, [:real_channels, :virtual_channel]), {:continue, :stop}}
+    else
+      {:reply, resp, state, {:continue, :stop}}
+    end
   end
 
+  defp pool_enabled?, do: Application.get_env(:grpc, :pool_enabled, true)
```
Nezteb:
```suggestion
  defp pool_enabled?, do: Application.get_env(:grpc, :pool_enabled, false)
```

Since this would be a new, perhaps we keep the default to disabled?

Thoughts? 😄

grpc/lib/grpc/client/pool.ex
```diff
@@ -0,0 +1,71 @@
...
+  @moduledoc false
+
+  alias GRPC.Channel
+  alias GRPC.Client.Pool.Config
+
+  @call_timeout 7_000
+
+  @spec start_for_address(Channel.t(), term(), non_neg_integer(), keyword()) ::
+          {:ok, Channel.t()} | {:error, any()}
+  def start_for_address(%Channel{} = vc, host, port, norm_opts) do
```
Nezteb, line 11: Previously `RPC.Client.Connection` was doing this:

```elixir
defp connect_real_channel(%Channel{scheme: "unix"} = vc, path, port, opts, adapter) do
  %Channel{vc | host: path, port: port}
  |> adapter.connect(opts[:adapter_opts])
end

defp connect_real_channel(%Channel{} = vc, host, port, opts, adapter) do
  %Channel{vc | host: host, port: port}
  |> adapter.connect(opts[:adapter_opts])
end
```

Should `GRPC.Client.Pool.start_for_address/4` also take care of the path vs host thing?

I worry doing this would break for anyone who tries to do something like:
```elixir
GRPC.Stub.connect("unix:///tmp/grpc.sock")
```

grpc/lib/grpc/client/pool/implementation.ex
```diff
@@ -0,0 +1,290 @@
...
+  defp remove_lease_from_leases_by_channel_id(
+         %State{leases_by_channel_id: leases} = state,
+         channel,
+         caller_pid
+       ) do
+    old_leases = Map.get(leases, channel.id, [])
+    old_lease = Enum.find(old_leases, fn %State.Lease{caller_pid: pid} -> pid == caller_pid end)
+
+    new_leases =
+      Enum.reject(old_leases, fn %State.Lease{caller_pid: pid} -> pid == caller_pid end)
```
Nezteb, line 205: Can there be more than one matching "old" lease? If so, it returns just one and then removes all of them, instead of just the one it finds. Would that be an issue? 🤔 

```suggestion
    old_leases = Map.get(leases, channel.id, [])
    old_lease = Enum.find(old_leases, fn %State.Lease{caller_pid: pid} -> pid == caller_pid end)

    new_leases = List.delete(old_leases, old_lease)
```

VladoPlavsic: gRPC call is blocking, and we lease to a PID that requires a channel, so even if you were to do from a single process something like 

```
Task.async(fn -> execute_your_grpc end)
```

we would lease a channel to the task process, so, to answer your question, no, there shouldn't (can't?) be a case where we lease a channel to same PID twice.

grpc/lib/grpc/client/pool.ex
VladoPlavsic:
```
defp connect_real_channel(%Channel{scheme: "unix"} = vc, path, port, opts, adapter) do
  %Channel{vc | host: path, port: port}
  |> adapter.connect(opts[:adapter_opts])
end

defp connect_real_channel(%Channel{} = vc, host, port, opts, adapter) do
  %Channel{vc | host: host, port: port}
  |> adapter.connect(opts[:adapter_opts])
end
```

as far as I can tell these two functions are identical - except for second parameter name. That's why I decided to remove it (the resolution happens before this function call)

grpc/lib/grpc/client/connection.ex
VladoPlavsic: IDK, I mean, if you are upgrading the library I think you should look into what's new (1)

If maintainers agree, pool should be the default mode (default sets it to single connection no overflow - basically same as before) - so making this config default to false whilst being bound to remove it after full transition (possibly next release) I think would cause even more confusion (2)

I guess let's leave this for maintainers to decide?

sleipnir:
@VladoPlavsic you can resolve the conflicts?

VladoPlavsic:
@sleipnir up

sleipnir:
@VladoPlavsic sorry for long delay, but you can resolve the conflicts again? If you do this quickly, I promise to prioritize this PR.

Ping @Nezteb

VladoPlavsic:
@sleipnir sorry, I have a lot on my plate atm…and then I’m on pto, probably won’t be able to come to this until end of October

PR #526, opened 2026-04-23
fix(connection): make named channels node-local

smartinio:
## Problem

Named channels are currently registered with `:global`, which makes them effectively cluster-global across connected BEAM nodes.

In distributed deployments, this results in:
1. One node creates the named channel registration
2. Another node trying to connect the same named channel sees `:already_started`, then fails to resolve the channel locally and returns `{:error, :no_connection}`.

## Root cause

`GRPC.Client.Connection` registers named channels globally via `:global`, but the selected channel state is still read from node-local `:persistent_term`. I.e. registration is shared across nodes, while channel state is not.

## Fix

Scope named channels to the current BEAM node instead of sharing them across connected nodes.

- Start a node-local `Registry` for client connections
- Register named channels through that `Registry` instead of `:global`
- Update connection tests to look up the node-local registration

## Test plan

- [x] Added regression test that verifies two connected peer nodes don't conflict on the same name
- [x] Added a small shim for `:peer` to pass OTP24 interop tests, since `:peer` is an OTP25+ module


grpc/lib/grpc/client/connection.ex
```diff
@@ -6,8 +6,8 @@ defmodule GRPC.Client.Connection do
   A `Conn` process manages one or more underlying gRPC connections
   (`GRPC.Channel` structs) and exposes a virtual channel to be used by
   client stubs. The orchestration process runs as a `GenServer` registered
-  globally (via `:global`), so only one orchestrator exists per connection
-  in a BEAM node.
+  in a node-local `Registry`, so named channels are scoped to the current
+  BEAM node rather than shared across connected nodes.
```
sleipnir:
```suggestion
  BEAM node.
```

sleipnir (changes requested):

sleipnir (approved):

sleipnir:
@polvalente Would you also like to take a look here, or can we continue?

smartinio:
@sleipnir @polvalente Hi, this has been approved for 2 weeks now. Any chance we could get it merged? No rush, just making sure it's not forgotten.

sleipnir:
> @sleipnir @polvalente Hi, this has been approved for 2 weeks now. Any chance we could get it merged? No rush, just making sure it's not forgotten.

No, we haven't forgotten about you.

sleipnir:
@smartinio I've done the merge, but we need to analyze other PRs before releasing a new version in hex.

smartinio:
@sleipnir Thanks. No rush, we cannot install the hex version until https://github.com/elixir-grpc/grpc/pull/519 is released. We're currently running that branch in production.

PR #527, opened 2026-04-24
feat: ETS-backed LB with per-request pick, remove persistent_term writes on reconcile

cgreeno:
## Summary

Follow-up to #520 putting the LB module in control of the pick decision.

`pick_channel/2` now goes through `lb_mod.pick(lb_state)` on every RPC. The hot path is fully lock-free:

    pick_channel/2
      → :persistent_term.get({Connection, ref})    # compile-time-constant lookup
      → lb_mod.pick(lb_state)                       # :atomics.add_get + elem (RR) or :ets.lookup (PF)
      → {:ok, channel}

**NOTE: this PR does cause performance degradation see metrics below for more details**

No GenServer call on the hot path. No `:persistent_term.put` on reconcile (and so no global GC pass when backends scale in/out).

The shape generalizes: weighted RR, least-requests, health-aware routing, and xDS all need richer per-backend state than a single atom — "LB owns its ETS table + atomics" carries them without forcing each strategy to invent state plumbing.

### Architecture

| Where | What it holds | Lifecycle |
| --- | --- | --- |
| `:persistent_term` keyed `{Connection, ref}` | `{lb_mod, lb_state}` | put once on `init/1`, erase on `disconnect/1` + `terminate/2` |
| Per-Connection ETS table (one per Connection, owned by the GenServer) | `{:channels, channels_tuple}` | replaced wholesale on reconcile |
| `:atomics` ref (in `lb_state`) | RR cursor (unsigned 64-bit) | reset to 0 on reconcile |

The `lb_state` published to `persistent_term` (the `tid` + `atomics` ref) is stable for the life of the connection — reconcile mutates ETS and atomics in place rather than republishing, so `:persistent_term.put` fires once per connect, not every 30s. That keeps the [global GC pass](https://www.erlang.org/doc/apps/erts/persistent_term.html) `put` triggers off the steady-state path.

### Micro-benchmark (Apple M1 Max)

**TL;DR:** the pick path adds ~290 ns over the pre-PR cached read (52 ns → 345 ns) in exchange for a real per-request RR rotation across N channels. The reconciliation write path replaces `:persistent_term.put` with `:ets.insert + :atomics.put`, eliminating the global GC pass that previously fired every 30s per Connection.

All comparisons are vs the **pre-PR baseline**: a single cached `:persistent_term.get` of one channel, no rotation — what shipped before #520.

**Single-process pick**

| Variant | Throughput | Latency | Notes |
| --- | ---: | ---: | --- |
| Pre-PR baseline (`:persistent_term.get`, cached) | 19.08M ips | 52 ns | no rotation, single channel |
| NEW pick (`persistent_term + ets + atomics`) | 2.90M ips | 345 ns | full per-request RR rotation |

**32-process concurrent pick**

| Variant | Total throughput | Notes |
| --- | ---: | --- |
| Pre-PR baseline (`:persistent_term.get`) | ~35M picks/sec | reads scale linearly |
| NEW pick | ~6.0M picks/sec | atomics counter scales — limited by cache-line ping-pong, not a software lock |

The ~6× delta is the cost of moving from a pure read to a real RR rotation. Critically, `:atomics.add_get/3` is a hardware CAS with no key-level or table-level lock, so the contention story is just cache coherence rather than a serialization point. ~6M picks/sec at 32 procs is well above any realistic per-Connection RPC volume.

**Single-process reconciliation write**

| Variant | Throughput | Latency | Notes |
| --- | ---: | ---: | --- |
| Pre-PR baseline (`:persistent_term.put`) | 5.37M ips | 186 ns | **+ global GC pass across every BEAM process** |
| NEW reconcile (`:ets.insert + :atomics.put`) | 1.90M ips | 527 ns | local, no global side effects |

The headline `persistent_term.put` latency hides the real cost. Per the [BEAM docs](https://www.erlang.org/doc/apps/erts/persistent_term.html), every write forces a global GC pass — every process in the node has to scan its heap. With periodic DNS re-resolution this fires every 30s per Connection; the new design pays ~340 ns more per reconcile and zero GC.

## Notable details

- `RoundRobin.pick/1` rescues `ArgumentError` — `disconnect/1` may delete the LB ETS table mid-pick and we want a tagged error rather than a propagated crash.
- `:persistent_term.put` runs last in `init/1` — `init/1` crashes skip `terminate/2`, so publishing earlier would leak entries on a failing `resolver.init/2`.
- `disconnect/1` calls `:persistent_term.erase` **before** closing transports, so in-flight picks fail fast with `:no_connection` rather than racing a dying adapter.
- `terminate/2` also erases (idempotent on the disconnect path — `:persistent_term.erase` returns `false` on a missing key without raising). Covers crash exits and supervisor stops.
- `rebalance_after_reconcile/3` does **not** republish to `persistent_term` — the published `lb_state` shape (tid + atomics ref) is stable for the life of the connection, so `update/2`'s in-place mutation is visible to picks immediately.

## Test coverage

- 19 LB unit tests (PickFirst, RoundRobin including 16-process concurrency stress).
- `disconnect/1` and `terminate/2` both free the LB's ETS table.
- 50 concurrent pickers × 100 picks vs `disconnect/1` — no crashes.
- 30 concurrent pickers × 200 picks during a shrinking reconcile — no crashes; survivors-only after settle.
- 500 connect/disconnect cycles — zero `{Connection, _}`-keyed `persistent_term` leak, zero ETS table leak (mirrors the #509 `persistent_term` leak regression).

Full suite: 286 tests, 0 failures.


grpc/lib/grpc/client/load_balacing/round_robin.ex
```diff
@@ -1,22 +1,57 @@
...
   end
 
   @impl true
-  def pick(%{addresses: addresses, index: idx, n: n} = state) do
-    %{address: host, port: port} = Enum.fetch!(addresses, idx)
+  # `disconnect/1` may delete the table between a caller's registry lookup and pick.
+  def pick(%{tid: tid} = state) do
+    case :ets.lookup(tid, @channels_key) do
+      [{@channels_key, channels}] when tuple_size(channels) > 0 ->
+        idx = :ets.update_counter(tid, @index_key, {2, 1})
```
hansihe: This locks:
* Key when `write_concurrency` enabled on table
* Whole table when `write_concurrency` is not enabled on the table

Essentially this serializes

grpc/lib/grpc/client/load_balacing/registry.ex
```diff
@@ -0,0 +1,31 @@
+defmodule GRPC.Client.LoadBalancing.Registry do
```
hansihe: Weird placement for this module, this module should mainly contain behaviour implementations.

grpc/lib/grpc/client/connection.ex
```diff
@@ -244,12 +241,15 @@ defmodule GRPC.Client.Connection do
   """
   @spec pick_channel(Channel.t(), keyword()) :: {:ok, Channel.t()} | {:error, term()}
   def pick_channel(%Channel{ref: ref} = _channel, _opts \\ []) do
-    case :persistent_term.get({__MODULE__, :lb_state, ref}, nil) do
-      nil ->
-        {:error, :no_connection}
+    case Registry.lookup(ref) do
```
hansihe: Could be worse, but why do we need two distinct tables?

Right now we have:
* One ets table as registry
* One ets table for the lb strategy

This seems unnecessary

This registry ets table might be possible to eliminate entirely if we name our connection-level ets table according to the connection `ref` (module name). Then we can look it up directly. Strategy could be a standard key in the ets table.

Alternatively, possibly better would be to use `persistent_term` at the top level. Register `ref` -> `{registry_mod, regitry_config}`, then that is a simple `persistent_term` lookup instead of `ets`. `registry_config` would be registry specific, but would usually be the id of the LB ETS table.

grpc/lib/grpc/client/load_balacing/round_robin.ex
cgreeno: Fixed and updated

grpc/lib/grpc/client/connection.ex
cgreeno: Ya changed it to 1 ets table

cgreeno:
@sleipnir I got a friend to review it first to try and save you some time. Happy to go in whatever direction you want however. :)

sleipnir:
> @sleipnir I got a friend to review it first to try and save you some time. Happy to go in whatever direction you want however. :)

Ok let me known when you are done!

cgreeno:
@sleipnir - Done - he reviewed it earlier. :)

sleipnir:
> @sleipnir - Done - he reviewed it earlier. :)

Okay, now I'm going to review it.

cgreeno:
@sleipnir - how is it looking? No Pressure ;)

cgreeno:
@sleipnir - gental nudge :)

sleipnir:
> @sleipnir - gental nudge :)

Sorry for the delay, I've been busy, but I haven't forgotten any of the PRs. We'll address them all. I'll review everything next week.

grpc/lib/grpc/client/load_balacing/pick_first.ex
```diff
@@ -1,16 +1,42 @@
...
+  @current_key :current
+
   @impl true
   def init(opts) do
-    case Keyword.get(opts, :addresses, []) do
-      [] -> {:error, :no_addresses}
-      addresses -> {:ok, %{addresses: addresses, current: hd(addresses)}}
+    case Keyword.get(opts, :channels, []) do
+      [] ->
+        {:error, :no_channels}
```
sleipnir: This is a detail, but it could break the system for those already using previous versions. We can maintain this pre-existing contract with no_addressess.

grpc/lib/grpc/client/load_balacing/pick_first.ex
```diff
@@ -1,16 +1,42 @@
...
+        {:ok, %{tid: tid}}
     end
   end
 
   @impl true
-  def pick(%{current: %{address: host, port: port}} = state) do
-    {:ok, {host, port}, state}
+  def pick(%{tid: tid} = state) do
+    case :ets.lookup(tid, @current_key) do
+      [{@current_key, nil}] -> {:error, :no_channels}
```
sleipnir: The same here :no_addresses

grpc/lib/grpc/client/load_balacing/pick_first.ex
```diff
@@ -1,16 +1,42 @@
...
+      [{@current_key, channel}] -> {:ok, channel, state}
+      [] -> {:error, :no_channels}
```
sleipnir: no_addresses

grpc/lib/grpc/client/load_balacing/round_robin.ex
```diff
@@ -1,22 +1,45 @@
 defmodule GRPC.Client.LoadBalancing.RoundRobin do
+  @moduledoc "Round-robin load balancer (ETS for the channel tuple, `:atomics` for the cursor)."
```
sleipnir: Regarding implementation details, we can simplify this documentation.

grpc/lib/grpc/client/load_balacing/round_robin.ex
```diff
@@ -1,22 +1,45 @@
...
   @behaviour GRPC.Client.LoadBalancing
 
+  @channels_key :channels
+
   @impl true
   def init(opts) do
-    addresses = Keyword.get(opts, :addresses, [])
+    case Keyword.get(opts, :channels, []) do
+      [] ->
+        {:error, :no_channels}
```
sleipnir: The same :no_addresses

grpc/lib/grpc/client/connection.ex
```diff
@@ -625,44 +574,36 @@ defmodule GRPC.Client.Connection do
...
+           | lb_mod: lb_mod,
+             real_channels: real_channels
+         }}
     end
   end
 
   defp build_direct_state(%__MODULE__{} = base_state, norm_target, norm_opts, adapter) do
     {host, port} = split_host_port(norm_target)
     vc = base_state.virtual_channel
+    lb_mod = GRPC.Client.LoadBalancing.PickFirst
```
sleipnir, line 602: Why this is fixed?

sleipnir (changes requested):

sleipnir:
Hi @cgreeno sorry for long delay. Just few comments and adjustments is necessary.

cgreeno:
Pushed fixes addressing the review + the OTP 25.x CI failure (`6609f8f`):

- **CI crash fix:** the failing `dns_resolver_test` was hitting a `CaseClauseError`. `build_direct_state` (the resolution-failure fallback) used a local `split_host_port/1` that only matched a bare `host:port`, but `norm_target` is always scheme-prefixed (`dns://`, `ipv4:`, `ipv6:`), so it crashed. It now delegates to the existing `EndpointResolver.split_host_port/1`, which already handles every scheme-prefixed and IPv6 form. Removed the broken duplicate.
- Restored the pre-existing `:no_addresses` error contract (was `:no_channels`) in `PickFirst`, `RoundRobin`, `Connection`, and tests.
- Simplified the `RoundRobin` moduledoc to drop the ETS/atomics implementation detail.
- Documented why `build_direct_state` hardcodes `PickFirst` (a direct target is a single endpoint, so any LB policy only ever has one channel).

**Follow-up worth discussing (left out of scope here):** when a `dns://` target genuinely fails to resolve, the fallback still dials the literal hostname (e.g. `//my-service.local`). It no longer crashes, but dialing a name that just NXDOMAIN'd is unlikely to succeed — returning `{:error, reason}` may be the more honest behavior. Happy to address in this PR or a follow-up, whichever you prefer.

cgreeno:
@sleipnir - any chance I can get another review please. :)

sleipnir:
> @sleipnir - any chance I can get another review please. :)

of course my dear

sleipnir:
> **Follow-up worth discussing (left out of scope here):** when a `dns://` target genuinely fails to resolve, the fallback still dials the literal hostname (e.g. `//my-service.local`). It no longer crashes, but dialing a name that just NXDOMAIN'd is unlikely to succeed — returning `{:error, reason}` may be the more honest behavior. Happy to address in this PR or a follow-up, whichever you prefer.

We can handle this in another PR!

sleipnir:
@cgreeno Add a note about the performance degradation in the readme and we'll proceed with the merge.

sleipnir (approved):

sleipnir:
Hi @cgreeno you can take a look on failure tests?

cgreeno:
@sleipnir all OK now?

sleipnir:
Hi @cgreeno 
Hard work! Thank you for your dedication <3

PR #529, opened 2026-04-27
Correctly cast boolean query params

### LINKED ISSUE #528: Boolean query params with Transcoding are not correctly deserialized
**Describe the bug**
Query params are always strings, but proto bool fields require actual Elixir booleans. Protobuf.JSON.from_decoded throws {:bad_bool, field, value} when given a string, causing the request to fail silently with a 204 No Content response. (hat tip to @shawnjoseph for the report and fix in my fork)

**To Reproduce**
A gRPC service that takes a message like
```
message FooRequest {
  bool truthy = 4;
}

...

service Sbraf {
  rpc GetFoo(GetFooRequest) returns (GetFooResponse) {}
}
```

**Expected behavior**
A boolean query param should be allowed and be cast appropriately

**Logs**
If applicable, add logs to help explain your problem.

**Protos**
If applicable, add protos to help explain your problem.

**Versions:**
- OS: macOS Tahoe
- Erlang: 27
- Elixir: 1.17.3
- mix.lock(grpc, gun, cowboy, cowlib): current master

**Additional context**

the-destro:
Fixes #528 
(hat tip to @shawnjoseph for the find and fix)


sleipnir (approved):
Tks @the-destro

PR #535, opened 2026-05-13
Fix typing of RPCError

aj-foster:
Hello 👋🏼 

Just a quick fix to the typing of the `RPCError` struct. It appears to be expected that the `details` field will be `nil`, for example when calling `exception/2`. This makes Dialyzer happy for those who observe it.

❤️


sleipnir (approved):

sleipnir:
Thanks @aj-foster

PR #538, opened 2026-06-03
feat(server): surface inbound gRPC deadline on GRPC.Server.Stream

cgreeno:
## Summary

Surfaces the inbound request's gRPC deadline on `GRPC.Server.Stream` so server-side interceptors and handlers can read it. Today the Cowboy handler decodes the `grpc-timeout` header **only** to arm its internal cancellation timer — the deadline is never exposed, so an interceptor has nothing to read.

This is the enabling primitive for deadline budget propagation (clamping each outbound call to the time the caller still has). It also brings elixir-grpc in line with other gRPC implementations, which all expose the inbound deadline to the handler — `ctx.Deadline()` (Go), `Context.current().getDeadline()` (Java), `servicer_context.time_remaining()` (Python).

## Changes

**`GRPC.Server.Stream`**
- New `deadline` field: absolute monotonic-time (ms) by which the inbound request must complete; `nil` when the caller set no deadline.
- New `remaining_ms/1`: milliseconds left (clamped at 0 so an expired deadline never goes negative), or `:infinity` when there is no deadline.

**`GRPC.Server.Adapters.Cowboy.Handler.init/2`**
- Decodes `grpc-timeout` **once**, before building the stream, and reuses the value for both the deadline and the cancellation timer (previously decoded inline only for the timer, after the stream was already built and the RPC process spawned).

## Design notes

- `GRPC.Transport.Utils.decode_timeout/1` returns **milliseconds** (the existing timer already feeds its result straight into `Process.send_after/3`), so it's safe to reuse for the absolute-deadline math.
- The deadline uses `System.monotonic_time(:millisecond)`, never wall-clock time, so it is immune to clock adjustments, and each hop can compute `remaining = deadline − now` locally.
- **Backwards compatible:** callers that ignore `deadline` are unaffected; `remaining_ms/1` returns `:infinity` for the `nil` default, so adapters that don't set it behave safely.

## Testing

- **Unit:** `grpc_server/test/grpc/server/stream_test.exs` covers `remaining_ms/1` (`:infinity` when no deadline; time-left; clamped at 0).
- **End-to-end:** `grpc/test/grpc/integration/server_deadline_test.exs` asserts a request with `grpc-timeout` reaches the server with `stream.deadline` set, and a request without the header (`timeout: :infinity`) has `deadline == nil`. It is tagged `:skip` because the `grpc` app depends on the **published** `grpc_server` hex package, which doesn't yet carry the field. It passes when `grpc` is path-linked to local source — verified locally, including that the existing timeout-cancellation test (`stub_test.exs` "returns error when timeout") still passes against the modified handler, i.e. no behavioural change to timeouts.
- `grpc_server`: full suite green (192 tests, 0 failures), `mix format` clean, `--warnings-as-errors` clean.

## Out of scope

Just the library primitive. Reading it and clamping outbound calls is a follow-up interceptor/plug, not part of this PR.


cgreeno:
heya @sleipnir - sorry to keep tagging you in new PR's. We have had a few incidents in Fresha with retry storms and downstream process running well after the initial call dropped the connection. I would really like to create a plug that passes the timing though calls so we know we can drop connections after the initial call is no longer waiting.

sleipnir:
> heya @sleipnir - sorry to keep tagging you in new PR's. We have had a few incidents in Fresha with retry storms and downstream process running well after the initial call dropped the connection. I would really like to create a plug that passes the timing though calls so we know we can drop connections after the initial call is no longer waiting.

Hi, no problem tagging me, that's what I'm here for. Let me review this PR and I'll get back to you with more comments. This weekend I'll be working on launching a new release, so it might take me a while to get back here (a few days).

sleipnir:
Tks @cgreeno

PR #552, opened 2026-07-01
fix(grpc): remove specs and enforce compiler warnings

smartinio:
## Problem

Dialyzer reports `GRPC.Stub.connect/2` as returning a value outside its own spec.

`connect/2` returns a `GRPC.Channel` struct, but `GRPC.Channel.t()` is stricter than the channel states built by the client connection code. In our project, running dialyzer errors on calls to `connect/2` which sees a broken public type:

```text
::warning file=lib/grpc/client/connection.ex,line=190,title=call::The function call finalize_connection will not succeed.
::warning file=lib/grpc/client/connection.ex,line=193,title=call::The function call finalize_connection will not succeed.
::warning file=lib/grpc/client/connection.ex,line=354,title=no_return::Function finalize_connection/2 has no local return.
::warning file=lib/grpc/client/connection.ex,line=355,title=call::The function call pick_channel will not succeed.
```

`GRPC.Client.Stream.t()` has the same issue for fields that start as `nil` and are filled later.

## Root cause

The `Channel` and `Stream` structs support incrementally populating their values, but the public types describe only fully populated state.

## Fix

[Project direction](https://github.com/elixir-grpc/grpc/pull/551#issuecomment-4859531775) is to stop supporting Dialyzer and instead leverage Elixir compiler warnings:
- Remove explicit `@spec` annotations from source and README examples
- Compile with warnings as errors in CI
- Keep tests running with warnings as errors in CI
- Fix the Gun stream response state update that Elixir 1.20 flags under warnings as errors

## Review attention

Elixir's type system did not catch the issue related to stream trailer parsing flagged by Dialyzer, so I'm deferring that to a future PR to avoid conflating things:
> ::warning file=lib/grpc/client/adapters/gun.ex,line=435,title=guard_fail::The guard clause can never succeed.

## Test plan

- [x] Pulled in `grpc` and `grpc_core` from this branch in our project and successfully ran dialyzer

### DIFF
--- a/grpc/lib/grpc/client/adapters/gun/stream_response_process.ex
+++ b/grpc/lib/grpc/client/adapters/gun/stream_response_process.ex
@@ -22,12 +22,10 @@ defmodule GRPC.Client.Adapters.Gun.StreamResponseProcess do
           done: boolean()
         }
 
-  @spec start_link() :: GenServer.on_start()
   def start_link do
     GenServer.start_link(__MODULE__, [])
   end
 
-  @spec await(pid(), timeout()) :: tuple()
   def await(pid, timeout) do
     GenServer.call(pid, {:await, timeout}, :infinity)
   catch
@@ -110,7 +108,7 @@ defmodule GRPC.Client.Adapters.Gun.StreamResponseProcess do
   defp push_message(%{waiter: {from, timeout_ref}} = state, message, terminal?) do
     cancel_timeout(timeout_ref)
     GenServer.reply(from, message)
-    new_state = %{state | waiter: nil, done: state.done or terminal?}
+    new_state = %{state | waiter: nil, done: terminal?}
 
     if terminal? do
       {:stop, :normal, new_state}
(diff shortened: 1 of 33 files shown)


grpc/lib/grpc/client/adapters/gun/stream_response_process.ex
```diff
@@ -120,7 +118,7 @@ defmodule GRPC.Client.Adapters.Gun.StreamResponseProcess do
   end
 
   defp push_message(%{messages: messages} = state, message, terminal?) do
-    {:noreply, %{state | messages: :queue.in(message, messages), done: state.done or terminal?}}
+    {:noreply, %{state | messages: :queue.in(message, messages), done: terminal?}}
```
polvalente, line 121: Gotta double-check this change and the one above it

polvalente (approved):

grpc/lib/grpc/client/adapters/gun/stream_response_process.ex
polvalente: Ok, this change makes sense! There's a clause above catching done: true, so the type checker knows done is false in all clauses below

polvalente:
Thanks for pivoting into this!

PR #554, opened 2026-07-11
fix: stop leaking client supervisor children on failed and closed connections

### LINKED ISSUE #531: `GRPC.Client.Supervisor` persistent memory growth
**Describe the bug**

I have an application that connects to hundreds of devices and reads data from them over GRPC. I use short-lived connections for each, e.g. create a channel with `GRPC.Stub.connect/2`, then disconnect with `GRPC.Stub.disconnect/1`. For some reason, when I run `:recon.proc_count(:memory, 10)`, `GRPC.Client.Supervisor` is always at the top of the list. When I call `:recon.bin_leak(10)` after the application runs for a while, `GRPC.Client.Supervisor` is also at the top of the list:

```
iex(5)> :recon.bin_leak(10)
[
  {#PID<0.210.0>, -1401,
   [
     GRPC.Client.Supervisor,
     {:current_function, {:gen_server, :loop, 5}},
     {:initial_call, {:proc_lib, :init_p, 5}}
   ]},
   ...
]
```

All my processes to retrieve this data live in short-lived supervised `Task`s. Maybe this is a feature, but I don't think the process memory should be increasing like this over time? Maybe I'm doing something wrong? I don't know.

**To Reproduce**

1. spin up connections to a bunch of GRPC servers via `GRPC.Stub.connect/1`
2. make some function calls on each one
3. close the connections with `GRPC.Stub.disconnect/1`

**Expected behavior**

Binary memory is freed, at least to some degree.

**Logs**

Can provide if needed

**Versions:**
- OS: MacOS Sequoia 15.6
- Erlang: 28.3.1
- Elixir: 1.18.4
- mix.lock(grpc, gun, cowboy, cowlib):

```
  "grpc": {:hex, :grpc, "1.0.0-rc.1", "790336fc827f0a22521d443c1c89e941502ce1e3ef09160c7e4798b2e148b53d", [:mix], [{:castore, "~> 0.1 or ~> 1.0", [hex: :castore, repo: "hexpm", optional: true]}, {:grpc_core, "~> 1.0.0-rc.1", [hex: :grpc_core, repo: "hexpm", optional: false]}, {:gun, "~> 2.0", [hex: :gun, repo: "hexpm", optional: false]}, {:mint, "~> 1.5", [hex: :mint, repo: "hexpm", optional: false]}], "hexpm", "c60dcda2fb143769ba496bd86d33420a9e44d633555bc781deaba3668138372b"},
  "grpc_core": {:hex, :grpc_core, "1.0.0-rc.1", "d82957bca32937bb52df06596cca7550783acc139a06b70202a982ef8b59490e", [:mix], [{:googleapis, "~> 0.1.0", [hex: :googleapis, repo: "hexpm", optional: false]}, {:jason, ">= 0.0.0", [hex: :jason, repo: "hexpm", optional: false]}, {:protobuf, "~> 0.14", [hex: :protobuf, repo: "hexpm", optional: false]}, {:telemetry, "~> 1.0", [hex: :telemetry, repo: "hexpm", optional: false]}], "hexpm", "c76233ea374421da562b5b022c22614e81f9cf862da93543cff93c37c085f136"},
  "gun": {:hex, :gun, "2.1.0", "b4e4cbbf3026d21981c447e9e7ca856766046eff693720ba43114d7f5de36e87", [:make, :rebar3], [{:cowlib, "2.13.0", [hex: :cowlib, repo: "hexpm", optional: false]}], "hexpm", "52fc7fc246bfc3b00e01aea1c2854c70a366348574ab50c57dfe796d24a0101d"},
  "cowlib": {:hex, :cowlib, "2.16.0", "54592074ebbbb92ee4746c8a8846e5605052f29309d3a873468d76cdf932076f", [:make, :rebar3], [], "hexpm", "7f478d80d66b747344f0ea7708c187645cfcc08b11aa424632f78e25bf05db51"},
```

**Additional context**

mvanhorn:
## Summary
Short-lived client connections no longer leak `GRPC.Client.Supervisor` children. Orchestrator processes that fail to finalize a connection are terminated instead of being left under the DynamicSupervisor, and disconnected connections are fully cleaned up, so the supervisor's binary memory stays flat under connect/disconnect churn.

## Why this matters
Reported in #531: a user running hundreds of short-lived connections (`GRPC.Stub.connect/2` then `GRPC.Stub.disconnect/1`) on 1.0.0-rc.1 sees `GRPC.Client.Supervisor` top `:recon.proc_count(:memory, 10)` and `:recon.bin_leak(10)` with persistent growth; downgrading to 0.10.2 (no DynamicSupervisor in the client path) eliminates it, and a collaborator confirmed the diagnosis and labeled it a bug. The regression is in the 0.11+/1.0 client connection architecture: `GRPC.Client.Connection.connect/2` leaves the orchestrator child running when `finalize_connection/2` returns `{:error, :no_connection}`, and the disconnect path leaves per-connection state behind.

## Changes
- `connect/2` now terminates the orchestrator child when connection finalization fails, so error paths cannot accumulate supervised processes.
- The disconnect path releases the connection's supervised resources instead of leaving the child alive under `GRPC.Client.Supervisor`.

## Testing
Added regression tests that drive repeated connect/disconnect cycles and connect-failure cycles, asserting `DynamicSupervisor.count_children/1` on `GRPC.Client.Supervisor` returns to baseline after each cycle. Ran the client connection test files with mix test.

Fixes #531

### DIFF
--- a/grpc/lib/grpc/client/application.ex
+++ b/grpc/lib/grpc/client/application.ex
@@ -5,7 +5,12 @@ defmodule GRPC.Client.Application do
   def start(_type, _args) do
     children = [
       {Registry, [keys: :unique, name: GRPC.Client.Registry]},
-      {DynamicSupervisor, [name: GRPC.Client.Supervisor]}
+      {DynamicSupervisor,
+       [
+         name: GRPC.Client.Supervisor,
+         hibernate_after: 15_000,
+         spawn_opt: [fullsweep_after: 20]
+       ]}
     ]
 
     opts = [strategy: :one_for_one, name: GRPC.Supervisor]
--- a/grpc/lib/grpc/client/application.ex
+++ b/grpc/lib/grpc/client/application.ex
@@ -3,13 +3,15 @@ defmodule GRPC.Client.Application do
   use Application
 
   def start(_type, _args) do
+    supervisor_config = Application.get_env(:grpc, __MODULE__, [])
+
     children = [
       {Registry, [keys: :unique, name: GRPC.Client.Registry]},
       {DynamicSupervisor,
        [
          name: GRPC.Client.Supervisor,
-         hibernate_after: 15_000,
-         spawn_opt: [fullsweep_after: 20]
+         hibernate_after: Keyword.get(supervisor_config, :hibernate_after, 15_000),
+         spawn_opt: [fullsweep_after: Keyword.get(supervisor_config, :fullsweep_after, 20)]
        ]}
     ]
 
--- a/grpc/lib/grpc/client/connection.ex
+++ b/grpc/lib/grpc/client/connection.ex
@@ -283,7 +283,17 @@ defmodule GRPC.Client.Connection do
       {:ok, pid} ->
         case ready_status(name, validated.resolver_target, timeout) do
           :ok ->
-            finalize_connection(name)
+            case finalize_connection(name) do
+              {:ok, %Channel{}} = result ->
+                result
+
+              {:error, _reason} = error ->
+                # A connection that reported ready but cannot hand out a
+                # channel is unusable; leaving it under the supervisor is the
+                # leak this fixes.
+                _ = DynamicSupervisor.terminate_child(GRPC.Client.Supervisor, pid)
+                error
+            end
 
           {:error, reason} ->
             _ = DynamicSupervisor.terminate_child(GRPC.Client.Supervisor, pid)
--- a/grpc/lib/grpc/client/connection.ex
+++ b/grpc/lib/grpc/client/connection.ex
@@ -291,7 +291,7 @@ defmodule GRPC.Client.Connection do
                 # A connection that reported ready but cannot hand out a
                 # channel is unusable; leaving it under the supervisor is the
                 # leak this fixes.
-                _ = DynamicSupervisor.terminate_child(GRPC.Client.Supervisor, pid)
+                :ok = DynamicSupervisor.terminate_child(GRPC.Client.Supervisor, pid)
                 error
             end
 
--- a/grpc/lib/grpc/client/connection.ex
+++ b/grpc/lib/grpc/client/connection.ex
@@ -289,8 +289,7 @@ defmodule GRPC.Client.Connection do
 
               {:error, _reason} = error ->
                 # A connection that reported ready but cannot hand out a
-                # channel is unusable; leaving it under the supervisor is the
-                # leak this fixes.
+                # channel is unusable. Keeping it in the supervisor could leak memory.
                 :ok = DynamicSupervisor.terminate_child(GRPC.Client.Supervisor, pid)
                 error
             end
--- a/grpc/lib/grpc/client/connection.ex
+++ b/grpc/lib/grpc/client/connection.ex
@@ -52,6 +52,20 @@ defmodule GRPC.Client.Connection do
   first establishment attempt finishes and returns `{:error, reason}` (tearing
   the process down) if that attempt fails.
 
+  ## Supervisor memory
+
+  Connection start arguments (for example large default headers) can linger in
+  the `GRPC.Client.Supervisor` process heap after children exit. That supervisor
+  therefore hibernates when idle and runs fullsweep GCs more often. Tune via
+  config:
+
+      config :grpc, GRPC.Client.Application,
+        hibernate_after: 15_000,
+        fullsweep_after: 20
+
+  * `:hibernate_after` – idle ms before the supervisor hibernates (default: `15_000`)
+  * `:fullsweep_after` – minor GCs between fullsweeps on the supervisor (default: `20`)
+
   ## Target syntax
 
   The `target` argument to `connect/2` accepts URI-like strings that are resolved
--- a/grpc/test/grpc/client/connection_test.exs
+++ b/grpc/test/grpc/client/connection_test.exs
@@ -81,15 +81,13 @@ defmodule GRPC.Client.ConnectionTest do
                )
 
       assert_receive {:connection_started, pid}, 500
-      assert_eventually(fn -> not Process.alive?(pid) end)
+      monitor_ref = Process.monitor(pid)
+      assert_receive {:DOWN, ^monitor_ref, :process, ^pid, _reason}, 500
 
-      assert_eventually(fn ->
-        Registry.lookup(GRPC.Client.Registry, {Connection, ref}) == []
-      end)
+      assert Registry.lookup(GRPC.Client.Registry, {Connection, ref}) == []
 
-      assert_eventually(fn ->
-        DynamicSupervisor.count_children(GRPC.Client.Supervisor).active == supervisor_children
-      end)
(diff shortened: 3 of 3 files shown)


sleipnir:
@mvanhorn Some tests are failing, could you please check them?

mvanhorn:
The OTP-matrix failures were the connection tests asserting `active == 0` on the shared `GRPC.Client.Supervisor` — under the full suite other tests leave children behind (CI showed `active: 44`), so the assertions failed even though the lifecycle fix works. 5a6f752 captures the supervisor's pre-existing child count in setup and asserts the count returns to that baseline instead. All 14 connection tests pass locally.

polvalente:
@mvanhorn could you resolve the merge conflicts?

mvanhorn:
Rebased onto master — conflicts resolved, `mix test` is green (366 passed, 2 skipped) and `mix format --check-formatted` is clean.

Worth flagging what the rebase surfaced: #557 already added `terminate_child` on the `ready_status` error path, so half of this PR landed independently. What's left is the `finalize_connection` failure path — a connection that reports ready but can't hand out a channel still returns `{:error, :no_connection}` to the caller while staying under `GRPC.Client.Supervisor`. That's the leak this now fixes, in the shape `connect/2` has after #557.

I also dropped my `restart: :temporary` change and the test that asserted a crashed connection is not restarted. #557's supervised connections deliberately restart on crash (and `terminate/2` documents relying on it), so `:transient` stays as you have it.

Remaining diff is the finalize-path cleanup plus a `hibernate_after` / `fullsweep_after` spawn_opt on the client supervisor so long-lived connection processes release binary-heavy heaps.

grpc/lib/grpc/client/application.ex
```diff
@@ -5,7 +5,12 @@ defmodule GRPC.Client.Application do
...
     children = [
       {Registry, [keys: :unique, name: GRPC.Client.Registry]},
-      {DynamicSupervisor, [name: GRPC.Client.Supervisor]}
+      {DynamicSupervisor,
+       [
+         name: GRPC.Client.Supervisor,
+         hibernate_after: 15_000,
+         spawn_opt: [fullsweep_after: 20]
+       ]}
     ]
```
polvalente, line 16: I wonder if these should be configurable via Application env

grpc/lib/grpc/client/connection.ex
```diff
@@ -283,7 +283,17 @@ defmodule GRPC.Client.Connection do
...
-            finalize_connection(name)
+            case finalize_connection(name) do
+              {:ok, %Channel{}} = result ->
+                result
+
+              {:error, _reason} = error ->
+                # A connection that reported ready but cannot hand out a
+                # channel is unusable; leaving it under the supervisor is the
+                # leak this fixes.
+                _ = DynamicSupervisor.terminate_child(GRPC.Client.Supervisor, pid)
```
polvalente: Should we hard-match on the result?

grpc/test/grpc/client/connection_test.exs
```diff
@@ -327,6 +398,19 @@ defmodule GRPC.Client.ConnectionTest do
...
+  defp assert_eventually(fun, attempts) when attempts > 0 do
+    if fun.() do
+      :ok
+    else
+      Process.sleep(10)
+      assert_eventually(fun, attempts - 1)
+    end
+  end
+
+  defp assert_eventually(fun, 0), do: assert(fun.())
```
polvalente: Let's not add `asssert_eventually` in favor of more robust strategies. I believe waiting for the DOWN message should be enough with other assertions happening after that.

mvanhorn:
Good call on dropping `assert_eventually` -- chasing the resulting CI failure turned up an actual race rather than just a slow assertion.

Waiting on the `:DOWN` isn't sufficient on its own here. `Registry` removes the entry from *its* monitor of the connection process, and that monitor and the test's fire on the same exit with no ordering guarantee, so the lookup can legitimately still see the entry. That's the OTP-matrix failure at `connection_test.exs:155` (`left: [{#PID<...>, nil}]`).

d4ad461 keeps the `:DOWN` wait and adds a synchronous call to the partition holding the monitor, which drains that message before the table is read. I put a small harness on it locally -- register / kill / read, 3000 iterations: 303 stale reads without the barrier, 0 with it. So it's deterministic rather than a narrower window, which I think is the "more robust strategy" you were after.

The assertions after the `:DOWN` are unchanged. Full suite is 366 passed / 2 skipped and `mix format --check-formatted` is clean.

grpc/lib/grpc/client/connection.ex
```diff
@@ -283,7 +283,17 @@ defmodule GRPC.Client.Connection do
...
           :ok ->
-            finalize_connection(name)
+            case finalize_connection(name) do
+              {:ok, %Channel{}} = result ->
+                result
+
+              {:error, _reason} = error ->
+                # A connection that reported ready but cannot hand out a
+                # channel is unusable; leaving it under the supervisor is the
+                # leak this fixes.
```
polvalente:
```suggestion
                # A connection that reported ready but cannot hand out a
                # channel is unusable. Keeping it in the supervisor could leak memory.
```

grpc/test/grpc/client/connection_test.exs
```diff
@@ -40,6 +59,37 @@ defmodule GRPC.Client.ConnectionTest do
...
+  end
+
+  describe "connect/2 - failed finalize" do
+    test "terminates a newly started child when no channel can be picked", %{
+      ref: ref,
+      adapter: adapter,
+      supervisor_children: supervisor_children
+    } do
+      Application.put_env(:grpc, NoPickResolver, self())
+      on_exit(fn -> Application.delete_env(:grpc, NoPickResolver) end)
```
polvalente, line 74: Not for this PR, but I think this test suite needs refactoring so we can speed it up

grpc/test/grpc/client/connection_test.exs
```diff
@@ -327,6 +397,18 @@ defmodule GRPC.Client.ConnectionTest do
...
+  # That monitor and the one these tests set up both fire on the same exit with
+  # no ordering guarantee, so receiving `:DOWN` does not mean the registry has
+  # processed its copy yet. A system call to the partition holding the monitor
+  # drains that message before we read the table. This is deterministic, where
+  # a bounded retry only makes the flake rarer -- locally the unsynchronized
+  # read is stale about 10% of the time under load.
+  defp await_registry_cleanup(registry) do
+    _ = :sys.get_state(Module.concat(registry, "PIDPartition0"))
+    :ok
+  end
```
polvalente: How about we pass the test as a listener to the registry? `https://elixir.hexdocs.pm/Registry.html#t:listener_message/0` 

This await seems a bit too reliant on internals of Registry.

polvalente: I'm looking into this possibility

polvalente (approved):

mvanhorn:
Really appreciate the merge here. Nice to stop leaking supervisor children on failed connections, thanks for the review.

PR #556, opened 2026-07-15
Fix Mint adapter: BadMapError in State.stream_response_pid/2

### LINKED ISSUE #555: Mint adapter: BadMapError in State.stream_response_pid/2 when :cancel_request races end-of-stream (:done)
**Describe the bug**

`GRPC.Client.Adapters.Mint.ConnectionProcess` crashes with `BadMapError: expected a map, got: nil` in `State.stream_response_pid/2` when a `{:cancel_request, ref}` GenServer call arrives after the server's end-of-stream (`:done`) frame for that same ref has already been processed (or after a previous cancel has already popped the ref).

`handle_call({:cancel_request, ref}, ...)` unconditionally calls `process_response({:done, ref}, state)`, which does `state.requests[ref].stream_response_pid`. If the ref was already popped, `state.requests[ref]` is `nil` and the map access raises. Broader case: `handle_info` reduces a batch of `responses` from `Mint.HTTP.stream/2` through `process_response/2`; if any frame in that batch references an already-popped ref, the same crash occurs.

**To Reproduce**

The race is timing-dependent in real traffic. It reliably surfaces when using `broadway_cloud_pub_sub` 2.0.0-rc.0 streaming pull against Google Cloud Pub/Sub - the `StreamManager` cancels the bidi stream on graceful shutdown / drain / stream refresh, and can race the server's `:done` for the same ref.

A self-contained unit reproducer (matches the exact production stack). Drop into `test/grpc/adapters/mint/connection_process_test.exs`:

```elixir
describe "handle_call/2 - cancel_request" do
  setup :valid_connection
  setup :valid_stream_request
  setup :valid_stream_response

  test "cancel arriving after ref was popped raises BadMapError",
       %{request_ref: request_ref, state: state} do
    # Simulate the server having already sent :done for this ref
    # (or a prior cancel having processed it).
    {_ref, state_without_ref} = pop_in(state.requests[request_ref])
    assert %{} == state_without_ref.requests

    ConnectionProcess.handle_call({:cancel_request, request_ref}, nil, state_without_ref)
  end
end
```

**Expected behavior**

Cancelling a request whose ref is no longer in `state.requests` should be a no-op on the internal state (`Mint.HTTP2.cancel_request/2` should still be called so any residual HTTP/2 stream state is released), and it should never crash the `ConnectionProcess` GenServer. Symmetrically, `process_response/2` should be tolerant of server frames referencing a ref that has already been popped.

**Logs**

Production log from a `broadway_cloud_pub_sub` streaming pull subscriber:

```
[error] @ gen_server.error_info/7 (gen_server.erl:2785) [#PID<0.9418.0>]
GenServer #PID<0.9418.0> terminating
** (BadMapError) expected a map, got:

    nil

    (grpc 1.0.1) lib/grpc/client/adapters/mint/connection_process/state.ex:72: GRPC.Client.Adapters.Mint.ConnectionProcess.State.stream_response_pid/2
    (grpc 1.0.1) lib/grpc/client/adapters/mint/connection_process/connection_process.ex:284: GRPC.Client.Adapters.Mint.ConnectionProcess.process_response/2
    (grpc 1.0.1) lib/grpc/client/adapters/mint/connection_process/connection_process.ex:184: GRPC.Client.Adapters.Mint.ConnectionProcess.handle_call/3
    (stdlib 8.0.2) gen_server.erl:2470: :gen_server.try_handle_call/4
    (stdlib 8.0.2) gen_server.erl:2499: :gen_server.handle_msg/3
    (stdlib 8.0.2) proc_lib.erl:333: :proc_lib.init_p_do_apply/3
Last message (from ...StreamManager_1): {:cancel_request, #Reference<0.731702437.709623812.68593>}

Initial Call: GRPC.Client.Adapters.Mint.ConnectionProcess.init/1
Ancestors: [...StreamManager_1, ...Broadway.Producer_1, ...Broadway.ProducerSupervisor,
            ...Broadway.Supervisor, ...Subscriber, ...Subscriber.Supervisor, ...]
```

Failing output from the unit reproducer above, on unpatched `1.0.2`:

```
** (BadMapError) expected a map, got: nil
    (grpc 1.0.2) lib/grpc/client/adapters/mint/connection_process/state.ex:72:
        GRPC.Client.Adapters.Mint.ConnectionProcess.State.stream_response_pid/2
    (grpc 1.0.2) lib/grpc/client/adapters/mint/connection_process/connection_process.ex:267:
        GRPC.Client.Adapters.Mint.ConnectionProcess.process_response/2
    (grpc 1.0.2) lib/grpc/client/adapters/mint/connection_process/connection_process.ex:167:
        GRPC.Client.Adapters.Mint.ConnectionProcess.handle_call/3
```

Line numbers differ slightly between `1.0.1` and `1.0.2` but the crash path is identical on both versions.

**Protos**

Not applicable - reproducible with any bidirectional streaming RPC that gets cancelled by the client. Originally observed against Google Cloud Pub/Sub `StreamingPull` (`google.pubsub.v1.Subscriber/StreamingPull`).

**Versions:**

- OS: Linux (production containers)
- Erlang: OTP 29 
- Elixir: 1.20
- mix.lock (`grpc`, `gun`, `cowboy`, `cowlib`):
  ```
  "grpc":   {:hex, :grpc,   "1.0.1", ...}   # also verified on 1.0.2
  "gun":    {:hex, :mint,    "1.9.2", ...} 
  ```

**Additional context**
- Affected adapter is `GRPC.Client.Adapters.Mint`. I have not verified whether the Gun adapter has an analogous race.

rockneurotiko:
Fixes #555

### DIFF
--- a/grpc/lib/grpc/client/adapters/mint/connection_process/connection_process.ex
+++ b/grpc/lib/grpc/client/adapters/mint/connection_process/connection_process.ex
@@ -165,7 +165,7 @@ if Code.ensure_loaded?(Mint.HTTP) do
     end
 
     def handle_call({:cancel_request, request_ref}, _from, state)
-        when State.has_request_ref?(state, request_ref) do
+        when State.has_request_ref(state, request_ref) do
       state = process_response({:done, request_ref}, state)
 
       case Mint.HTTP2.cancel_request(state.conn, request_ref) do
--- a/grpc/lib/grpc/client/adapters/mint/connection_process/connection_process.ex
+++ b/grpc/lib/grpc/client/adapters/mint/connection_process/connection_process.ex
@@ -13,6 +13,7 @@ if Code.ensure_loaded?(Mint.HTTP) do
     alias GRPC.Client.Adapters.Mint.StreamResponseProcess
 
     require Logger
+    require State
 
     @connection_closed_error "the connection is closed"
 
@@ -163,18 +164,18 @@ if Code.ensure_loaded?(Mint.HTTP) do
        {:continue, :process_request_stream_queue}}
     end
 
-    def handle_call({:cancel_request, request_ref}, _from, state) do
-      state =
-        if State.has_ref?(state, request_ref) do
-          process_response({:done, request_ref}, state)
-        else
-          state
-        end
+    def handle_call({:cancel_request, request_ref}, _from, state)
+        when State.has_request_ref?(state, request_ref) do
+      state = process_response({:done, request_ref}, state)
 
       case Mint.HTTP2.cancel_request(state.conn, request_ref) do
         {:ok, conn} -> {:reply, :ok, State.update_conn(state, conn)}
         {:error, conn, error} -> {:reply, {:error, error}, State.update_conn(state, conn)}
       end
+    end
+
+    def handle_call({:cancel_request, _request_ref}, _from, state) do
+      {:reply, :ok, state}
     end
 
     @impl true
@@ -232,14 +233,6 @@ if Code.ensure_loaded?(Mint.HTTP) do
     def terminate(_reason, _state) do
       :normal
     end
-
-    defp process_response({_kind, request_ref, _}, state)
-         when not is_map_key(state.requests, request_ref),
-         do: state
-
-    defp process_response({_kind, request_ref}, state)
-         when not is_map_key(state.requests, request_ref),
-         do: state
 
     defp process_response({:status, request_ref, status}, state) do
       State.update_response_status(state, request_ref, status)
--- a/grpc/lib/grpc/client/adapters/mint/connection_process/connection_process.ex
+++ b/grpc/lib/grpc/client/adapters/mint/connection_process/connection_process.ex
@@ -164,7 +164,12 @@ if Code.ensure_loaded?(Mint.HTTP) do
     end
 
     def handle_call({:cancel_request, request_ref}, _from, state) do
-      state = process_response({:done, request_ref}, state)
+      state =
+        if State.has_ref?(state, request_ref) do
+          process_response({:done, request_ref}, state)
+        else
+          state
+        end
 
       case Mint.HTTP2.cancel_request(state.conn, request_ref) do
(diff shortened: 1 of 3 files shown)


grpc/lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -228,6 +233,14 @@ if Code.ensure_loaded?(Mint.HTTP) do
       :normal
     end
 
+    defp process_response({_kind, request_ref, _}, state)
+         when not is_map_key(state.requests, request_ref),
+         do: state
+
+    defp process_response({_kind, request_ref}, state)
+         when not is_map_key(state.requests, request_ref),
```
polvalente: you could make State.has_ref? a guard (suggested has_request_ref as the guard name) to use both in these clauses and in the usage above

rockneurotiko: I changed it to use the guard, and I removed over-checked, I kept it focused on the cancel which is the bug I found

grpc/lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -163,7 +164,8 @@ if Code.ensure_loaded?(Mint.HTTP) do
        {:continue, :process_request_stream_queue}}
     end
 
-    def handle_call({:cancel_request, request_ref}, _from, state) do
+    def handle_call({:cancel_request, request_ref}, _from, state)
+        when State.has_request_ref?(state, request_ref) do
```
polvalente: guards should generally not have the `?` suffix, having `has_` or `is_` as their prefix.

https://elixir.hexdocs.pm/naming-conventions.html#trailing-question-mark-foo

rockneurotiko: Totally right sorry... Using LLMs is making my brain worse day by day 😂

polvalente (approved):
Thanks for the report + fix!

PR #557, opened 2026-07-16
feat(grpc): support supervised client connections

yordis:
- Long-running clients need connection lifecycles owned by their supervision tree; calling `connect/2` at runtime forces applications to hand-roll ownership, retry, and channel-handle plumbing.
- A backend that is unreachable at boot should not cascade into supervisor startup failures; the connection process now stays alive and retries with exponential backoff, while `connect/2` keeps its historical fail-fast contract for existing callers.
- A stable named channel handle lets stubs resolve to a healthy underlying connection on every RPC, so handles survive reconnects and endpoint changes instead of being invalidated by them.


yordis:
hey @sleipnir @polvalente could you give me a hand with this one? my intent is to make it as declarative to manage these channels, today, I see a lot of code around configuring and writing genserver around it to register to things (this is similar intent as conn_grpc pkg)

Porfa help me here 🙏🏻

sleipnir:
> hey @sleipnir @polvalente could you give me a hand with this one? my intent is to make it as declarative to manage these channels, today, I see a lot of code around configuring and writing genserver around it to register to things (this is similar intent as conn_grpc pkg)
> 
> Porfa help me here 🙏🏻

I will take a look!

grpc/lib/grpc/client/connection.ex
```diff
@@ -259,6 +419,52 @@ defmodule GRPC.Client.Connection do
...
+      def await_ready(timeout \\ 5_000) do
+        GRPC.Client.Connection.await_ready(__MODULE__, timeout)
+      end
+
+      def disconnect, do: GRPC.Client.Connection.disconnect(__MODULE__)
+
+      defoverridable child_spec: 1
+    end
+  end
+
```
polvalente: I don't like that we're defining too many methods for starting a connection. Let's remove this macro as it makes things too implicit IMO. I don't see a clear advantage in keeping it.

polvalente (changes requested):
Let's discuss the `use` macro

grpc/test/grpc/client/connection_test.exs
```diff
@@ -287,4 +358,17 @@ defmodule GRPC.Client.ConnectionTest do
...
+  defp eventually(fun, 0), do: fun.()
+
+  defp eventually(fun, retries) do
+    if fun.() do
+      true
+    else
+      Process.sleep(50)
+      eventually(fun, retries - 1)
+    end
+  end
```
polvalente: I think this smells of something missing. Perhaps a telemetry event to which we could subscribe and assert in the test?

grpc/test/grpc/client/connection_supervised_test.exs
```diff
@@ -0,0 +1,319 @@
...
+             end)
+
+      assert {:ok, %GRPC.Channel{}} = Connection.pick_channel(%GRPC.Channel{ref: name})
+    end
+  end
+
+  describe "await_ready/2 waiter lifecycle" do
+    setup do
+      Application.put_env(:grpc, :grpc_test_failing_hosts, ["127.0.0.1"])
+      on_exit(fn -> Application.delete_env(:grpc, :grpc_test_failing_hosts) end)
```
polvalente: This is a code smell IMO. @sleipnir let's refactor this into an option for hte FailingClientAdapter to enable tests to be async: true

polvalente: Not related to this PR, though.

yordis:
@polvalente I am not attach to all the public functions in the module at all, what I care the most is around the child spec, config and things like that, so I leave that to y'all

I will wait for @sleipnir feedback so I can address them all at once

polvalente:
> @polvalente I am not attach to all the public functions in the module at all, what I care the most is around the child spec, config and things like that, so I leave that to y'all
> 
> I will wait for @sleipnir feedback so I can address them all at once

Ok! I think the best guideline is to answer "what is the smallest API surface we have to introduce to suit the needs of this feature?"

yordis:
@polvalente fixed all, check again

grpc/lib/grpc/client/connection.ex
```diff
@@ -23,6 +23,55 @@ defmodule GRPC.Client.Connection do
   * Each call to `pick/2` dispatches to the LB module, which selects a channel
     per request. DNS re-resolution reconciles the LB's channel list in place.
 
+  ## Supervised connections
+
+  For long-running clients, declare the connection in your own supervision
+  tree instead of calling `connect/2` at runtime. In that mode, establishment
+  happens asynchronously after process start: if the backend is unreachable,
+  the process stays alive and retries with exponential backoff instead of
+  failing supervisor startup. Use `await_ready/2` to block until the channel
```
sleipnir, line 32: I like it when every mention of a specific function's usage is accompanied by a specific example. Could you provide one for `await_ready`?

grpc/lib/grpc/client/connection.ex
```diff
@@ -23,6 +23,55 @@ defmodule GRPC.Client.Connection do
...
+
+      # config/runtime.exs
+      config :my_app, MyApp.PaymentsConnection,
+        target: "dns://payments.internal:50051",
+        lb_policy: :round_robin
+
+      # in your supervision tree
+      children = [MyApp.PaymentsConnection]
+
+      channel = GRPC.Client.Connection.get_channel!(MyApp.PaymentsConnection)
```
sleipnir: Perhaps a tuple like `{:ok, channel}`, `{:error, :initializing}`, and/or `{:error, error}` would be better API than returning just the channel. 
But that raises questions regarding the rest of the API. What do you think, @polvalente ?

sleipnir (changes requested):

grpc/lib/grpc/client/connection.ex
yordis: @sleipnir I feel somewhat strong around the exception raising here, most of the time, not having such connection is actually "stop the world and get out" since that is a system wide problem than "domain" problem to deal with.

The alternative is to pattern match on `{:ok, channel} = ...` which will cause a pattern matching exception, essentially, hard signal to deal with in o11y tools.

I do my best to avoid exception raising, but this is one of those that I rather dont deal with another `with` statement for something that MUST always be up in most cases.

Being said, I do not feel strong about it, I can wrap modules around :)

sleipnir: I don't have a strong opinion on this either; in fact, my stance on APIs has always been that "if something is documented, then it's valid." Let's let @polvalente weigh in, and we'll go from there.

polvalente: `get_channel!` should only exist if `get_channel` returning the tagged tuple works. So if both ways exist, then we can satisfy both people who want to follo @yordis's approach or @sleipnir's. So perhaps we should just add a comment above this call with `# use get_channel for the non-raising variant`

grpc/lib/grpc/client/connection.ex
polvalente: This was resolved with no response nor a provided example. I've unresolved it until either is met

grpc/lib/grpc/client/connection.ex
polvalente: Also unresolved as this should be removed

polvalente: Likewise, the `use_opts` from `start_link` should also go away

grpc/lib/grpc/client/connection.ex
```diff
@@ -269,45 +502,99 @@ defmodule GRPC.Client.Connection do
...
+        {:reply, :ok, state}
+
+      true ->
+        {:reply, {:error, state.last_error || :connecting}, state}
+    end
+  end
+
+  def handle_call(:await_ready, {caller_pid, _tag}, %__MODULE__{established?: true} = state) do
+    await_ready_start(state, caller_pid)
+    await_ready_stop(state, caller_pid, System.monotonic_time(), :ok)
```
polvalente: I think these should be `telemetry_await_*` or `*_event`. `await` is the wrong verb here

grpc/lib/grpc/client/connection.ex
```diff
@@ -332,23 +629,241 @@ defmodule GRPC.Client.Connection do
...
+    Enum.find_value(real_channels, fn
+      {_key, {:failed, reason}} -> reason
+      _ -> nil
+    end)
+  end
+
+  defp maybe_init_resolver(%__MODULE__{resolver_state: rs}) when not is_nil(rs), do: rs
+
+  defp maybe_init_resolver(%__MODULE__{} = state) do
+    if Code.ensure_loaded?(state.resolver) and function_exported?(state.resolver, :init, 2) do
```
polvalente, line 757: Not sure if this should be happening in a genserver like this. Not a blocker though, but this has some smell of "heavy code in hot path"

grpc/lib/grpc/client/connection.ex
```diff
@@ -332,23 +629,241 @@ defmodule GRPC.Client.Connection do
...
+  end
+
+  defp lifecycle_metadata(%__MODULE__{} = state) do
+    %{name: state.virtual_channel.ref, target: state.resolver_target, pid: self()}
+  end
+
+  defp backoff_delay(attempt) do
+    base = min(@backoff_initial * :math.pow(@backoff_multiplier, attempt), @backoff_max)
+    jitter = (:rand.uniform() * 2 - 1) * @backoff_jitter * base
+    max(trunc(base + jitter), 0)
```
polvalente:
```suggestion
    trunc(base + jitter)
```

I don't think this can ever be < 0. this is effectively `base * (1 + jitter_percent)` with `jitter_percent` bounded from -0.2 to 0.2, which has the lowest point in `base * 0.8`, and `base` is non-negative, with a min value of `1000`. By the way, I suggest lowering it to `@backoff_initial 100`, so attempts are 100, 160, 256, 400ish, 700ish, 1000ish (that is, the first 4 attempts accumulate to 1s).

grpc/lib/grpc/client/connection.ex
yordis: @polvalente to be clear, only the child_spec is left behind? not understand the final version you want

polvalente: I don't think `__using__` should be defined at all

yordis: Do you mind sharing your reason? That is what actually drove me to create the PR; related to https://github.com/TheRealReal/conn_grpc/pull/20

I feel that this is a common pattern that helps to reduce the "custom" way of doing the exact same thing in the ecosystem.

Otherwise, `conn_grpc` will stick around. I feel this is a decent compromise.

grpc/lib/grpc/client/connection.ex
yordis: Added an Examples section to `await_ready/2`'s docs in fc352ab; the supervised-connection example in the moduledoc also shows it in context (3338b92).

yordis: I don't recall resolving this one, completely missed it, sorry

grpc/lib/grpc/client/connection.ex
yordis: Somewhat "hot path" since it is one per startup, regardless, what do you recommend me to do here?

grpc/lib/grpc/client/connection.ex
polvalente: Too many ways of doing the same thing. The macro does that cover any use case that can't be done by the plain functions.

grpc/lib/grpc/client/connection.ex
sleipnir: I think that if  `resolve`, can block, so it should be in a `continue` statement.

```
{:ok, %{state | resolver_state: nil}, {:continue, :init_resolver}}
```

grpc/lib/grpc/client/connection.ex
yordis: As the end-user of the pkg, I rather appreciate this way, including the previous version, I never come across anything that would require me other than `use ConnGRPC.Channel` since all the time that was what we needed.

Every team, across every codebase, is familiar with 1-way to do it, just that it is not the one you promoted.

Honestly, I strong disagree with the pushback on this one, I seen enough way to configure a trivial `child_spec` by now, we all have to relearn what things are all the time for such trivial thing to fix.

polvalente: I rather just expose a child_spec function in the module that accepts the name/id as a parameter.

polvalente: Having a macro like this adds more maintenance load and API surface. Unless there's hard evidence that this macro is needed (for example code snippets that show a practical use case where this macro makes code a lot easier to handle), we shouldn't add it.

yordis: > Having a macro like this adds more maintenance load and API surface.

This give no room for discussion, and ultimate, you are the maintainer to decide that, no point in arguing. This isn't a new pattern around the ecosystem at all.

> Unless there's hard evidence that this macro is needed

The tricky part is "who defines what needed means?" I gave you hard evidence since this comes from real, production evidence of what some, like us, are doing around the package.

If you feel strong about it, lets just move on from it and we effective continue to add something like `conn_grpc` to simply put back a user experience that understand the maintain load in production software as well.

polvalente:
```elixir
children = [
  {GRPC.Client.Connection,
   name: MyApp.PaymentsConnection,
   target: "dns://payments.internal:50051",
   lb_policy: :round_robin},
  {GRPC.Client.Connection,
   name: MyApp.UsersConnection,
   target: "dns://users.internal:50051"}
]
```

What does this method of defining things doesn't do that the macro solves?

yordis: that isn't the full code in a production environment, most of those configurations comes from `config/*` files where people need to define,

1. What scope (would be under `config :myapp` or `config :myapp, ModuleName`
2. Is it `target` and `lb_policy` or will somebody come up with their own keys?
3. Do you read from `System` or `Application` env?
4. Do you create a reusable helper in `application.ex` or osmething else?

Like, far too many bike shedding over things, at some point, such abstraction is introduced so developers focus on more important matter, without leaving behind a trail of "so many ways" to accomplish the same thing we all need to learn.

That is what I found in production software that isn't a hand full of people.

polvalente: Wouldn't the decision from where these values come from would still happen with the macro?

polvalente: Specially, the macro solves compile-time values, not runtime ones, which would still appear in the same pattern I showed above with the single difference that `:name` would be omitted and become the module to be started instead.

yordis: that values are read at `child_spec` time, so, at that point, you merge the compile-time defaults and the runtimes that could be set from `config` files

polvalente: Ah, I see the one thing the macro simplifies which is the `grpc_connection_otp_app` key that is passed to `start_link`!

I don't think that is a very strong reason to add a macro given that the substitution is as simple as `{GRPC.Client.Connection, [name: name] ++ Application.get_env(otp_app, name)}`.

polvalente: > that values are read at child_spec time, so, at that point, you merge the compile-time defaults and the runtimes that could be set from config files

I was thinking about the `use` opts, but the `otp_app` option does indeed hide away the Application.get_env thing. Sorry it seems I ignored this message, it popped up while I was typing my previous reply.

yordis: > {GRPC.Client.Connection, [name: name] ++ Application.get_env(otp_app, name)}

You are assuming that people will do that, I gave you a list of combination of what people do in production code.

Being said, I removed it.

grpc/lib/grpc/client/connection.ex
polvalente: Let's move on with just this. These shouldn't be starting up too often anyway.

polvalente (approved):

grpc/lib/grpc/client/connection.ex
yordis: `init/1` doesn't do any blocking work, unless I missed something.

It only registers the `persistent_term` handle and returns `{:ok, state, {:continue, :establish}}`, so resolution, dialing, and resolver `init` all already run in `handle_continue` and never block the supervisor.

This path also only runs once per connect/reconnect; RPCs read the channel from persistent_term and never go through the GenServer, so there's no per-request hot path here.

Also, aren't we doing what you are suggesting?

sleipnir (approved):

PR #561, opened 2026-07-23
refactor: remove GRPC.Channel.t() typespec (unbreaks connect/2 success typing)

pfitz:
## Summary

`GRPC.Channel`'s `@type t` was narrower than the struct's actual value space,
which made Dialyzer infer that `GRPC.Stub.connect/2` could only return
`{:error, _}`. Any downstream code matching `{:ok, channel}` on the result then
got a spurious `pattern_match` warning ("The pattern `{:ok, _}` can never match
the type `{:error, _}`"), even though at runtime `connect/2` returns
`{:ok, channel}` normally.

Per review, rather than relaxing the type to match the `defstruct` defaults,
this PR **removes the `GRPC.Channel.t()` typespec altogether**. Typespec-only,
no behavior change.

## Root cause

```elixir
# lib/grpc/channel.ex
@type t :: %__MODULE__{
        host: String.t(),          # defstruct default: nil
        port: non_neg_integer(),   # defstruct default: nil
        ...
        interceptors: [],          # real value is a non-empty list
        ...
      }
defstruct host: nil, port: nil, ...
```

`GRPC.Client.Connection.build_initial_state/2` builds the **virtual channel**
without `:host`/`:port` (they stay `nil`) and with a non-empty `:interceptors`
list — a legitimate value for a virtual / load-balanced channel, but one the type
above forbids.

`connect/2`'s success branch flows through `finalize_connection/2 → pick_channel/2`,
and `pick_channel/2` carried `@spec pick_channel(GRPC.Channel.t(), keyword())`.
Dialyzer therefore proved `pick_channel(virtual_channel, _)` could never satisfy
`GRPC.Channel.t()` (`nil ∉ String.t()`/`non_neg_integer()`, non-empty list `∉ []`),
so `finalize_connection/2` was `no_return` and `connect/2`'s success typing collapsed
to `{:error, _}`.

(v1.0.1 didn't hit this because its `connect/2` returned `{:ok, ch}` directly from
the `start_child` branch, without the `pick_channel` spec check.)

## Fix

Delete `@type t` from `GRPC.Channel` and replace the ~16 `Channel.t()` /
`GRPC.Channel.t()` references across specs and callbacks with `struct()`:

```elixir
@spec connect(String.t(), keyword()) :: {:ok, struct()} | {:error, any()}
```

`struct()` imposes no field constraints, so the over-narrow type is gone and the
false positive cannot recur — without having to keep a hand-maintained,
easily-drifting struct type in sync with `defstruct`. The now-unused
`alias GRPC.Channel` in `adapter.ex` and `load_balacing.ex` is dropped too.

## Reproduction

Any consumer matching `connect/2`'s result tripped it:

```elixir
def connect_probe do
  case GRPC.Stub.connect("localhost:50051", []) do
    {:ok, ch} -> ch          # dialyzer: pattern {:ok, _} can never match {:error, _}
    {:error, r} -> r
  end
end
```

`mix dialyzer` flagged the `{:ok, ch}` clause. With this change the warning
disappears.

### DIFF
--- a/grpc/lib/grpc/channel.ex
+++ b/grpc/lib/grpc/channel.ex
@@ -16,20 +16,6 @@ defmodule GRPC.Channel do
     * `:adapter_payload` - payload the adapter uses
   """
 
-  @type t :: %__MODULE__{
-          host: String.t() | nil,
-          port: non_neg_integer() | nil,
-          scheme: String.t() | nil,
-          cred: GRPC.Credential.t() | nil,
-          ref: reference() | nil,
-          adapter: atom(),
-          adapter_payload: any(),
-          codec: module(),
-          interceptors: list(),
-          compressor: module() | nil,
-          accepted_compressors: [module()],
-          headers: list()
-        }
   defstruct host: nil,
             port: nil,
             scheme: nil,
--- a/grpc/lib/grpc/channel.ex
+++ b/grpc/lib/grpc/channel.ex
@@ -17,16 +17,16 @@ defmodule GRPC.Channel do
   """
 
   @type t :: %__MODULE__{
-          host: String.t(),
-          port: non_neg_integer(),
-          scheme: String.t(),
-          cred: GRPC.Credential.t(),
+          host: String.t() | nil,
+          port: non_neg_integer() | nil,
+          scheme: String.t() | nil,
+          cred: GRPC.Credential.t() | nil,
           ref: reference() | nil,
           adapter: atom(),
           adapter_payload: any(),
           codec: module(),
-          interceptors: [],
-          compressor: module(),
+          interceptors: list(),
+          compressor: module() | nil,
           accepted_compressors: [module()],
           headers: list()
         }
--- a/grpc/CHANGELOG.md
+++ b/grpc/CHANGELOG.md
@@ -15,7 +15,7 @@
 
 ### Bug Fixes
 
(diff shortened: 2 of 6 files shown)


grpc/lib/grpc/channel.ex
```diff
@@ -17,16 +17,16 @@ defmodule GRPC.Channel do
   """
 
   @type t :: %__MODULE__{
```
polvalente: Let's just remove the typespec altogether

pfitz: Ok - removed the specs alltogether.

polvalente (approved):

yordis:
Hey folks,

This would been a breaking change, they are packages using such type, eg https://github.com/TheRealReal/conn_grpc/blob/b40034e4c77fa2ab1033c94fa92acb4aec99e6af/lib/conn_grpc/pool.ex#L119 along with some private code that was pointing to it as well

We are forced to disable the Dialyzer situation as follow:

```
[
 ~r/Unknown type: GRPC\.Channel\.t\/0/
]
```

😭

polvalente:
@yordis We dropped support for Dialyzer quite a while back, this was just a leftover bug from us not verifying things.

The library you mentioned specifies 0.0 as the requirement, and we did many breaking changes since the library last updated dependencies.

edit: the library also specifies grpc as a dev and test only dependency, so I'm not sure how users of both would be affected. This seems like a latent bug in grpc_pool

yordis:
> We dropped support for Dialyzer

What does that means? no Dialyzer types at all anymore?

polvalente:
> > We dropped support for Dialyzer
> 
> 
> 
> What does that means? no Dialyzer types at all anymore? 

Exactly. We haven't run dialyzer on CI or as a dependency for, if I'm not mistaken, years. So it's basically impossible for us to properly support typespecs as a hard dependency. So any time a Dialyzer bug is reported, the only proper fix is to drop that annotation because it's wrong. I'll remove all annotations and release 1.0.4.

Consider any 1.0.x patches still in the same spirit of "we're under the prerrogative of breaking against 0.y"

polvalente:
@yordis 1.0.4 has been released with explicitly dropping all @type annotations, which should avoid PRs like this in the future and removes any public facing type interfaces (types, specs; callbacks absorb the annotations inline).

yordis:
🫡 aye!

PR #559, opened 2026-07-17
Fix Mint adapter: dead stream response process crashes the connection

ycastorium:
Fixes #558

PR #562, opened 2026-07-24
fix(mint): handle stream-level errors in process_response instead of crashing

freevova:
## Problem

  When a server (or an LB/ingress in front of it) sends GOAWAY while requests are in flight, `Mint.HTTP2.stream/2` returns the affected requests as `{:error, request_ref,
  %Mint.HTTPError{reason: :unprocessed, module: Mint.HTTP2}}` entries in the responses list — documented Mint behavior: error tuples are one of the possible response entries.

  `ConnectionProcess.process_response/2` only has clauses for `:status`, `:headers`, `:data` and `:done`, so the error entry raises `FunctionClauseError` inside `handle_info/2` (the
  `Enum.reduce(responses, state, &process_response/2)` call) and the whole connection process crashes:

  ```
  ** (FunctionClauseError) no function clause matching in GRPC.Client.Adapters.Mint.ConnectionProcess.process_response/2
      (grpc 1.0.2) lib/grpc/client/adapters/mint/connection_process/connection_process.ex:248:
          GRPC.Client.Adapters.Mint.ConnectionProcess.process_response(
            {:error, #Reference<0.3496360168.1708654593.148949>,
             %Mint.HTTPError{reason: :unprocessed, module: Mint.HTTP2}},
            %GRPC.Client.Adapters.Mint.ConnectionProcess.State{
              conn: %Mint.HTTP2{state: {:goaway, :no_error, ""}, ...},
              ...
            })
      (grpc 1.0.2) lib/grpc/client/adapters/mint/connection_process/connection_process.ex:212:
          GRPC.Client.Adapters.Mint.ConnectionProcess.handle_info/2
  ```

  ## Why this is worse than a single failed request

  The crash is an abnormal exit, which bypasses both recovery mechanisms of the adapter:

  1. the `retry`/`attempt_reconnect` loop never runs (it only handles graceful disconnects), and
  2. the `{:elixir_grpc, :connection_down, pid}` notification to the parent is never sent (it is only sent from the graceful path).

  The channel owner only receives a linked `{:EXIT, pid, {:function_clause, ...}}` — and since `connect/2` sets `Process.flag(:trap_exit, true)` on the calling process, the owner does not
  crash either; unless it explicitly handles the EXIT message it keeps serving a channel whose `conn_pid` is dead, and every caller that had a request in flight at crash time exits or
  hangs.

  We run this client in production against a gRPC service behind an nginx ingress that recycles idle HTTP/2 connections (GOAWAY roughly every 60s of idle). The race fires whenever GOAWAY
  lands while unary calls are in flight — for us that is several times per hour, each occurrence killing or wedging the background jobs whose calls were in flight.

  ## Reproduction

  1. Connect with the Mint adapter to any HTTP/2 server that sends GOAWAY on idle/rotation (an nginx ingress with default keepalive settings works).
  2. Keep a steady stream of unary calls so that requests are in flight when GOAWAY lands.
  3. Eventually `Mint.HTTP2.stream/2` returns `{:error, ref, %Mint.HTTPError{reason: :unprocessed}}` for the raced requests → `FunctionClauseError` → dead connection process.

  ## Fix

  Add a `process_response({:error, request_ref, error}, state)` clause that delivers the error to the request's stream response process (`consume(:error, ...)` + `done`) and drops the
  request from state — mirroring what `send_connection_close_and_end_stream_response/1` does on the disconnect path. The raced request then surfaces to the caller as a normal per-request
  error (`:unprocessed` is retryable by contract — the server did not process the request), and connection-level GOAWAY handling proceeds through the existing graceful paths.

  The full test suite passes with the patch (342 tests, 2 skipped).

  ## Note on tests

  I did not find a way to deterministically reproduce the GOAWAY race against the in-repo test server, since it requires the server to send GOAWAY while a stream is mid-flight. Happy to
  add a regression test if you can suggest a preferred approach (e.g. a unit-level seam for `process_response/2`, or a test server capable of forcing GOAWAY).


grpc/lib/grpc/client/adapters/mint/connection_process/connection_process.ex
```diff
@@ -238,91 +245,63 @@ if Code.ensure_loaded?(Mint.HTTP) do
...
+    # connection process with a FunctionClauseError (which also skips the
+    # retry logic and the :connection_down notification).
+    defp process_response({:error, request_ref, error}, state) do
+      pid = State.stream_response_pid(state, request_ref)
+      :ok = StreamResponseProcess.consume(pid, :error, error)
+      :ok = StreamResponseProcess.done(pid)
 
       {_ref, new_state} = State.pop_ref(state, request_ref)
       new_state
     end
```
polvalente, line 294: Can we add a test that exercises this?

polvalente:
@freevova can you enable the "allow edits from maintainers" flag?

freevova:
Superseded by #566 — while rebasing I found the crash is already fixed on master by #559; what's left is the regression test, now opened from a personal fork (GitHub doesn't offer "allow edits from maintainers" for organization-owned forks, which is where this branch lived). Closing.

PR #566, opened 2026-07-27
test(mint): cover incoming stream-level errors in process_response

freevova:
Follow-up to #562, replacing it (opened from a personal fork this time — the previous branch lived in an organization-owned fork, where GitHub doesn't offer "allow edits from maintainers" at all, and was also cut from the v1.0.2 tag, which dragged unrelated commits into the diff).

  While rebasing onto master I found the crash itself is already fixed there by #559 — including the unknown-ref guard, which also covers a second production crash variant we'd hit (the connection process dying on `consume_response` to an already-dead stream response process). So what's left of the original PR is the regression test, adapted to master.

  The test injects a raw RST_STREAM(REFUSED_STREAM) frame for an in-flight request as if it arrived on the socket — the same `{:error, ref, %Mint.HTTPError{}}` response-entry shape that GOAWAY produces for requests it races. It runs against a bare TCP server that only completes the HTTP/2 preface and then stays silent, so no real server frames race the injected one. Removing the `process_response` error clause from #559 makes the test fail with the original `FunctionClauseError`.

  One ask: could you cut a release? hex 1.0.2 crashes in production under routine ingress GOAWAYs (several times per hour for us), so we're running a pinned fork until a release with #559 is out.

### DIFF
--- a/grpc/test/grpc/adapters/mint/connection_process_test.exs
+++ b/grpc/test/grpc/adapters/mint/connection_process_test.exs
@@ -590,6 +590,41 @@ defmodule GRPC.Client.Adapters.Mint.ConnectionProcessTest do
     end
   end
 
+  describe "handle_info - incoming stream errors" do
+    setup :quiet_connection
+    setup :valid_stream_request
+    setup :valid_stream_response
+
+    test "delivers per-stream errors to the stream response process instead of crashing",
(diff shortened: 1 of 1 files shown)


sleipnir (approved):

sleipnir:
Thank you @freevova

PR #568, opened 2026-07-30
fix: recover supervised connections when the transport process dies

enilsen16:
When a transport process dies, `GRPC.Client.Connection` never notices: Gun transports live under the adapter's supervisor with no link or monitor back to the connection process. The dead channel stays in the load-balancing rotation, and every RPC that picks it crashes with a `FunctionClauseError` in the Gun adapter — permanently, until the connection process is restarted.

## What this PR does

- **Detects transport death.** Every connected transport is monitored; a `:DOWN`/`:EXIT` from a tracked transport removes its channel from rotation immediately.
- **Reconnects.** If other channels remain healthy, the dead endpoint is redialed (via the resolver when there is one, otherwise a repair loop over the last resolved address set). If it was the last channel, the connection re-enters its establish loop — immediately for a stable connection, with exponential backoff when deaths come within `:flap_window` (default 10s) of establishing.
- **Fails cleanly while down.** RPCs return `{:error, %GRPC.RPCError{status: 14}}` (UNAVAILABLE) instead of crashing the caller; request-streaming calls raise, since their return value is a stream. Failures still flow through the interceptor chain and telemetry.
- **Tracks recovery accurately.** `await_ready/2` and `connect/2` succeed as soon as any healthy channel exists, whichever path restored it.

## API additions

- `:flap_window` connection option (ms, default `10_000`).
- Optional `terminate/1` callback on `GRPC.Client.LoadBalancing`, so re-establishment reuses ETS-backed balancer state instead of leaking one table per reconnect.

## Edge cases covered

Policy flips dispose the old balancer only after the new one initializes; a resolve failure during redial keeps the last known addresses instead of downgrading the LB policy; the optional `Resolver.update/2` callback is guarded; retry timers dedup so outages can't accumulate concurrent retry loops; a reconnected-then-dead channel can't fake readiness; call options are validated even while the connection is down, so config errors raise instead of masking as UNAVAILABLE.


grpc/lib/grpc/client/connection.ex
```diff
@@ -581,36 +671,122 @@ defmodule GRPC.Client.Connection do
...
 
-  def handle_info({:DOWN, mon, :process, pid, reason}, state) do
-    case Enum.split_with(state.waiters, fn {_pid, _from, m, _started_at} -> m == mon end) do
-      {[], _} ->
-        Logger.warning(
-          "#{inspect(__MODULE__)} received :DOWN from #{inspect(pid)} with reason: #{inspect(reason)}"
-        )
+  # A connection that dies within this window of establishing counts as a
+  # flap, and each consecutive flap backs the redial off further.
+  @flap_window 10_000
```
sleipnir: This could be configured dynamically, right?

yordis:
hey @sleipnir any updates on this one?

sleipnir (approved):
@enilsen16 Please resolve the conflicts so we can proceed.

sleipnir:
> hey @sleipnir any updates on this one?

coming soon @yordis

sleipnir:
Thank you @enilsen16

enilsen16:
Thanks I have a small followup commit :)

Will have a PR shortly

PR #579, opened 2026-08-27
fix(mint): explicitly support infinite retries

arrangabriel:
# Explicitly support `retry: :infinity` in the Mint adapter

The option already worked, because atoms are above integers in erlang term ordering.
This behavior was undocumented, and arguably supported only by accident.

*Changes*

- :retry now documents and tests :infinity as an intentional value.
- The exhaustion check moved into a State.retries_exhausted?/1 guard that names the special case instead of leaning on term ordering.
- Invalid :retry values now raise ArgumentError instead of silently behaving in an unexpected way.
- New optional GRPC.Client.Adapter.validate_opts/1 callback. GRPC.Client.Connection.connect/2 invokes it when the adapter exports it. Implemented only in the Mint adapter.

*Heads-up for downstream test suites*

Mox generates optional callbacks on mocks, so an existing Mox.defmock for
GRPC.Client.Adapter will export validate_opts/1 and raise
UnexpectedCallError unless it is stubbed. Adapters using the behaviour without
defining validate_opts/1 are fine.


arrangabriel:
Sorry to pester you, but the failure here also seems to be due to a flaky test @sleipnir

PR #583, opened 2026-08-31
fix(client): harden supervised connection recovery edge cases

enilsen16:
Follow-ups to #568 — a set of recovery edge cases found while reviewing that change after merge.

## Connection (`lib/grpc/client/connection.ex`)

- **Stale retry timers**: `schedule_retry` deduped on a boolean but never cancelled a pending timer, so a transport death after a resolver-driven recovery could sit out the remainder of an old long backoff (up to ~2min) instead of redialing at the newly computed delay. Now stores the timer ref, cancels/reschedules, and `adopt_established` cancels any pending retry.
- **Dead-channel adoption race**: `rebalance_after_reconcile` adopted on `connected != []` without the `channel_alive?` check the `:retry_establish` handler applies for exactly this race, so a channel that died with its `:DOWN` still queued could produce a false `:connected` (waiters released, telemetry emitted) until the `:DOWN` unwound it.
- **Stale LB state for pickers**: the LB state returned by `lb_mod.update/2` was only kept in GenServer state; `pick_channel` reads `:persistent_term`, which was only re-put in `establish/1`. A behaviour-compliant LB that returns fresh state from `update/2` (rather than mutating ETS in place) left pickers on the stale state indefinitely. Now republished when the state changes (no-op for the built-ins).
- **Backoff ladder reset after flaps**: `adopt_established` reset `retry_attempt` to 0 but not `flaps`, so an establish failure right after a flap-death restarted the dial ladder at the shortest delay — the backoff could shrink after a failure. `handle_channel_down` now seeds `retry_attempt` from the flap count.
- **Silent resolver failures**: non-`{:ok, _}` returns from `resolver.update/2` were silently discarded (the removed `handle_cast(:resolve_now)` used to assert the match). Now logged, keeping the previous resolver state.

## Stub (`lib/grpc/stub.ex`)

- **Crash instead of UNAVAILABLE**: `fallback_channel`'s catch-all passed a payload like `%{conn_pid: nil}` — the documented return of `Connection.disconnect/1` — straight to the adapter, crashing on a nil pid. Any payload carrying `:conn_pid` is now liveness-checked (nil/dead/remote → UNAVAILABLE); only payloads without the key pass through.
- **Pid-less adapters treated as dead**: `resolve_channel` used `Map.get(payload, :conn_pid)`, conflating a missing key with nil, while `channel_alive?` treats the same shape as alive — so a healthy connection over an adapter that exposes no transport pid could never serve an RPC from the stub side. Now aligned with `channel_alive?`.
- **Failure results bypassed connection config**: `unavailable_result` ran the caller handle's (often empty) interceptor list with struct-default codec/compressor, while the healthy path uses the picked channel's config — so bare `%Channel{ref: name}` handles skipped the connection's interceptors on failures only. Now resolves config via `Connection.get_channel/1` and honors the caller's `codec`/`compressor` opts.
- **Wasted re-picks on non-rotating policies**: the bounded re-pick loop burned all attempts on the identical dead channel under PickFirst (plus a `Logger.warning` per attempt). The loop now bails when the policy returns the same pid again.

## Verification

`mix test`: 10 doctests, 360 tests, 0 failures (2 pre-existing skips). `mix format --check-formatted` clean on both files.

### DIFF
--- a/grpc/lib/grpc/stub.ex
+++ b/grpc/lib/grpc/stub.ex
@@ -434,10 +434,6 @@ defmodule GRPC.Stub do
          req_stream,
          opts
        ) do
-    # A bare %Channel{ref: name} handle carries none of the connection's
-    # configuration, so interceptors/codec/compressor are resolved through the
-    # stored virtual channel — the same config a picked real channel inherits —
-    # keeping the failure path consistent with the healthy one.
     config_ch =
       case Connection.get_channel(channel.ref) do
         {:ok, %Channel{} = virtual_channel} -> virtual_channel
--- a/grpc/lib/grpc/stub.ex
+++ b/grpc/lib/grpc/stub.ex
@@ -278,13 +278,6 @@ defmodule GRPC.Stub do
   #    * Client streaming. A `GRPC.Client.Stream`
   #    * Server streaming. `{:ok, Enumerable.t} | {:ok, Enumerable.t, trailers_map} | {:error, error}`
   #
-  #  Any call made through a named connection's virtual channel fails with
-  #  UNAVAILABLE while the connection has no healthy underlying channel to
-  #  resolve to: `{:error, %GRPC.RPCError{status: 14}}` for unary and
-  #  server-streaming calls, raised as `GRPC.RPCError` for request-streaming
-  #  calls (their return value is a stream). Both flow through the channel's
-  #  interceptors and client telemetry.
-  #
   #  Options
   #
   #    * `:timeout` - request timeout. Default is 10s for unary calls and `:infinity` for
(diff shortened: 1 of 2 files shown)


grpc/lib/grpc/stub.ex
```diff
@@ -422,13 +431,34 @@ defmodule GRPC.Stub do
...
          res_mod,
-         req_stream
+         req_stream,
+         opts
        ) do
-    stream = %{stream | request_mod: req_mod, response_mod: res_mod}
+    # A bare %Channel{ref: name} handle carries none of the connection's
+    # configuration, so interceptors/codec/compressor are resolved through the
+    # stored virtual channel — the same config a picked real channel inherits —
+    # keeping the failure path consistent with the healthy one.
```
sleipnir: unascessary comments, please remove it

sleipnir (changes requested):

grpc/lib/grpc/stub.ex
enilsen16: Updated thank you

sleipnir:
Thank @enilsen16

